PR #2, opened 2021-02-26
Add data adapter

bobular:
Would like review of this please.

Have incorporated histogram and barplot requests into the `<Distribution>` component.

One thing I still need to figure out is detecting the empty results from two different flavours of `variableSummary`.  Though this will likely change when Danielle tweaks the return format of the JSON to be more object-y.


dmfalke:
This is really great! I have a couple of comments.

First, maybe we should hold off on merging this into `main`. I can work on integrating the histogram component on this branch. We can merge into `main` once it is fully functional. Thoughts?

Second, it looks like you are planning to have separate methods for numeric vs date datatypes. This is good. I assume the `getDateHistogramNumBins` method will have a date-specific response type/decoder. Otherwise, if both methods share the same response type, it would be littered with `union([number, string])` types, which would get tedious to deal with. Sorry to be pedantic. Just want to make sure you agree.

Thanks!

PR #3, opened 2021-03-02
handled all four histogram requests but not tested yet

bobular:
NOTE: merges into `add-data-adapter` not main!

Following the newly updated API specs here: https://veupathdb.github.io/EdaDataService/api.html

I made adapter methods for all four histogram endpoints.

I don't think it's going to work in practice because the responses will not always have 
`data[].overlayVariableDetails` or `data[].facetVariableDetails` in them, so the io-ts validator needs to be aware of that.

(and I just saw that the latter is an array - will fix this ASAP)

### DIFF
--- a/src/lib/core/api/data-service.ts
+++ b/src/lib/core/api/data-service.ts
@@ -10,6 +10,7 @@ import {
   array,
   type,
   tuple,
+  union,
   intersection,
   partial,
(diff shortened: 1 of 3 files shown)


dmfalke:
@bobular I pushed a commit that addresses the optional properties in `HistogramResponseData`.

PR #5, opened 2021-03-03
data-api - facet request tidy up

bobular:
I had missed the multiple values for facetVariable on the request side - so tidied this up.

Might also allow an empty array in the facetVariable response after I make this PR.


bobular:
Not sure how to do io-ts for an empty array/tuple for the facetVariable response, sorry!

dmfalke (approved):

PR #10, opened 2021-03-18
Hello world basic histogram with two variable pickers

bobular:
Making a draft PR so this is easier to review, but no plans to merge until the wiring and "nesting" is more mature.


bobular:
Closing because merging into main was too hard.

See `new-vis-histo` branch which has the same code, mostly!

PR #17, opened 2021-03-24
Add entity diagram size prop

chowington:
NOTE: Depends on https://github.com/VEuPathDB/web-components/pull/80

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@veupathdb/eda",
   "version": "0.1.12",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


chowington:
Also, is there a reason there are two entity diagram components? At this point there's a lot of duplication between the two. Should we consolidate?

dmfalke:
https://github.com/VEuPathDB/web-components/pull/80#issuecomment-806974541

dmfalke (approved):

PR #20, opened 2021-03-25
Viz tab prototype

dmfalke:
# Description

This PR adds the basic infrastructure for the **Visualize** tab in the EDA Workspace application. It delivers the basic mechanism for adding new Visualizations for the Passthrough computation.

# Changes / additions

* New type definitions for **Visualization** and **Computation**
* `computations` entry to **Session** type
* Routing infrastructure for computations and visualizations
* Example **PassThroughComputation** component to illustrate basics of wiring visualizations into a computation.

# Details

The **Visualize** tab is driven by **Computations** configured for the EDA Workspace Application.

### Computation

A **Computation** represents an analysis that can be applied to a subset. The output of a **Computation** is essentially an additional set of variables that can be used in a **Visualization**, in addition to the _native_ variables provided by the study.

Currently, only one **Computation** has been implemented: the _Passthrough_ computation. It does not provide any additional output. It simply passes data through from the subsetting service.

### Visualization

A **Visualization** is a graphical and/or textual summary of a set of variables, applied to a **Computation**. Typically, a **Visualization** will include a plot, controls, and some set of statistics. In many cases, a **Visualization** can be used in more than one **Computation**. Exactly how this will be done is an open question, but will likely involve a way to statically configure an **Visualization** for the context of a **Computation**.

## Screens

There are three primary screens in the **Visualize** tab:
1. Configured Visualizations
2. Selection window to add a new Visualization
3. A "fullscreen" view of a Visualization.

These screens are scoped to a **Computation**. It is ultimately up to the **Computation** component to determine how the user chooses **Visualizations** to add, but the current expectation is to use the set of screens above. They are encapsulated in a `VisualizationsContainer` component. See [PassThroughComputation.tsx](https://github.com/VEuPathDB/web-eda/blob/viz-tab-prototype/src/lib/core/components/computations/PassThroughComputation.tsx) for an implementation.

If more than one **Computation** is available for a study, there will also be a set of screens for viewing configured Computations and for adding a new Computation. These have yet to be implemented, as they are not required for the immediate deliverable.

### `Visualization`

```typescript
interface Visualization {
  id: string,
  computationId: string,
  type: string,
  configuration: unknown,
}
```

This type describes the object that is stored in the `visualizations` array of the session object. Note that `configuration` is `unknown`. It is up to the rendering components (see below) to validation the configuration. It is strongly recommended to use an `io-ts` Type to perform the validation. We will likely want to add some utilities to help with this.

### `Computation`

```typescript
interface Computation {
  id: string,
  type: string,
  configuration: unknown,
}
```

This type describes the object that is stored in the `computations` array of the session object. Similar to `Visualization`, `configuration` is `unknown`. Validation should be done in the **Computation** component.

### `VisualizationProps`

```typescript
interface VisualizationProps {
  visualization: Visualization;
  computation: Computation;
}
```

This interface describes the props passed to **Visualization** components (see below).

### `VisualizationType`

```typescript
interface VisualizationType {
  type: string;
  displayName: string;
  gridComponent: React.ComponentType<VisualizationProps>;
  fullscreenComponent: React.ComponentType<VisualizationProps>;
  selectorComponent: React.ComponentType;
  createDefaultConfig: () => unknown;
}
```

This type describes how to render a **Visualization** in various contexts, and how to generate a default configuration. The default configuration is used to create a new instance of a **Visualization** when a user chooses it from the selector screen. There will probably be parts of the configuration that are undefined, such as variables. This should be taken into account when implementing the various components.

# Remaining TODOs

_These will be completed in a follow-up_

- [ ] Add configurable display name to `Visualization`, including rendering with an editable title
- [ ] Add support for configuring `Computations` for a particular website. E.g., pass an array to `WorkspaceRouter`.


dmfalke:
A follow up thought: it is likely that the configuration of an app or visualization will change over time. We should think about how to migrate an old configuration to a new configuration, or some other UX for this.

dmfalke:
**Todos to address before removing draft status** 

- [ ] Update type names to match description
- [ ] Convert inline styles to SCSS
- [ ] Put test visualization component in `impl` directory

dmfalke:
After some discussion today, it was decided to refer to **Apps** (short for **Analysis Apps**) to **Computations**. I will update this branch and PR description to reflect that.

dmfalke:
The description has been updated to reflect the change of **App** to **Computation**.

PR #22, opened 2021-03-29
Histogram into viz infrastructure

bobular:
Basic histogram now working in Dave's framework.

Although there a many things to fix/improve, one major one is the VariableTrees not highlighting the current selected variable.

Another is to get the pass-through App to tell the histogram back end calls to use the `.../pass/...` URL path.


dmfalke (approved):
This all looks good. I plan to make some changes to how some things are passed to **App** components, after this is merged, in the `viz-tab-prototype` branch.

PR #30, opened 2021-04-01
Missing visualization management actions

### LINKED ISSUE #23: Visualize tab: Additional UI management actions for visualizations
The following UI controls should be added to visualizations:
1. Edit name of visualization.
2. Delete a visualization.

dmfalke:
Add ability to name and delete visualizations. Closes #23 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "0.2.4",
+  "version": "0.2.5",
   "dependencies": {
     "@material-ui/core": "^4.11.3",
     "@types/debounce-promise": "^3.1.3",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "@types/debounce-promise": "^3.1.3",
     "@veupathdb/components": "^0.2.36",
     "@veupathdb/wdk-client": "^0.1.1",
-    "@veupathdb/web-common": "^0.1.1-alpha.11",
+    "@veupathdb/web-common": "^0.1.1",
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
     "io-ts": "^2.2.13",
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "@veupathdb/eda",
-  "version": "0.1.15",
+  "version": "0.1.17",
   "dependencies": {
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.2.9",
+    "@veupathdb/components": "^0.2.10",
     "@veupathdb/wdk-client": "^0.1.1-alpha.14",
     "@veupathdb/web-common": "^0.1.1-alpha.11",
     "debounce-promise": "^3.1.2",
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
   "name": "@veupathdb/eda",
-  "version": "0.2.5",
+  "version": "0.2.6",
   "dependencies": {
     "@material-ui/core": "^4.11.3",
     "@types/debounce-promise": "^3.1.3",
     "@veupathdb/components": "^0.2.36",
-    "@veupathdb/wdk-client": "^0.1.1",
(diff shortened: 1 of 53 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "react-dom": "^17.0.1",
     "uuid": "^8.3.2"
   },
+  "sideEffects": false,
```
bobular, line 17: Curious what this is for...

bobular (commented):
I haven't been through all the code - it's my day off!  But I do have some general observations/suggestions after playing with the checked-out code for a while

- After clicking "Copy session" the new session name could be "<Old session name> copy"
- The visualizations could also have a copy button (with same behaviour as above)
- While in subsetting tab, "Copy session" creates new session, back at the default subsetting variable (see also #35 )
- While in viz tab, "Copy session" creates a "Visualization not found." message on the screen.  However, if I switch to subsetting and back to viz, the visualizations were copied just fine.
- The grid view viz's should show the viz name.

bobular:
PS. the delete session button in the actual session page (not the session list) seems non-operational.

dmfalke:
Thanks for the feedback, @bobular. I think I am going to incorporate addressing those issues in this PR. I will make github issues for those. I might also include the variable picker updates in this PR. I haven't decided yet.

dmfalke:
I've merged the latest from `main` into this branch. I think it's ready for review.

package.json
moontrip: @bobular  I was curious about that so I googled it to learn about the option :) Although you may find it easily, here is a summary. It is related to Webpack's optimization, so-called tree shaking. Simply put, Webpack tries to exclude unused external library/package to minimize its compiled file size (optimization), but certainly it often makes codes not work correctly, so-called side effects. The optimization is performed when the sideEffects is set to false (meaning side effects are free), so any side effects by Webpack's optimization become developers' responsibility. Actually, the option can be set at a library's package.json too.

Certainly it is more complex than what I said earlier: please see https://webpack.js.org/guides/tree-shaking/ if you are interested

dmfalke:
> I haven't been through all the code - it's my day off! But I do have some general observations/suggestions after playing with the checked-out code for a while
> 
>     * After clicking "Copy session" the new session name could be `<Old session name> copy`
> 
>     * The visualizations could also have a copy button (with same behaviour as above)
> 
>     * While in subsetting tab, "Copy session" creates new session, back at the default subsetting variable (see also #35 )
> 
>     * While in viz tab, "Copy session" creates a "Visualization not found." message on the screen.  However, if I switch to subsetting and back to viz, the visualizations were copied just fine.
> 
>     * The grid view viz's should show the viz name.

@bobular I just realized I forgot about these issues you raise. I will address these soon. I guess I will put this back in draft mode for now.

dmfalke:
@bobular I've addressed the issues you raised. This is ready for review now. Thanks!

bobular:
Looks great. Functionally I can't find any issues - and noted the analysis actions are also working nicely :-)

I'm not sure that the viz's should be shown in most-recently-edited first order.  Seems to be any config edit (title or plot settings).  This clearly isn't urgent for the upcoming demo.  Shall I make a ticket for this?   I guess a discussion on what grid view really is (is it a proto-dashboard?) would be useful first.  Should the grid items be re-orderable by dragging for example?

In terms of aesthetics and UX, I'll just make a couple of comments, even though I'm sure it's going to get a proper look at in a separate ticket

1. expanding a plot in grid view should simply involve clicking on it (I know this depends on some plot-components props standardisation by me, e.g. for switching off mouse-over tooltips that are still active on Histogram)
2. in grid-view, the copy and delete icons could appear on mouse-over of the minimised viz item (no icons shown when not-moused over)
3. in fullscreen view, you probably already know this, the copy and minimise icons are visually similar/confusing.

Also for another ticket/mockup meeting, font sizes and overall page layout could be made clearer. (For example, the study name is prefixed with "Study:" but the analysis and viz aren't.)  No idea what the solution is though (not my strong suit).
image

dmfalke:
I am going to merge this. I will make issues for things discussed here.

PR #33, opened 2021-04-01
Subsetting date fixes

bobular:
There are a few small fixes for HistogramVisualization too.

The binWidth slider works nicely now for all variables.  (The only issue is a back-end 500 when date binWidth is >= 10 months.)

All the date variables in GEMS bin to month resolution, so it hasn't been tested very hard for days/weeks/years etc.  When we do the zooming https://github.com/VEuPathDB/web-components/issues/70 we should get other date resolutions.

This PR depends on https://github.com/VEuPathDB/web-components/pull/91 

I acknowledge there's a fair bit of duplication between HistogramFilter and HistogramVisualisation - so that might be good to refactor some time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.17",
   "dependencies": {
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.2.10",
+    "@veupathdb/components": "^0.2.13",
     "@veupathdb/wdk-client": "^0.1.1-alpha.14",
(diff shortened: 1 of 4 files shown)


dmfalke (changes requested):
This looks good.

https://github.com/VEuPathDB/web-components/pull/91 has been merged and a new version has been published to npm.

You can upgrade to the latest with
```
yarn add @veupathdb/components
```

You will need to do that and commit the changes to `package.json` and `yarn.lock` before this can be merge.

Issue #41, opened 2021-03-16
Plot controls: reduce font size


dmfalke:
We can start by trying for 90%. Part of the challenge is that material-ui is using **rem** units. We might be able to use a scale CSS transform. Other options may exist.

PR #42, opened 2021-04-12
Variable selector

### LINKED ISSUE #25: Visualization: Variable selector TODOs
### High-level TODOs
- [ ] Render as enhanced select box. _Might want a generalized `CheckboxTreeSelect` component._
- [ ] Update `VariableTree` to support disabled entries. _May require updates to `CheckboxTree`._
- [ ] Integrate with plot controls. _May require adding support for ad-hoc control panel._
- [ ] Use constraints to disable incompatible variables.
- [ ] Include a filter toggle to hide incompatible variables.

### Details
Every visualization will include some form of variable selection. The user interface for this should be consistent and unobtrusive. The current thinking is to use an enhanced select widget that, when focused, displays the variable tree and allow the user to select an item. This will likely involve a new, generalize select component to support this.

A visualization can also specify variable constraints that encode the allowed permutations of variable selection. In some cases, when a visualization has multiple variables, the selection of one can constrain the available set of variables for another. The most common case for this would be constraining to entity.

The current proposed type interface for variable constraints is:
```typescript
interface PatternSet {
  [AxisName: string]: {
    required?: boolean;
    inheritEntityFrom?: string;
    patterns: {
      type?: string[];
      shape?: string[];
    }[];
  }
}

```
_This is subject to change as requirements evolve._

This information will be used to
1. Disable incompatible variables.
2. Hide incompatible variables based on filter toggle state.

The `CheckboxTree` component in WDKClient will soon support adding such filter toggles.

dmfalke:
This PR adds an `InputVariables` component, which renders a set of `VariableTreeDropdown` components. It also incorporates computation app metadata into the **Visualize** tab.

Depends on https://github.com/VEuPathDB/web-components/pull/100
Fixes #25 #41

### DIFF
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -1,47 +1,43 @@
-import { VisualizationProps, VisualizationType } from '../VisualizationTypes';
-
-import React, { useCallback, useMemo } from 'react';
-import { StudyEntity, StudyVariable } from '../../../../core';
+import HistogramControls from '@veupathdb/components/lib/components/plotControls/HistogramControls';
 import Histogram, {
   HistogramProps,
 } from '@veupathdb/components/lib/plots/Histogram';
-import { HistogramData } from '@veupathdb/components/lib/types/plots';
 import {
   ErrorManagement,
   NumberOrTimeDelta,
   NumberOrTimeDeltaRange,
   TimeDelta,
 } from '@veupathdb/components/lib/types/general';
-import HistogramControls from '@veupathdb/components/lib/components/plotControls/HistogramControls';
-import Switch from '@veupathdb/components/lib/components/widgets/Switch';
-import { useDataClient, useStudyMetadata } from '../../../hooks/workspace';
-import { usePromise } from '../../../hooks/promise';
+import { isTimeDelta } from '@veupathdb/components/lib/types/guards';
+import { HistogramData } from '@veupathdb/components/lib/types/plots';
 import { Loading } from '@veupathdb/wdk-client/lib/Components';
+import { preorder } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
+import { getOrElse } from 'fp-ts/lib/Either';
+import { pipe } from 'fp-ts/lib/function';
+import * as t from 'io-ts';
+import { isEqual } from 'lodash';
+import React, { useCallback, useMemo } from 'react';
 import {
   DataClient,
   DateHistogramRequestParams,
   NumericHistogramRequestParams,
 } from '../../../api/data-api';
-import { PromiseType } from '../../../types/utility';
+import { usePromise } from '../../../hooks/promise';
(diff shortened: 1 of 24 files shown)


dmfalke:
@bobular pinging you for a review :)

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -444,17 +428,16 @@ export function histogramResponseToData(
 function getRequestParams(
   studyId: string,
   filters: Filter[],
-  entity: StudyEntity,
-  variable: HistogramVariable,
-  overlayEntity?: StudyEntity,
-  overlayVariable?: StudyVariable, // TO DO: ?CategoricalVariable?
+  variable: Variable,
+  variableType: 'number' | 'date',
+  overlayVariable?: Variable,
```
bobular, line 433: Well this is certainly much cleaner - not passing variable and entity for both variables!

bobular (approved):
There's a lot here!

I've looked through all the diff and understand the main changes but haven't reviewed every line!

Functionally it's perfect except for the placeholder cartoon histogram being fleetingly very large in grid view (when it shows visualisations you've already made).  I will see if I can put in a suggestion or commit to fix this during my next meeting.

What's the plan for using the computation's name (e.g. `pass`) in the visualisation's get-data call? I notice the `/pass/` is still hardcoded in data-api.ts   Will that be another ticket?

src/lib/workspace/ComputationRoute.tsx
```diff
@@ -11,6 +12,14 @@ export interface Props {
...
   const { sessionState } = props;
   const { url } = useRouteMatch();
+  const dataClient = useDataClient();
+  const computations = usePromise(
+    useCallback(() => dataClient.getApps(), [dataClient])
+  );
+  const computationAppOverview = computations.value?.apps.find(
+    (app) => app.name === 'pass'
+  );
+  if (computationAppOverview == null) return null;
```
bobular, line 22: I was also wondering if we need a `<Suspense>` for this bit?
Or are we assuming the getApps() call will always be fast?

dmfalke: It should be pretty fast. I don't think this endpoint does any database lookups, but we should still handle possible slow responses. I'll add this.

dmfalke:
Ah, I guess I forgot about using the computation name for the data call. This isn't really urgent since we don't have other computations yet. It's probably a good idea to deal with it now.

bobular:
Another thing is that the OverlayVariable `clear` link clears both variables, not just the overlay one

image

dmfalke:
> Another thing is that the OverlayVariable `clear` link clears both variables, not just the overlay one

D'oh! Good catch. I will also address soon.

dmfalke:
@bobular I've addressed the issues you raised:

* Loading indicator for apps data (note that Suspense is not production ready for data fetching. I added a new component `PromiseResult` that can be used with the return value of the `usePromise` hook.
* Use computation `type` property when fetch visualization data.
* Fix the **clear** button in `InputVariables.tsx`

I am going to merge. Thanks!

PR #49, opened 2021-04-14
Variable tree & chips styling changes

chowington:


chowington:
Committing styling as a starting point per meeting discussion.

PR #56, opened 2021-04-16
Add new controls to Histogram

### LINKED ISSUE #43: Histogram: Use all relevant controls
This depends on https://github.com/VEuPathDB/web-components/pull/87.

This will require changes to `HistogramFilter` and `HistogramVisualization`. For `HistogramFitler`, the state should be stored as a part of `UIState`. New entries will need to be added, and the `updateUIState` callback will also need to be updated.

The associated data fetching function in both components will also need to be updated to include the x-axis range as `viewportMin` and `viewportMax`, if modified (see https://veupathdb.github.io/EdaDataService/api.html#resources:/apps/pass/visualizations/numeric-histogram-bin-width:post).

Something similar will need to be done for `HistogramVisualization`, whose state is stored in `HistogramConfig`.

The final set of controls should be:
- general
  - selected range
- y-axis
  - log scale
  - range
- x-axis
  - bin width
  - range

Let me know if you have questions.

### LINKED ISSUE #67: Update client code for unified Histogram API
Was four endpoints, it's now one.  Changes will be required to data-api.ts and the histogram-using components.

bobular:
Closes #43 

This will now include update to new API version (single histogram endpoint).

Closes #67

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -5,7 +5,6 @@ import Histogram, {
 import {
   DateRange,
   ErrorManagement,
-  NumberRange,
   TimeDelta,
   NumberOrTimeDelta,
   NumberOrTimeDeltaRange,
@@ -18,7 +17,7 @@ import {
 import { Loading } from '@veupathdb/wdk-client/lib/Components';
 import { getOrElse } from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/function';
-import { number, string, partial, TypeOf } from 'io-ts';
+import { number, partial, TypeOf, boolean } from 'io-ts';
 import React, { useCallback, useMemo } from 'react';
 import {
   DataClient,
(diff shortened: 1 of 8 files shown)


src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -140,7 +145,14 @@ export function HistogramFilter(props: Props) {
...
   );
   const data = usePromise(
-    useCallback(() => getData(uiState), [getData, uiState])
+    useCallback(() => getData(uiState), [
+      getData,
+      uiState.binWidth,
+      uiState.binWidthTimeUnit,
+      uiState.independentAxisRange,
+    ])
+    // is there some more concise utility to remove a key or keys from an object?
```
bobular, line 154: is there some more concise utility to remove a key or keys from an object?
I tried lodash.omit and it created an endless loop of API calls...

Another option would be to add some structure to uiState: split it into back-end and front-end pieces.

bobular:
I will get the histogram visualization working again (probably without any fancy controls) before pushing for merge.

dmfalke (approved):

PR #57, opened 2021-04-16
Incorporate variable constraints

dmfalke:
This PR uses variable constraints to limit the variables available for a visualization.

### DIFF
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -29,6 +29,7 @@ export interface Props {
   entities: StudyEntity[];
   /**
    * Current set of values for `inputs`.
+   * In other words, the currently selected variables.
    */
   values: ValueByInputName;
   /**
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -84,7 +84,14 @@ const useStyles = makeStyles(
 );
 
 export function InputVariables(props: Props) {
-  const { inputs, entities, values, onChange, constraints } = props;
+  const {
+    inputs,
+    entities,
+    values,
+    onChange,
+    constraints,
+    dataElementDependencyOrder,
+  } = props;
   const classes = useStyles();
   const handleChange = (inputName: string, value?: Variable) => {
     onChange({ ...values, [inputName]: value });
@@ -97,14 +104,23 @@ export function InputVariables(props: Props) {
   // 2. If next defined, root is next's entity, otherwise default root entity
   // 3. If prev is defined, remove prev's entity's children
   // 4. Return tree.
-  const rootEntities = inputs.map((input, index) => {
+  const rootEntities = inputs.map((input) => {
+    if (dataElementDependencyOrder == null) return entities[0];
+
     // 1
-    const inputValues = inputs.map((input) => values[input.name]);
-    const prevValue = inputValues
+    const index = dataElementDependencyOrder.indexOf(input.name);
+    // return root entity if dependencyOrder is not declared
+    if (index === -1) return entities[0];
+
+    const prevValue = dataElementDependencyOrder
       .slice(0, index)
+      .map((n) => values[n])
       .reverse()
       .find((v) => v != null);
-    const nextValue = inputValues.slice(index + 1).find((v) => v != null);
+    const nextValue = dataElementDependencyOrder
+      .slice(index + 1)
+      .map((n) => values[n])
+      .find((v) => v != null);
 
     // 2
     const rootEntityId = nextValue?.entityId ?? entities[0].id;
@@ -123,6 +139,7 @@ export function InputVariables(props: Props) {
           rootEntity
         );
   });
+
   return (
     <div className={classes.root}>
       <div className={classes.inputs}>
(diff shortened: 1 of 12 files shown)


jtlong3rd (approved):

bobular:
Augmenting the in-code documentation so that it makes sense more quickly to a new developer...
Will commit soon!

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -0,0 +1,123 @@
...
+              constraint.allowedTypes.includes(variable.type)))
+      ),
+      children,
+    }),
+    (e) => e.children ?? [],
+    rootEntity
+  );
+}
+
+export type ValueByInputName = Partial<Record<string, Variable>>;
```
bobular, line 36: Could you explain the use of `Partial<>` here for me please?  Does it allow a record with no entries?

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -0,0 +1,123 @@
...
+      const constraintB = constraintMapB[key];
+      return [
+        key,
+        constraintA == null
+          ? constraintB
+          : constraintB == null
+          ? constraintA
+          : {
+              isRequired: constraintA.isRequired || constraintB.isRequired,
+              maxNumVars: Math.max(
```
bobular, line 103: My hunch says the maxNumVars should be merged with Math.min() (and likewise min with max).

Merging two constraints that are min: 1, max: 6, and min: 2 and max: 3 should merge to min: 2, max: 3, right?

If the two constrains were min:1, max:2 and min:4, max:5, then these couldn't be merged, right?

Also, how are multiple (e.g. 2x facet) variables handled by InputVariables?  Can `InputVariables.props.values` store two variables for `facetVariable`?

src/lib/core/components/visualizations/InputVariables.tsx
```diff
@@ -82,28 +83,60 @@ const useStyles = makeStyles(
...
+  const { inputs, entities, values, onChange, constraints } = props;
   const classes = useStyles();
-  const handleChange = (inputName: string, value?: VariableDescriptor) => {
+  const handleChange = (inputName: string, value?: Variable) => {
     onChange({ ...values, [inputName]: value });
   };
+  const flattenedConstraints =
+    constraints && flattenConstraints(values, entities, constraints);
+
+  // Steps
```
bobular, line 94: I don't really get what this is about, sorry!
It's something to do with providing the entity/variable data tree to each menu in the InputVariables interface. Is the ordering something to do with `dataElementDependencyOrder`? It can't be, because this is unused at the moment.

bobular (commented):
It looks great, but I'd appreciate some more explaining how it works in one place please.
I'll commit/push my documentation updates after submitting this review.

src/lib/core/utils/data-element-constraints.ts
dmfalke: `Partial<T>` means that some values in `T` may not be defined. With `Record<string, Variable>` on its own, all values are considered to be `Variable`. Adding `Partial` means that values can be `Variable` or `undefined`.

src/lib/core/utils/data-element-constraints.ts
dmfalke: I think you are right. This is incomplete work. I think it can be addressed in a follow-up.

src/lib/core/components/visualizations/InputVariables.tsx
dmfalke: This does have to do with `dataElementDependencyOrder`. Line 99 should begin with
```typescript
   const rootEntities = dataElementDependencyOrder.map((inputName, index) => {
```

Good catch!

dmfalke: I will work on fixing this up

dmfalke: I pushed a fix for this. Not exactly what I put above.

src/lib/core/utils/data-element-constraints.ts
dmfalke: I fixed the min/max issue. I'm not sure how to handle incompatible cases. It might put a damper on this entire approach!

dmfalke: It may be safe to assume that these values will not vary across constraints. It's hard to imagine a case where {min,max}NumVars would change based on other criteria. We could start with throwing an error if the numbers aren't the same.

src/lib/core/components/visualizations/InputVariables.tsx
bobular: Glad to have helped, without fully understanding this part!

src/lib/core/utils/data-element-constraints.ts
bobular: Yeah I agree - it's unlikely that e.g. facetVariable can take 5-7 variables for "date vs number" plots and 1-3 variables for "number vs number".

Definitely a can to kick down the road if we put in place the safeguard you mention.

> Also, how are multiple (e.g. 2x facet) variables handled by InputVariables? Can InputVariables.props.values store two variables for facetVariable?

Is this also something to deal with when we have a faceting strategy (on the components side of things) ?

dmfalke: Yes, I think we can deal with it then.

dmfalke:
I'm going to merge this. I will make issues for the items identified in this review. Thanks!

PR #58, opened 2021-04-16
Add missing pagination-related props and uiState

### LINKED ISSUE #38: Subsetting: SCORE study variables geohash_5 to geohash_7 and village name give empty category barchart tables
The back end seems to be returning data fine.  It returns 252 categories for geohash_5.  The problem may be with the front end (haven't investigated further).

image

Edit to add: so do text-categorical variables with 443 values, e.g. Household->Geographic location->Village
Back end response looks OK to me.

dmfalke:
Fixes #38

### DIFF
--- a/src/lib/core/components/filter/TableFilter.tsx
+++ b/src/lib/core/components/filter/TableFilter.tsx
@@ -149,12 +149,14 @@ export function TableFilter({
 
   const uiStateKey = `${entity.id}/${variable.id}`;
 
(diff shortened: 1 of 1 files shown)


bobular (approved):
Tested in action and reviewed the code.

I understand the use of partial UIState: if a new client version has a new, extra UIState key, the stored, old UIState won't fail the validator, and the pipe/map/getOrElse merges any missing key/values in from the defaultUIState.

PR #64, opened 2021-04-19
Date to string fixes

bobular:
Dates are now represented as strings.

Dates used in filters must be padded Zulu date-times, so padISODateTime() function modified from previous ISODateStringToZuluDate().

Requires https://github.com/VEuPathDB/web-components/pull/106 to compile/pass checks

Two bin sorts (on binStart) were removed from HistogramFilter and HistogramVisualization because Histogram itself now does a full sort of all bins (mainly to determine the min and max range for the x-axis).

Ready to merge if approved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "dependencies": {
     "@material-ui/core": "^4.11.3",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.2.24",
+    "@veupathdb/components": "^0.2.30",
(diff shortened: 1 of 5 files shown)


dmfalke (approved):
This all looks good to me. Once https://github.com/VEuPathDB/web-components/pull/106 is merged, and a new package is published, you should be able to upgrade the package version in this branch.

bobular:
Can close this PR because the branch is already merged into #56

PR #74, opened 2021-04-27
Scatter plot viz

moontrip:
A basic scatter plot viz is made but it needs more works, e.g.,:

a) add line plot
b) add density plot
: note that scatter, line, and density plots are handled at the same plot component, <ScatterAndLinePlotGeneral>
c) typing (typescript): some are currently assigned as any
d) probably add plot controls: this may be handled in the other PR

### DIFF
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -192,6 +192,66 @@ export const BarplotResponse = type({
   ),
 });
 
+//DKDK scatterplot
+export interface ScatterplotRequestParams {
+  studyId: string;
+  filters: Filter[];
+  config: {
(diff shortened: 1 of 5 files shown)


moontrip:
a) added lineplot: note that it does not work at this stage because the backend respond incorrect data format starting with scatterplot, not lineplot. Discussed with Danielle and she confirmed the issue and would fix it
b) changed axes labels as selected X and Y variables' displayName
c) merging main: fixed error on scatter plot viz caused by merging

dmfalke (approved):
I skimmed the changes, and nothing major stands out as problematic. Overall I think this is good to merge. There may be follow-up changes in the future.

PR #75, opened 2021-04-28
Fixes (for barplot/categorical table only) to catch up with back end changes

bobular:
I couldn't fix the histogram endpoints because these are already half-fixed in my issue-43 branches...


bobular:
You can merge this into your working branches, or merge it to main, or both, I don't mind!

dmfalke (approved):

PR #79, opened 2021-05-06
Default variable behavior

dmfalke:
This PR includes
* Better selection of default variable when an entity is selected. It is now based on the tree structure of the variables.
* Automatically collapse inactive entities in the variable tree, while retaining the expansion state of internal nodes. This way the user can go back and forth between entities without losing expansion state.

### DIFF
--- a/src/lib/core/components/EDAWorkspaceContainer.tsx
+++ b/src/lib/core/components/EDAWorkspaceContainer.tsx
@@ -18,7 +18,7 @@ export interface Props {
   sessionClient: SessionClient;
   subsettingClient: SubsettingClient;
   dataClient: DataClient;
-  makeVariableLink?: (entityId: string, variableId: string) => string;
+  makeVariableLink?: (entityId?: string, variableId?: string) => string;
 }
 
 export function EDAWorkspaceContainer(props: Props) {
--- a/src/lib/core/components/EDAWorkspaceContainer.tsx
+++ b/src/lib/core/components/EDAWorkspaceContainer.tsx
@@ -3,7 +3,10 @@ import React from 'react';
 import { SubsettingClient } from '../api/subsetting-api';
 import { DataClient } from '../api/data-api';
 import { SessionClient } from '../api/session-api';
-import { WorkspaceContext } from '../context/WorkspaceContext';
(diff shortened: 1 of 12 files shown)


moontrip:
@dmfalke I have tested this and mostly works fine as described. One minor thing I notice is the "expand all" feature on top of the variable tree functions does not work correctly: it always expands the first Entity only (Household in this case) regardless where the selected variable is. Simply put, it does not expand all :)

bobular (approved):
Have reviewed/understood all the changes.  Apart from the SCSS which I will need to read up on.

dmfalke:
> @dmfalke I have tested this and mostly works fine as described. One minor thing I notice is the "expand all" feature on top of the variable tree functions does not work correctly: it always expands the first Entity only (Household in this case) regardless where the selected variable is. Simply put, it does not expand all :)

Thanks for the review, @moontrip. I'm aware of this issue. I should have added a note to the description. I think I want to get some UX related feedback on this. I think the "expand all" feature is incompatible with the notion of collapsing other entities, but others may have ideas on this. Feel free to make suggestions in the comments.

moontrip:
> > @dmfalke I have tested this and mostly works fine as described. One minor thing I notice is the "expand all" feature on top of the variable tree functions does not work correctly: it always expands the first Entity only (Household in this case) regardless where the selected variable is. Simply put, it does not expand all :)
> 
> Thanks for the review, @moontrip. I'm aware of this issue. I should have added a note to the description. I think I want to get some UX related feedback on this. I think the "expand all" feature is incompatible with the notion of collapsing other entities, but others may have ideas on this. Feel free to make suggestions in the comments.

Honestly, I am not quite sure whether the feature, expand all, is that important part of the tree view, especially with lots of sub-trees and variables inside: on the contrary, collapse all may be useful and it works well. This was the reason why I said 'minor thing' :)

bobular:
I should add that I didn't actually test the branch!

dmfalke:
I spoke with @aurreco-uga about the "expand all" issue. The conclusion is to only collapse other entities when navigating externally (e.g., click a box in the entity diagram).

PR #80, opened 2021-05-06
accommodating the new data response of bestFit and more

moontrip:
This commit/PR includes:

a) accommodated the new data response, bestFitLineX/Y, in the scatter plot viz: it is responded with the option of `valueSpec: bestFitLineWithRaw` post. Note that currently the `valueSpec` option is manually set, which utilizes `smoothedMeanWithRaw` for now. A plot control will be able to offer to change the `valueSpec` option through a plot/UI control.

b) added R-square value (a part of the best fit response) at plot legend texts (for now)

c) added a condition to check whether X-axis Variable is identical to X-axis Variable

d) some clean-up

a screenshot is attached here:

scatter-bestfit1

### DIFF
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -198,7 +198,7 @@ export interface ScatterplotRequestParams {
   filters: Filter[];
   config: {
     outputEntityId: string;
-    //DKDK add bestFitLineWithRaw
+    // add bestFitLineWithRaw
     valueSpec:
       | 'raw'
       | 'smoothedMean'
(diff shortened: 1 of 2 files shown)


moontrip:
@bobular thankfully reviewed scatter plot viz and found a bug, citing "It looks like there's something wrong with confidence intervals when the x-variable is negative." As of now, it seems to me that it was because of incorrect plot option, thus pushed a quick fix. Here is a screen capture of smoothedMean only (without series data) for better display after this fix.  

negative-scatter2

dmfalke (approved):

moontrip:
thank you for your review @dmfalke 👍

PR #85, opened 2021-05-12
Subsetting summary numbers

### LINKED ISSUE #76: Basic stats numbers are missing from HistogramFilter and wrong for Categorical Table/barplot
In the screenshot from the live site, we're looking at Observation -> Sample -> Stool characteristics -> Stool nature

image

Note the 60,958 number.

In the current EDA, this number is wrongly showing 22,567 - which is the number of participants.

image

It seems that 60,858 = EUPATH_0000738 "information of process" 45107 +  EUPATH_0000310 "information on observation details" 15851 (numbers come from `count` requests)

In the new EDA I guess the number we should be using comes from the `count` backend requests (used to shade the Entity Diagram).

Then we need to display the same information for the Histogram Filter

bobular:
Currently requires `web-components/barplot` (because I merged the `barplot-viz` branch into this to use some io-ts work from @moontrip)

This now uses completeCases data from the back end for histogram and table filters.  We need to decide if the API is stable with regard to this information, before merging.

Confirmed that the numbers are NOT sensitive to histogram zooming.

Fixes #76


bobular:
Created back end ticket https://github.com/VEuPathDB/EdaDataService/issues/32

bobular:
This could be good to merge, or we wait for news on the incompleteCases back end issue.

image

dmfalke:
Thanks, @bobular. Good detective work on this! I'm inclined to wait to see if the backend gets fixed. I also understand that this information will be included with the variable data in the future.

I'm wary of doing the calculation on the client -- there may be cases where we don't have all of the data in hand (e.g., when zooming in in a histogram).

bobular:
@dmfalke Danielle commented on the back end issue https://github.com/VEuPathDB/EdaDataService/issues/32#issuecomment-842382225

Seems like further discussion is required (with Steve too).

bobular:
Seems like I could have used sampleSizeTable too? It has the same numbers.

dmfalke:
Is this the relevant range to review?
https://github.com/VEuPathDB/web-eda/pull/85/files/907af88124dffb71b2f6ba23f10e78a95f7721e6..768f3d8bd5e21554438bda4c08497845a2933992

bobular:
> Is this the relevant range to review?
> https://github.com/VEuPathDB/web-eda/pull/85/files/907af88124dffb71b2f6ba23f10e78a95f7721e6..768f3d8bd5e21554438bda4c08497845a2933992

I merged in master again and the changes for this specific issue are now much easier to review.

src/lib/core/components/filter/TableFilter.tsx
```diff
@@ -318,7 +322,7 @@ export function TableFilter({
         tableSummary.value.variableId === variable.id && (
           <MembershipField
             displayName={entity.displayName}
-            dataCount={tableSummary.value.entitiesCount}
+            dataCount={totalEntityCount}
```
bobular, line 325: I'm actually wondering if the following line should be. It would seem to make sense. Hard to know for sure, because MembershipField doesn't seem to use this prop.

```
filteredDataCount={filteredEntityCount}
```

dmfalke: It should be the background count. It is used here: https://github.com/VEuPathDB/WDKClient/blob/028c0e3472b5a002ebc5a90a9bd70050011aa37d/Client/src/Components/AttributeFilter/MembershipField.jsx#L492

dmfalke (approved):

src/lib/core/components/filter/TableFilter.tsx
bobular: Sorry I was talking about line 326.  I wasn't clear.

`MembershipField.filteredDataCount` doesn't seem to be used except in [FilterList](https://github.com/VEuPathDB/WDKClient/blob/1d179fe1825dac38134273fa0e1d8de670e96f22/Client/src/Components/AttributeFilter/FilterList.jsx#L92) which is not used in EDA.

But for symmetry I think it should be `filteredDataCount={filteredEntityCount}`

dmfalke: Oh, I see. Sorry, that line didn't show up in the review section. Can you remind me why `tableSummary.value.entitiesCount` isn't good here?

bobular:
I'm not really sure why it's there, TBH.

git blame says you did it to "get filters working"

https://github.com/VEuPathDB/web-eda/blame/24a80b197e910d7ead5d066f56c80a931d6b54ee/src/lib/core/components/filter/TableFilter.tsx

dmfalke:
What I mean is, why did _you_ do this: https://github.com/VEuPathDB/web-eda/commit/c9141094fbfc9e893a7a83611d853b7c43b68d41? :slightly_smiling_face:

bobular:
Short version - to make the numbers correct like the live site.

Long version: MembershipField includes an <UnknownCount> component, which
needs

   - activeFieldState.summary.internalsCount (numerator)
   - dataCount (denominator)

UnknownCount implementation:
```
      <b>{activeFieldState.summary.internalsCount.toLocaleString()}
({toPercentage(activeFieldState.summary.internalsCount,dataCount).toLocaleString()}%)

        of {dataCount.toLocaleString()}</b> {displayName} have data for
this variable
```

dataCount needs to be `totalEntityCount`, not `tableSummary.value.
entitiesCount`

`tableSummary.value.entitiesCount` has already been passed as
`activeFieldState.summary.internalsCount`, see
https://github.com/VEuPathDB/web-eda/blob/19b25d30939a0c83eba7ec44c79cb541c6064f5d/src/lib/core/components/filter/TableFilter.tsx#L200

beforehand it was always showing X/X (100%)

dmfalke:
Thanks for the info! That all makes sense. Honestly, I get confused with
all of the different numbers.

bobular:
OK, cool.  I've made the filteredDataCount change for good measure.  I'll
merge tonight my time, if you agree.

dmfalke (approved):
All looks good. Feel free to merge.

PR #86, opened 2021-05-12
Add stats table

chowington:


chowington:
Feedback requested:

- Using `<table>`
- The method I implemented of using either the `MosaicData` or `TwoByTwoData` type in the mosaic viz component--maybe better to use `io-ts` for this?
- Table formatting

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -240,15 +248,22 @@ function MosaicViz(props: Props) {
...
-          )
-        : dataClient.getMosaic(computation.type, params as MosaicRequestParams);
-      return isTwoByTwo
-        ? twoByTwoResponseToData(await response)
-        : mosaicResponseToData(await response);
+
+      if (isTwoByTwo) {
+        const response = dataClient.getTwoByTwo(
+          computation.type,
+          params as MosaicRequestParams
```
dmfalke, line 255: I don't think this is necessary
```suggestion
          params
```

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -240,15 +248,22 @@ function MosaicViz(props: Props) {
...
+        );
+
+        return twoByTwoResponseToData(await response);
+      } else {
+        const response = dataClient.getMosaic(
+          computation.type,
+          params as MosaicRequestParams
```
dmfalke, line 262: I don't think this is necessary
```suggestion
          params
```

dmfalke (changes requested):
Hi @chowington , this is looking good. There are some unnecessary type assertions. They should be removed to prevent issues in the future.

dmfalke:
> Feedback requested:
> 
>     * Using `<table>`

I think using a table element is fine.

> 
>     * The method I implemented of using either the `MosaicData` or `TwoByTwoData` type in the mosaic viz component--maybe better to use `io-ts` for this?

I don't understand this.

> 
>     * Table formatting

I would add some padding between table cells and lighten the background a bit (maybe `#eaeaea`).

Can you also add this to the RXC viz?

src/lib/core/api/data-api.ts
```diff
@@ -321,6 +321,44 @@ export const MosaicResponse = type({
   }),
 });
 
+export type TwoByTwoResponse = TypeOf<typeof TwoByTwoResponse>;
```
bobular, line 324: I think you could reuse MosaicResponse here and union in the extra bits needed - especially since both responses start with the `mosaic` key.

Also it looks like conttable/MosaicResponse should include validation for  `sampleSizeTable` ?

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -240,15 +248,22 @@ function MosaicViz(props: Props) {
...
-        ? dataClient.getTwoByTwo(
-            computation.type,
-            params as MosaicRequestParams
-          )
-        : dataClient.getMosaic(computation.type, params as MosaicRequestParams);
-      return isTwoByTwo
-        ? twoByTwoResponseToData(await response)
-        : mosaicResponseToData(await response);
+
+      if (isTwoByTwo) {
```
bobular, line 252: Regarding your question
> The method I implemented of using either the MosaicData or TwoByTwoData type in the mosaic viz component--maybe better to use io-ts for this?

Is this where you mean?  I think it's fine to make two different requests and have data-api.ts validate them with two different validators...

bobular (commented):
I have some other comments I can't make here, and doesn't just apply to this visualization - I'll add a new issue about it.

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
chowington: Thanks Dave! Resolved.

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
chowington: Resolved

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
chowington: Yes, here and the above. Just seems like I'm doing the "either" thing io-ts is good at

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -203,7 +210,7 @@ function MosaicViz(props: Props) {
   );
 
   const data = usePromise(
-    useCallback(async (): Promise<MosaicData> => {
+    useCallback(async (): Promise<MosaicData | TwoByTwoData> => {
```
chowington, line 213: > The method I implemented of using either the MosaicData or TwoByTwoData type in the mosaic viz component--maybe better to use io-ts for this?

Here

chowington:
> 
> 
> > Feedback requested:
> > ```
> > * Using `<table>`
> > ```
> 
> I think using a table element is fine.
> 
> > ```
> > * The method I implemented of using either the `MosaicData` or `TwoByTwoData` type in the mosaic viz component--maybe better to use `io-ts` for this?
> > ```
> 
> I don't understand this.
> 
> > ```
> > * Table formatting
> > ```
> 
> I would add some padding between table cells and lighten the background a bit (maybe `#eaeaea`).
> 
> Can you also add this to the RXC viz?

Thanks Dave! I added comments to where I was referencing. I can work on the stats table for the RxC, yeah.

src/lib/core/api/data-api.ts
chowington: Thanks Bob! Updated to reuse MosaicResponse using io-ts `intersection` (not `union`, confusingly).

`sampleSizeTable` can be in a generic MosaicResponse, yes.

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
dmfalke: I agree with @bobular. This looks fine to me, as well.

dmfalke (approved):
Thanks, @chowington. These changes look good.

PR #92, opened 2021-05-19
Barplot viz

moontrip:
With the addition of a barplot component at web-components, basic barplot viz is made. Certainly, this PR requires the approve/merge the [barplot component PR](https://github.com/VEuPathDB/web-components/pull/124).

A screenshot is attached here:

barplot1

### DIFF
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -28,6 +28,19 @@ const AppsResponse = type({
 
 type ZeroToTwoVariables = [] | [Variable] | [Variable, Variable];
 
+//DKDK define sampleSizeTableArray
+const sampleSizeTableArray = array(
+  partial({
+    // set union for size as it depends on the presence of overlay variable
+    size: union([number, array(number)]),
+    overlayVariableDetails: type({
+      entityId: string,
+      variableId: string,
+      value: string,
+    }),
+  })
+);
+
 export interface HistogramRequestParams {
   studyId: string;
   filters: Filter[];
@@ -108,9 +121,15 @@ export interface BarplotRequestParams {
   //  derivedVariables:  // TO DO
   config: {
     outputEntityId: string;
-    valueSpec: 'count' | 'identity';
+    //DKDK add proportion as it seems to be coming
+    valueSpec: 'count' | 'identity' | 'proportion';
     xAxisVariable: {
       // TO DO: refactor repetition with HistogramRequestParams
+      entityId: string;
+      variableId: string;
+    };
+    //DKDK barplot add prop
+    overlayVariable?: {
       entityId: string;
       variableId: string;
     };
@@ -128,13 +147,29 @@ export const BarplotResponse = type({
       }),
     }),
     data: array(
-      type({
-        label: array(string),
-        value: array(number),
-      })
+      intersection([
+        type({
+          //DKDK label can be number too?
+          // label: array(string),
+          label: union([array(string), array(number)]),
+          value: array(number),
+        }),
+        partial({
+          overlayVariableDetails: type({
+            entityId: string,
+            variableId: string,
+            value: string,
+          }),
+          facetVariableDetails: union([
+            tuple([StringVariableValue]),
+            tuple([StringVariableValue, StringVariableValue]),
+          ]),
+        }),
+      ])
     ),
   }),
-  // TO DO: sampleSizeTable
+  //DKDK sampleSizeTable
+  sampleSizeTable: sampleSizeTableArray,
 });
 
 // scatterplot
@@ -192,18 +227,6 @@ const ScatterplotResponseData = array(
   })
 );
 
-// define sampleSizeTableArray
-const sampleSizeTableArray = array(
-  partial({
-    // set union for size as it depends on the presence of overlay variable
-    size: union([number, array(number)]),
-    overlayVariableDetails: type({
-      entityId: string,
-      variableId: string,
-      value: string,
-    }),
-  })
-);
 export type ScatterplotResponse = TypeOf<typeof ScatterplotResponse>;
 export const ScatterplotResponse = type({
   scatterplot: type({
--- a/src/lib/core/components/filter/TableFilter.tsx
+++ b/src/lib/core/components/filter/TableFilter.tsx
@@ -93,17 +93,22 @@ export function TableFilter({
           });
         }
       );
+      //DKDK by setting union for label (array number or string), ts error occurs
       const fgValueByLabel = Object.fromEntries(
-        distribution.foreground.barplot.data[0].label.map((label, index) => [
-          label,
-          distribution.foreground.barplot.data[0].value[index] ?? 0,
-        ])
+        distribution.foreground.barplot.data[0].label.map(
+          (label: string | number, index: number) => [
+            label,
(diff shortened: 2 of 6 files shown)


moontrip:
@bobular @dmfalke Somehow slack is down so I just pushed a new commit here with the change of default barLayout from overlay to group. With the same variables shown by Bob, here is the screenshot with group:

barplot-overlaygroup

moontrip:
I could also confirm that the latest change by Danielle (or someone else) to add `proportion` option (for config.valueSpec prop in Post), in conjunction with `count`, works - shows from 0 to 1 in y-axis with the `proportion` option

moontrip:
a new commit is made for adopting variable constraints

moontrip:
empty data (or data error) with empty plot is implemented. Screenshots of full screen and grid views are attached:

barplot-empty1
barplot-empty

src/lib/core/api/data-api.ts
```diff
@@ -128,13 +159,29 @@ export const BarplotResponse = type({
       }),
     }),
     data: array(
-      type({
-        label: array(string),
-        value: array(number),
-      })
+      intersection([
+        type({
+          //DKDK label can be number too?
```
bobular, line 164: The [API Spec](https://veupathdb.github.io/EdaDataService/api.html#resources:/apps/pass/visualizations/barplot:post) says label is an array of string.

What discussions have you had with @d-callan about this?  I see that the labels of variables such as `GEMS->Participant->Enrollment year` are being returned as numbers.

Should the API specs be changed or should the service return strings?

I'm just making sure everyone has had a chance to discuss this.

src/lib/core/components/filter/TableFilter.tsx
```diff
@@ -93,17 +93,22 @@ export function TableFilter({
...
         }
       );
+      //DKDK by setting union for label (array number or string), ts error occurs
       const fgValueByLabel = Object.fromEntries(
-        distribution.foreground.barplot.data[0].label.map((label, index) => [
-          label,
-          distribution.foreground.barplot.data[0].value[index] ?? 0,
-        ])
+        distribution.foreground.barplot.data[0].label.map(
+          (label: string | number, index: number) => [
```
bobular, line 99: See the comment on data-api.ts - **if** we roll that back, don't forget to roll this back.

bobular (changes requested):
Looks great DK.
See the questions about validation of `label: array(string)`

Also, in FF and Chrome, I see the page footer hidden underneath the plot.
Please also tweak the thumbnail plot size for grid view - it looks to me a bit off-centre and could fill up the box more?

src/lib/core/api/data-api.ts
moontrip: @bobular, first of all, thank you for your review 👍 

As for this issue, I know that the API doc shows the Label as string/string[], however, from my limited tests, it always returned number/number[]. I did not specifically ask Danielle about this as there have been many discrepancies between the doc and the return. Thus, I used union type instead, but we should ask Danielle about this as well.

src/lib/core/components/filter/TableFilter.tsx
moontrip: @bobular as you said, this is related to above. If we can confirm with Danielle whether the label is either string[] or number[], then I can use single type for label in there (and at data-api.ts).

moontrip:
@bobular 

> Also, in FF and Chrome, I see the page footer hidden underneath the plot.

**DK:** I am sorry that I cannot understand the issue. The page footer you meant is probably the bottom banner with VEuPathDB icons, but it shows me correctly for both FF and Chrome. Here is a screenshot at FF (I have reduced the browser size so that the captured image is not that big). Please correct me if I misunderstood your point.

barplot-ff2

> Please also tweak the thumbnail plot size for grid view - it looks to me a bit off-centre and could fill up the box more?

**DK:** I got your point. It is because I have set margin based on empty plot with X/Y ticklabels: you have changed them without ticklabels: will adjust it. 
By the way, since we are subject to thumbnail/gridview, I want to ask you about one thing. For barplot, there are many cases that X tick labels are quite lengthy (e.g., see my screenshot for X - breastfead & overlay - country). So I have been thinking that I should hide (at least) X tick labels for barplot's grid view. If that is the case, I need to introduce additional prop to show/hide axis tick label. What do you think?

barplot-gridview1

bobular:
Hi DK.
My replies in reverse order...

Yes, good idea to turn off X (and yes, I guess Y too) labels in thumbnail view.

Actually, after my changes, the empty plots do have ticklabels - but the font color is transparent. So it shouldn't affect spacing and layout?

Regarding the footer layout issue, here's a screenshot from my Linux FF:

image

It seems that if I had a taller screen, the problem would go away.  The footer appears at the bottom of the browser window, and then the barplot is drawn over the top of it.  Histogram and Scatter don't do this. (Haven't checked Mosaic!)

moontrip:
Thank you for your response, @bobular. Actually I cannot reproduce your case at my FF and Chrome (both at linux/Ubuntu VB). I tried to make very small height of browser, but it always show browser's scrollbar to separate (see my screenshot below). 
I will also add show/hide tick label props for both indep. and dep. axis.

barplot-scroll

dmfalke:
We should definitely clarify the API with Danielle. I haven’t looked at the
footer issue yet. Otherwise this is looking good!

moontrip:
@dmfalke Thank you for your review 👍 

By the way, may I ask one thing? If we have multiple reviewers (like this PR having three three reviewers), does a merge need all approvals from all reviewers or it can be made by one approval (from one reviewer) ? Actually similar situation happens on my PR regarding [barplot component PR](https://github.com/VEuPathDB/web-components/pull/124) at web-component repo (here): Bob approved the PR but not quite sure if I need both you and Jamie's approvals for merge as well.

dmfalke:
I think that should be the policy. Sorry for the confusion and delay.

bobular:
One thing (by email) - is the big Barplot header still there?  The other
plots don't have it.  Could you remove it please?

moontrip:
Good evening, Bob

I missed that! Thank you for finding. Will remove it.

moontrip:
Update:
a) implemented to hide axis tick labels for grid view/thumbnail (through [additional props](https://github.com/VEuPathDB/web-eda/pull/92/commits/acd6d37cefd635782693b52f558f57a9ab0416ab) at bar plot component); b) adjusted plot size and margin for grid view/thumbnail; c) removed big "Bar plot" header at full screen view (reported by Bob)

dmfalke (approved):
I'm okay with this being merged, as long as we make an issue in this repo to update the response type once we get clarity the intended type. Waiting is also fine. Whichever is more comfortable for you, @moontrip. Thanks!

moontrip:
> I'm okay with this being merged, as long as we make an issue in this repo to update the response type once we get clarity the intended type. Waiting is also fine. Whichever is more comfortable for you, @moontrip. Thanks!

Thank you for your review and comment @dmfalke 👍 Since the type change can be easily addressed later after merge and I don't know when Danielle will change it, it may be better to merge this and be introduced at feature site for others to test it, potentially revealing/improving something else we have missed.

jtlong3rd (approved):

bobular (approved):

PR #93, opened 2021-05-20
updates on scatter plot viz and InputVariables component

moontrip:
For InputVariables component: removed border and text (Data Inputs)

Note that this requires the merge of [the corresponding PR at web-components](https://github.com/VEuPathDB/web-components/pull/125) 

For scatter plot viz:
a) adjusted plot control to the button text with spaces
b) made colors up to 10: previously only two colors were manually set, which caused some issue with overlay variable more than 2 stratifications. Those colors are taken from plotly's default colors - screenshot attached
c) some minor changes

scatterplotviz1

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "dependencies": {
     "@material-ui/core": "^4.11.3",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.2.38",
+    "@veupathdb/components": "^0.3.0",
     "@veupathdb/wdk-client": "^0.1.2",
     "@veupathdb/web-common": "^0.1.2",
     "debounce-promise": "^3.1.2",
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -167,18 +167,17 @@ export interface ScatterplotRequestParams {
 }
 
 // unlike API doc, data (response) shows seriesX, seriesY, smoothedMeanX, smoothedMeanY, smoothedMeanSE
-const ScatterplotResponseData = array(
+export const ScatterplotResponseData = array(
   partial({
     // valueSpec = smoothedMean only returns smoothedMean data (no series data)
     //DKDK changed to string array
     seriesX: array(string),
     seriesY: array(string),
-    // smoothedMeanX: array(string),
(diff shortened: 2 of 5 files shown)


bobular (commented):
Looks good DK - please merge after removing commented out lines of code and the DKDK comments.

moontrip:
> Looks good DK - please merge after removing the DKDK comments.

@bobular Thank you for your review. Yep I had a plan to clean things up :)

bobular:
No need to have just the 'Data" legend item - skip the legend altogether in this case?
image

moontrip:
> No need to have just the 'Data" legend item - skip the legend altogether in this case?
> image

@bobular actually it was already addressed in this PR following your approach at histogram like disabling legend for single dataset, i.e., raw data without overlay 

https://github.com/VEuPathDB/web-eda/blob/2b88d0ed209e3fede57a8b7cacf461d69ae68365/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L415

moontrip:
adopted data API changes for scatter plot for now. However, I realized that the API doc had a couple of incorrect info based on the actual response from the backend. I have discussed with Danielle and she said there were incorrectness at both the doc and the backend. Accordingly, additional changes would be required in the near future.

moontrip:
Adopted additional data API change due to incorrect responses from backend. Also implemented empty plot and plot control for no data or data error. Screenshots are attached

scatter-empty2
scatter-empty1

moontrip:
scatter plot component at web-component repo is renamed to XYPlot, thus scatter plot viz is also changed to use the new component name, XYPlot

moontrip:
With the merge of corresponding [PR](https://github.com/VEuPathDB/web-components/pull/125) at web-component repo, @veupathdb/components version is update and scatterplot viz is changed to use XYPlotControls

bobular (approved):
Looks good to merge :-)

My only suggestion (for the future) is that we shouldn't have colour palette code in each visualization - it needs to be centralised somehow.

Thanks!

moontrip:
Many thanks for your review, @bobular 👍 I will merge this PR.

I agree with your opinion concerning the color palette. Perhaps we can load it at each plot viz if it is stored at a central place as you suggested :)

PR #111, opened 2021-06-01
Remember last visited variable or visualization

dmfalke:


src/lib/workspace/SessionPanel.tsx
```diff
@@ -36,6 +37,18 @@ export function SessionPanel(props: Props) {
   const filteredEntities = uniq(
     sessionState.session?.filters.map((f) => f.entityId)
   );
+  const location = useLocation();
+  const [lastVarPath, setLastVarPath] = useState('');
+  const [lastVizPath, setLastVizPath] = useState('');
+  useEffect(() => {
+    console.log({ location });
```
jtlong3rd, line 44: Were you planning to delete this `console.log` before merging, or keep it around for now?

jtlong3rd (approved):
Looks good to me! I had one inline comment.

src/lib/workspace/SessionPanel.tsx
dmfalke: I was planning to delete, thanks!

src/lib/workspace/SessionPanel.tsx
```diff
@@ -57,13 +69,22 @@ export function SessionPanel(props: Props) {
               deleteSession={deleteSession}
             />
             <Route
-              path={[`${routeBase}/variables/:entityId?`, `${routeBase}`]}
-              render={(props: RouteComponentProps<{ entityId?: string }>) => (
+              path={[
+                `${routeBase}/variables/:entityId?/:variableId?`,
+                `${routeBase}`,
```
bobular, line 74: Second element could be just `routeBase` - but there is also some nice symmetry to the way it is written now.

bobular (approved):
Tested functionally - can't find any holes!

Much more elegant than the solution I was going to try ;-) 

It is a shame that the `history` from `useHistory()` isn't searchable though.

src/lib/workspace/SessionPanel.tsx
dmfalke: I decided to just leave it. It's doesn't really hurt much :slightly_smiling_face:

Issue #116, opened 2021-06-02
Visualization: Remove pink warning "please select a variable" for new plots


bobular:
No need to save the space for errors

moontrip:
in this regard, I will work for scatter and bar plots

PR #119, opened 2021-06-02
remove error message and add 95% text for CI legend

moontrip:
These are related to issues [#115](https://github.com/VEuPathDB/web-eda/issues/115) and [#116](https://github.com/VEuPathDB/web-eda/issues/116) for scatter and/or bar plot viz. The legend with 95% confidence interval is shown in the following screenshot. Some clean-ups are done as well.

CI95


src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -149,22 +144,11 @@ function BarplotViz(props: Props) {
...
-        vizConfig.overlayVariable != null &&
-        overlayVariable != null &&
-        !isTableVariable(overlayVariable)
-      )
-        return Promise.reject(
-          new Error(
-            `'${overlayVariable.displayName}' is not suitable for this plot. Only categorical, binary, or ordinal type is allowed for the Overlay variable.`
-          )
-        );
+        return Promise.reject();
```
dmfalke, line 149: I am very wary of this. I think it would be better to just return `undefined`. This applies to all such cases.

As a side note, I would prefer to see errors thrown in async functions, rather than returning a rejected Promise.

dmfalke (changes requested):
I would like to see two changes, one relevant to this PR, and one a more general change.

First, we should not use an empty `Promise.reject()`. In this case, we should just return `undefined` and handle that case in the calling code.

Second, within an async function, we should throw errors, rather than return a rejected Promise. Part of the rationale for this is that `throw` statements are easier to scan for than `return Promise.reject(...)` expressions.

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
moontrip: @dmfalke thank you for your review and comments 👍 Actually I was gauging to use either empty `promise.reject()` or `undefined` as both worked fine for the purpose of not showing error message. And it makes sense to me for your second point as well.

changed to use `undefined` for both scatter and barplot vizs

moontrip: If the works at scatterplot and barplot vizs are acceptable, then I can probably do the same thing for other vizs like histogram and mosaic plot ones on the behalf of Bob and Connor.

dmfalke: I think this looks good. Thanks!

moontrip: @dmfalke Thank you for your review. Then I will work for other vizs as well as it may not be that difficult :)

moontrip:
@dmfalke I have worked for histogram and mosaic plot vizs not to show error message for variable selection. I noticed that the mosaic viz needs to check the case of X = Y variables so I leave that part for throwing error.

dmfalke (approved):

Issue #47, opened 2021-04-14
Visualization: Box / Beeswarm plot


dmfalke:
**API:** https://veupathdb.github.io/EdaDataService/api.html#resources:/apps/pass/visualizations/boxplot:post
**Plot component:** https://veupathdb.github.io/web-components/?path=/story/plots-box--basic

moontrip:
just for note to me: there are several updates that also affected viz component. Accordingly, this boxplot viz also needs to be updated. This is a record what should be done later (will keep updated): reference to `scatterplotvisualization.tsx` 

- https://github.com/VEuPathDB/web-eda/commit/1fe17d661fbb8f8319d4339f555a797744c20af5#diff-0c42189362f7c0d661b89bf7a2a60fbe4cebaf4258184514f1c8a4117d302cf6

- https://github.com/VEuPathDB/web-eda/commit/c2581bdf4d524e085b05bd76fe440978be012fec#diff-e51b150e769090d14ad16bf64584cca0bf8acda397695949448cb557cc620263

- https://github.com/VEuPathDB/web-eda/pull/131/files#diff-e51b150e769090d14ad16bf64584cca0bf8acda397695949448cb557cc620263

moontrip:
another updates concerning cleanup etc.

https://github.com/VEuPathDB/web-eda/commit/0f8fa33339640ff3bcb0afcdc4ab6d7ee378f35d#diff-e51b150e769090d14ad16bf64584cca0bf8acda397695949448cb557cc620263

https://github.com/VEuPathDB/web-eda/commit/4d4fe1b7dd9e4742041de1f5b255c60e5a73df97#diff-e51b150e769090d14ad16bf64584cca0bf8acda397695949448cb557cc620263

moontrip:
@bobular suggested a refactoring to use quartiles etc. instead of raw data in order to deal with many data: discussed at PR; https://github.com/VEuPathDB/web-eda/pull/120#issuecomment-859941053

PR #120, opened 2021-06-02
Boxplot viz

moontrip:
This viz relies on [the new PR at web-component repo](https://github.com/VEuPathDB/web-components/pull/138). A screenshot is attached in the following.

Please note that this viz is functional, but just made it as draft PR as this is probably not included in the Gates demo.

boxplot1


moontrip:
I noted at the [corresponding issue](https://github.com/VEuPathDB/web-eda/issues/47#issuecomment-855244571) with the requirement of some updates. They are now applied to this boxplot viz, thus this PR would work with yarn link (to web-component repo).

bobular:
Hi @moontrip 

So it turns out that the original Boxplot.tsx *is* what we need for the EDA. Allow me to explain.

We can and will have datasets with millions of data points.  In that case it's better to calculate the mean/median q1, q3 etc server-side and only transmit those plus the outliers (which will never be in the millions).

The old Boxplot.tsx *can* show all the raw data - but I think the default first back-end request by the boxplot viz would be for config.points: 'outliers'.  If the total number of datapoints (a sum over sampleSizeTable[].size[]) is less than some threshold, then allow the user to check a button/switch to send another request with config.points: 'all'.

Maybe the structure of the response and the `seriesX`, `seriesY` (no `rawData`??) means that overlay variables don't work with the approach I've described above?

Let's try to discuss on Monday?  I'll try to play with some more examples before then.

```
      // is outputEntityId correct?
      outputEntityId: xAxisVariable.entityId,
```
Answer: you'd need to double-check with @d-callan - I think both xAxisVar and yAxisVar both need to be from the same Entity (but this is not yet specified in the variable constraints? @dmfalke) - so it's currently OK :-)

moontrip:
> Hi @moontrip
> 
> So it turns out that the original Boxplot.tsx _is_ what we need for the EDA. Allow me to explain.
> 
> We can and will have datasets with millions of data points. In that case it's better to calculate the mean/median q1, q3 etc server-side and only transmit those plus the outliers (which will never be in the millions).
> 
> The old Boxplot.tsx _can_ show all the raw data - but I think the default first back-end request by the boxplot viz would be for config.points: 'outliers'. If the total number of datapoints (a sum over sampleSizeTable[].size[]) is less than some threshold, then allow the user to check a button/switch to send another request with config.points: 'all'.
> 
> Maybe the structure of the response and the `seriesX`, `seriesY` (no `rawData`??) means that overlay variables don't work with the approach I've described above?
> 
> Let's try to discuss on Monday? I'll try to play with some more examples before then.
> 
> ```
>       // is outputEntityId correct?
>       outputEntityId: xAxisVariable.entityId,
> ```
> 
> Answer: you'd need to double-check with @d-callan - I think both xAxisVar and yAxisVar both need to be from the same Entity (but this is not yet specified in the variable constraints? @dmfalke) - so it's currently OK :-)


@bobular Thank you for your review and suggestion 👍 I agree with your observations and will revisit boxplot and its viz next week: I didn't consider the case of numerous data points

For seriesX/Y, yeah the backend returns them when using the post option, `points: 'all'`, not rawData.

bobular:
Great, thanks - I opened an issue about the back end API docs

https://github.com/VEuPathDB/EdaDataService/issues/44

moontrip:
@bobular I have refactored boxplot and its viz to use quartiles etc. instead of raw data. Corresponding changes at boxplot component is in [this PR](https://github.com/VEuPathDB/web-components/pull/138/): the story file has changed alot using part of real EDA data, including overlay case: a screenshot is attached.

There seems to exist a plotly bug, which was really hard to find, but I made a way to bypass the bug.

Currently, this refactored version assumes to use the post option, `points: 'outliers'`, which means response includes outliers data only. The other option, points: 'all' is not fully addressed yet due to the known issue: responding with seriexX and Y, not rawData. If either rawData or seriesX/Y is set, then I will revisit to support `points: 'all'` option

One another issue I found is that there exists one exceptional case at least, though there may be more similar cases, `y-axis: enrollment year`. I am not quite sure if y-axis variable supports `date` type, but the enrollment year (like 2000, 2001, etc.) is set to `number` (in data type), not `date`. Unfortunately this causes an issue not to show boxplot correctly: I know how to fix this issue (setting y-axis type at plotly as `date`), but there is no generic way to know the enrollment year is `date` type, unless the data type is changed to `date`: the second screenshot shows the issue.

boxplot-breastfed

boxplot-issue

bobular (commented):
Had a quick look.  A couple of suggestions that you're probably already aware of

some `: any` types that can probably go away after #150 (e.g. using `BoxplotData` type)

vizType seems unused everywhere (a relic from scatterplot?)

moontrip:
@bobular Thank you for your review 👍 Yes, there are some `any` type which will be addressed after your PR and mine (boxplot component). 
Just made an update following your suggestions. I have removed unnecessary `vizType` . And, removed `enableOverlay` related ones as you raised in [this issue](https://github.com/VEuPathDB/web-eda/issues/183).

moontrip:
updated boxplot viz after changing corresponding box plot component. Now changed to PR from draft.

other updates are:
a) accommodating several changes for viz components (e.g., from Dave's work)
b) typing (several any cases)
c) adjust props for new box plot component, including legend title
d) changed data-api.ts to accomodate 'rawData' correctly
e) tested `points: all` option as well (returning rawData): worked fine but found a bug and discussed with Danielle to confirm it - [here is the ticket](https://github.com/VEuPathDB/plot.data/issues/51)

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -0,0 +1,403 @@
...
+          }
+          // show/hide independent/dependent axis tick label
+          showIndependentAxisTickLabel={true}
+          showDependentAxisTickLabel={true}
+          showMean={true}
+          interactive={true}
+          showSpinner={data.pending}
+          // this is required for date type
+          // not sure why enrollment year's type is number, not date
+          dependentValueType={findVariable(vizConfig.yAxisVariable)?.type}
```
bobular, line 276: I think we're going to remove all date handling from boxplot, so `dependentValueType` can probably also be removed.

Note that "enrollment year" is not really our problem.  We can add y-axis range controls so users can fix it, maybe.

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -0,0 +1,403 @@
...
+          orientation={'vertical'}
+          // orientation={'horizontal'}
+          // show/hide independent/dependent axis tick label
+          showIndependentAxisTickLabel={false}
+          showDependentAxisTickLabel={false}
+          showMean={true}
+          interactive={false}
+          displayLegend={false}
+          displayLibraryControls={false}
+          // margin={{ l: 30, r: 20, b: 0, t: 20 }}
```
bobular, line 297: Remove "version control comments" before final merge please ;-)

bobular (changes requested):
The changes requested are minimal - just removing the possibility of date handling (remove `dependentValueType` here and in Boxplot.tsx.

I am happy for us to merge this, but we need to make sure the worrying data problems (raw/outlier data in wrong order?) you discovered don't get forgotten.  So please make an issue in this repo to keep it on our radar.

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
moontrip: yeah since no 'date' data at dependentValueType is expected, the prop can be removed at viz

moontrip: deleted `dependentValueType` prop at viz

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
moontrip: removed unnecessary comments

moontrip:
boxplot viz is now up-to-date with main branch, especially updated with variable coverage table. A screenshot is attached in the following. 

boxplot-update1


I feel that the new width of 750px may be a bit small for some cases, especially with many labels (e.g., floor material) and stratifications (e.g., country), so we may need to consider this in the future. An example is attached here

boxplot-update2

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -0,0 +1,448 @@
...
+        outliers: data.outliers ? data.outliers : undefined,
+        // currently returns seriesX and seriesY for points: 'all' option
+        // it is necessary to rely on rawData (or seriesX/Y) for boxplot if points: 'all'
+        rawData: data.rawData ? data.rawData : undefined,
+        // this will be used as legend
+        name: data.overlayVariableDetails
+          ? data.overlayVariableDetails.value
+          : 'Data',
+        // this will be used as x-axis tick labels
+        label: data[response.boxplot.config.xVariableDetails.variableId],
```
bobular, line 416: I notice that the API docs and API response are not in agreement (sorry if I've missed this discussion elsewhere).

The `io-ts` validation is currently not validating this important part of the response, because the object key is different for different requests.  Here's a typical response (the third item in the reponse.boxplot.data array)  The key in this case is `EUPATH_0000096.PATO_0000047`.  

```
{
	"2": {
		"overlayVariableDetails": {
			"variableId": "PCO_0000024.EUPATH_0000730",
			"entityId": "PCO_0000024",
			"value": "Less frequent than a few times per week"
		},
		"EUPATH_0000096.PATO_0000047": [
			"Female",
			"Male"
		],
		"min": [
			0,
			1
		],
		"q1": [
			5,
			8
		],
		"median": [
			10,
			14
		],
               // and so on
	}
}
```

Should we go ahead with this PR as is, and make a ticket that the response should really have an array of strings something like this

```
boxplot.data[].xVariableValues  : array
boxplot.data[].xVariableValues[]  : string
```
?

bobular (approved):
Other than the comment about the API mismatch, this looks and works great.

One minor comment - the thumbnail version has a little too much height - compared to other thumbnails - see below.  Could you tweak it to have more bottom margin please?

image

I'm going to drop Jamie from the reviewers so we can get this merged.

Please also make a back end ticket with our suggestion for a consistently named object key in the boxplot response.

bobular:
Even if I subset heavily the outliers still have a disturbingly uniform distribution. So it's not an issue of point crowding.

image

bobular:
Maybe wait until we adopt the minor back-end change in the ticket mentioned above, before merging.

moontrip:
@bobular thank you for your reviews and comments 👍 First of all, I have changed the size of the grid view, which is identical to barplot's one: hope this would be sufficient :) 

boxplot-breastfed-gridview1

As for the outlier, honestly I didn't think of your case much as I primarily used other variables like Breastfed vs Age at eligibility... like the following screenshot. This looked okay to me. 

boxplot-breastfed1

Anyway, when I briefly checked your case, Sex vs Age at eligibility... I could see that the total data points are 22,567, while the numbers of outliers are 177 (Female) and 320 (Male). Note that all quartile values are shown correctly with the backend response at least. I am not quite sure if it is just data itself or it may just be visually congested in the small range. Just for your information, I have tried to plot it at larger size as shown in the following: outliers are distributed from 51 to 59 (max value).

boxplot-sex1

bobular:
Thanks @moontrip 

@d-callan, what do you think about the sex vs age at eligibility plot above?  I'm concerned there's no obvious tail at higher values of age, but maybe it's just quirky data and no bug?

d-callan:
I can tell you what I suspect only, would have to dig in and maybe talk w clinepi ppl to confirm.. but I think it's just the data and possibly a quirk of the bin width used for the histo. So the range of data for age at eligibility form completion is 0-59 months, and I think they're integer values (though I didn't confirm we never see ex 1.5 months). So if you use a bin width that's not also an integer you'll get strange humps in the histo, if you use integer bin widths you'll get something more uniform and in fact more truthful. I can enter a bin width of 3 and confirm that while there is a tail, it seems from 50-59 months there's not a significant decline in counts over that range. So I'm inclined to trust the boxplot, currently.

d-callan:
If anything, we could file an issue against the histo for integer bin widths when we have integer values

bobular:
> If anything, we could file an issue against the histo for integer bin widths when we have integer values

Thanks for your investigation - I'd have never thought of non-integer bin widths causing a problem.

Is the issue you propose intended for the back end?  If it's feasible to do there, then I'd agree with raising a ticket.  If it's front end, let me know.

@moontrip - I think you can merge now :tada:

d-callan:
Yes. I suppose it is a backend issue. Something for plot.data to add to its binning and bin slider logic

moontrip:
@bobular Thank you for your reviews and comments 👍 I asked Jamie to review the corresponding boxplot component PR and he reviewed and approved this morning. So, I have updated web-component version and now merged

PR #125, opened 2021-06-03
session => analysis

dmfalke:
Rename "session" to "analysis".

### DIFF
--- a/src/lib/mapveu/Mocks.ts
+++ b/src/lib/mapveu/Mocks.ts
@@ -1,47 +1,47 @@
-import { Session, NewSession } from '../core';
+import { Analysis, NewAnalysis } from '../core';
 import localforage from 'localforage';
 import { pipe } from 'fp-ts/lib/function';
-import { SessionClient } from '../core/api/session-api';
+import { AnalysisClient } from '../core/api/analysis-api';
 import { getOrElse, isLeft, isRight } from 'fp-ts/lib/Either';
 import { PathReporter } from 'io-ts/lib/PathReporter';
 
 const localStore = localforage.createInstance({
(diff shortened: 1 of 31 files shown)


jtlong3rd (changes requested):
Almost good to merge! You'll want to address a few cases where `sessions` was replaced with `analysiss` instead of `analyses`.

src/lib/mapveu/Mocks.ts
```diff
@@ -1,47 +1,47 @@
...
 const localStore = localforage.createInstance({
-  name: 'mockSessionStore',
+  name: 'mockAnalysisStore',
 });
 
-export const mockSessionStore: SessionClient = {
-  async getSessions() {
-    const records: Session[] = [];
+export const mockAnalysisStore: AnalysisClient = {
+  async getAnalysiss() {
```
jtlong3rd, line 13: Here...

src/lib/workspace/Mocks.ts
```diff
@@ -1,45 +1,45 @@
...
 const localStore = localforage.createInstance({
-  name: 'mockSessionStore',
+  name: 'mockAnalysisStore',
 });
 
-export const mockSessionStore: SessionClient = {
-  async getSessions() {
-    const records: Session[] = [];
+export const mockAnalysisStore: AnalysisClient = {
+  async getAnalysiss() {
```
jtlong3rd, line 11: ... and here!

jtlong3rd (changes requested):
Almost there! There are two lingering `analysiss`es.

dmfalke:
Oops! Fixed.

jtlong3rd (approved):

PR #131, opened 2021-06-04
My Variables

### LINKED ISSUE #103: My variables feature
The following comprise this feature
- Add a star button to variables in the variable tree
- Add a toggle to show only starred variables

The list of starred variables is held in the session state. Most of the wiring is done for this.

jtlong3rd:
Resolves #103.


src/lib/core/components/VariableList.tsx
```diff
@@ -159,8 +172,29 @@ export default function VariableList(props: VariableListProps) {
...
   );
 
+  const availableVariables = useMemo(() => {
+    const availableVariablesArray = getLeaves(
+      fieldTree,
+      (node) => node.children
+    ).map((node) => node.field.term.split('/')[1]);
+
+    return new Set(availableVariablesArray);
+  }, [fieldTree]);
```
dmfalke, line 182: We shouldn't use `getLeaves` here. There are cases where variables can have children. The [`isFilterField`](https://github.com/VEuPathDB/WDKClient/blob/master/Client/src/Components/AttributeFilter/AttributeFilterUtils.ts#L30) function can be used to find all variables.

dmfalke (changes requested):
This looks great. I have one requested change (see inline comment).

src/lib/core/components/VariableList.tsx
jtlong3rd: Good catch! PR updated.

dmfalke (approved):

bobular (commented):
Think this got merged while I was testing and reviewing.

Functionally no probs (copied an Analysis and the starred vars were copied too)

Read through and understood the code.

Nice job!

PR #162, opened 2021-06-08
Remove 'variable' wordage from InputVariables area

### LINKED ISSUE #153: Visualization: Remove the word "variable" from labels for variable selector

bobular:
Closes #153 

I have done as requested, but I also suggest we wrap the controls in a LabelledGroup box (currently implemented for Histogram only).

The layout is a bit wrong - InputVariables would need less padding and LabelledGroup doesn't currently work nicely with a `marginTop` provided via `containerStyles` prop.

image

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
@@ -180,15 +180,15 @@ function BarplotViz(props: Props) {
             inputs={[
               {
                 name: 'xAxisVariable',
-                label: 'Main variable',
+                label: 'Main',
               },
               {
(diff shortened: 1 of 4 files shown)


dmfalke:
Thanks @bobular. We recently removed the border box in an attempt to make the UI a little cleaner. I think we should stick to losing it for now.

bobular:
> Thanks @bobular. We recently removed the border box in an attempt to make the UI a little cleaner. I think we should stick to losing it for now.

Yeah - it was more about the "Variables" label (in the absence of "Main variable", "Overlay variable" etc), but if every drop-down says "Select a variable" then it's pretty clear what this is for ;-)

Issue #154, opened 2021-06-08
Visualization: Add viz type displayName as subtitle of visualization


dmfalke:
This is in the full-screen mode. It should be added underneath the editable name

E.g.:

**Unnamed Scatter plot**
*Scatter plot*

moontrip:
is this acceptable? I thought the `Unnamed Scatter Plot` is bold and `Scatter plot` is italic as you described.

![scatter-subtitle](https://user-images.githubusercontent.com/12802305/121221496-9efaa800-c853-11eb-9711-e75c9076d35e.png)

dmfalke:
I think that looks pretty good. I'd like to see the subtitle a little bit more gray. Maybe something like #333333.

PR #164, opened 2021-06-08
adding and styling viz title and subtitle

moontrip:
viz subtitle indicating plot, e.g., Scatter plot, is added and title and subtitle are styled according to the example shown by @dmfalke [here](https://github.com/VEuPathDB/web-eda/issues/154)

Here is a screenshot: text color for subtitle is #333333

scatter-subtitle-333333

### DIFF
--- a/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -315,7 +315,7 @@ function FullScreenVisualization(props: Props & { id: string }) {
             <SaveableTextEditor
               value={
                 viz.displayName ??
-                'Unnamed ' + overview?.displayName ??
+                'unnamed ' + overview?.displayName ??
                 'visualization'
               }
               onSave={(value) =>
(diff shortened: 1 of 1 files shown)


jtlong3rd (approved):

jtlong3rd (approved):

moontrip:
well sorry @jtlong3rd to make this review again and again. Some were my mistakes and the other one was added by automatic styling :( Thanks 👍

dmfalke (changes requested):
When I recommended the current styling, I had forgotten about the editable analysis name at the top of the page. It is italicized and not bold. I think we should be consistent. ~Could you also make that non-italicized and bold?~ Thanks!

After another think, I think we should actually keep the analysis name as-is, and change the viz name back to that style. The subtitle can be non-italicized.

dmfalke:
image

moontrip:
@dmfalke I have made changes: actually removed some unnecessary styling :)

scatter-subtitle2

jtlong3rd (approved):

dmfalke (approved):

PR #165, opened 2021-06-08
Scatterplot viz: remove SmoothedMean option and update package.json

moontrip:
SmoothedMean option is removed from scatter plot, thus corresponding changes are made. Corresponding update at XYPlotControls are already approved and [merged](https://github.com/VEuPathDB/web-components/pull/147), thus package.json is also updated with the new web-components version. It looks like this at viz: backend seems down now so just captured empty plot with control

scatter-control2

### DIFF
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -363,7 +363,6 @@ function ScatterplotViz(props: Props) {
               labelPlacement={'end'}
               // minWidth is used to set equivalent space per item
               minWidth={210}
-              // minWidth={75}
               buttonColor={'primary'}
               margins={['0', '0', '0', '5em']}
             />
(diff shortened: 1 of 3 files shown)


src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -363,6 +363,7 @@ function ScatterplotViz(props: Props) {
               labelPlacement={'end'}
               // minWidth is used to set equivalent space per item
               minWidth={210}
+              // minWidth={75}
```
dmfalke, line 366: I think this comment can be removed

dmfalke (approved):

jtlong3rd (approved):

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: @dmfalke yep. I tested the spacing but forgot to remove it - removed

moontrip:
thank you for your quick reviews, @dmfalke and @jtlong3rd 👍

PR #166, opened 2021-06-09
Cleanup

dmfalke:
Address eslint issues

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
@@ -125,7 +125,7 @@ function BarplotViz(props: Props) {
         facetVariable,
       });
     },
-    [updateVizConfig, vizConfig]
+    [updateVizConfig]
   );
 
   const findVariable = useCallback(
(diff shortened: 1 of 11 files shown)


src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -348,8 +350,10 @@ export function barplotResponseToData(
     })),
   };
 }
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
jtlong3rd, line 353: Can this line be deleted?

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -186,22 +191,23 @@ function HistogramViz(props: Props) {
         filters ?? [],
         vizConfig.xAxisVariable,
         xAxisVariable.type,
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
jtlong3rd, line 194: Can this line be deleted? It looks like you addressed the unexhaustive dependencies below.

jtlong3rd (changes requested):
Thanks for doing this! I've pointed out two places where disabling the linter may not be necessary, but otherwise, looks good to merge.

jtlong3rd (approved):

PR #178, opened 2021-06-15
Prevent page from jumping to top when switching tabs

### LINKED ISSUE #63: General: Switching between Browse and Viz tabs causes page to scroll to top in browser

dmfalke:
Fixes #63. This was primarily resolved by https://github.com/VEuPathDB/WDKClient/pull/133. All workspace links were also updated to "replace" the url in browser history, rather than "push" a new entry. We can debate the merits of replace vs push in the future :slightly_smiling_face: - the objective here was to have consistency. Finally, the analysis container was given a `min-height` of `100vh`. This further prevents jumping to the top of the screen.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
   "name": "@veupathdb/eda",
-  "version": "0.3.25",
+  "version": "0.3.29",
   "dependencies": {
     "@material-ui/core": "^4.11.3",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.3.11",
-    "@veupathdb/wdk-client": "^0.1.3",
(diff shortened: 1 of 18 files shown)


jtlong3rd (approved):

bobular (commented):
I looked up the React router Link.replace prop does and it replaces the history item rather than pushes another location on the stack.

This removes browser forward/back navigation from our app.

Do we really want to do this?

(Sorry, just read the comments on the Issue ticket and now appreciate this is a tricky one (well out of my area!))

dmfalke:
@bobular, I think we can debate which way to go separately. For now, using `replace` is more consistent. We should probably discuss in a future meeting.

bobular (approved):

PR #180, opened 2021-06-16
add X-axis offset for scatter plot viz

moontrip:
added X-axis offset for scatter plot viz. Discussed with 5% offset but it seems to be a bit large for some cases, thus I have set it to 2%. Two screenshots, which are based on Steve's screenshot at Slack (GEMS1 Case Control), are attached, without and with offset.

scatter-xaxis-without-offset

scatter-xaxis-offset


jtlong3rd (approved):

dmfalke (approved):

bobular (approved):

moontrip:
Thank you for your quick reviews @jtlong3rd @dmfalke and @bobular 👍

PR #184, opened 2021-06-22
Use Material UI tooltips

dmfalke:
This PR utilizes tooltips from the Material UI library. We may want to use these tooltips in other parts of the EDA workspace, such as places where we are currently relying on system tooltips (via the `title` attributes).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "dependencies": {
     "@material-ui/core": "^4.11.3",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.3.11",
+    "@veupathdb/components": "^0.3.14",
     "@veupathdb/wdk-client": "^0.1.4",
     "@veupathdb/web-common": "^0.1.2",
     "debounce-promise": "^3.1.2",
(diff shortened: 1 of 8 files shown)


dmfalke:
I'm looking for feedback on whether we should use the `Tooltip` component in more places, or stick with the current set of changes.

jtlong3rd (approved):
I like the new tooltips! I particularly appreciate how easy it is to make them "interactive."

I'm fine with `Tooltip` / `HtmlTooltip` being adopted in more places than just your current set of changes.

dmfalke:
I think the proper thing to do would be to add `HtmlTooltip` to https://github.com/veupathdb/web-components.

jtlong3rd:
> I think the proper thing to do would be to add `HtmlTooltip` to https://github.com/veupathdb/web-components.

Makes sense!

dmfalke:
One issue with using mui tooltips is that we are using some components from `wdk-client` that use a different tooltip component. This might be a good opportunity to move those components to `web-components` (but there may be obstacles with that...).

dmfalke:
Or, we can update wdk-client to use mui tooltips.

jtlong3rd (approved):

PR #189, opened 2021-06-28
Update variables type

dmfalke:
The EdaDataService API for **StudyVariable** is being updated. This PR incorporates that changes.

There are two primary changes to the **StudyVariable** type:
1. Variables have been factored into **StudyVariableCategory** and **StudyVariableVariable** types.
2. The **StudyVariableVariable** type has additional properties which will be used for variable UI/UX enhancements.

_Note: the names of these types could use improvement, but I'm also okay with keeping them as-is._

### DIFF
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -18,7 +18,7 @@ import {
 } from 'io-ts';
 import { Filter } from '../types/filter';
 import { TimeUnit } from '../types/general';
-import { Variable, StringVariableValue } from '../types/variable';
+import { VariableDescriptor, StringVariableValue } from '../types/variable';
 import { ComputationAppOverview } from '../types/visualization';
 import { ioTransformer } from './ioTransformer';
 
@@ -26,7 +26,10 @@ const AppsResponse = type({
   apps: array(ComputationAppOverview),
 });
 
-type ZeroToTwoVariables = [] | [Variable] | [Variable, Variable];
+type ZeroToTwoVariables =
+  | []
+  | [VariableDescriptor]
+  | [VariableDescriptor, VariableDescriptor];
 
 // define sampleSizeTableArray
 const sampleSizeTableArray = array(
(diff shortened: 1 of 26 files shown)


dmfalke:
Note, you will need to have a service running on branch code to test this code out. I'm happy to provide instructions.

jtlong3rd (approved):

dmfalke:
I just discovered that I missed some variable type specific properties. This will require some more factoring. I will update with those changes soon.

jtlong3rd (approved):

bobular (approved):
I haven't been able to test it via dfalke-b yet but have had a good look at the code.

The `StudyVariableXXX` naming thing has actually disturbed me and I've been thinking of alternatives...

The existing `StudyVariable` is essentially `StudyVariableTreeNode` which has two types of node: `StudyVariableTreeCategory` (internal) and `StudyVariableTreeVariable` (leaf)

Options include dropping the `Study` prefix - because really these are entity variable trees (there's one tree per entity, I think).   And maybe shortening `VariableTree` to `VarTree` ??

dmfalke:
Thanks @bobular. I'm fine with removing the `Study` prefix. I'm not so sure about shortening `VariableTree` to `VarTree`. I kind of like being explicit. I will work on new type names and will push. We can reevaluate then.

dmfalke:
I've taken a stab at some new type names. See https://github.com/VEuPathDB/web-eda/blob/edde7ee7006805aa3a922c19800939ebf7319c92/src/lib/core/types/study.ts.

dmfalke:
Hmm... maybe `CategoryVariable` should be `CategoryNode`.

bobular:
Yes I was just about to suggest that CategoryVariable needed some love.

dmfalke:
I renamed it `VariableCategory`, in that it is a category for a variable. Thoughts?

bobular:
> I renamed it `VariableCategory`, in that it is a category for a variable. Thoughts?

Yeah - all good now, IMO.

Is dfalke-b still on the way?

dmfalke:
dfalke-b is working now, but there are some other API changes. I'm going to merge this and will address the other major API change in a separate PR.

PR #190, opened 2021-06-28
handling date string for scatter plot viz

moontrip:
This PR involves several works concerning scatter plot and its viz. Since its plot component is changed, [the work at web-components](https://github.com/VEuPathDB/web-components/pull/161) needs to be merged as well.

Changes are:
a) handled date string for scatter plot (see the first screenshot)
  : simplified the function to make plotly friendly data
  : changed to use lodash for min/max and math operations
b) addressed David's suggestion to place lines on top of markers: owing to Dave's suggestion (refer to the first screenshot)
c) added overlay variable's value to legend for smoothed mean, confidence interval, and best fit line cases for better distinction
  : by addressing b), above cases are always shown after all markers (refer to the first screenshot)
d) dynamic plot control for its viz: only show Raw option when y-axis variable is date (refer to the second screenshot)
e) bug fix for y-axis range

scatter-date1

scatter-date2

### DIFF
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -1,10 +1,7 @@
 // load scatter plot component
-import XYPlot, {
-  ScatterplotProps,
-} from '@veupathdb/components/lib/plots/XYPlot';
+import XYPlot, { XYPlotProps } from '@veupathdb/components/lib/plots/XYPlot';
 import { ErrorManagement } from '@veupathdb/components/lib/types/general';
 
-import { Loading } from '@veupathdb/wdk-client/lib/Components';
 import { preorder } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
 import { getOrElse } from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/function';
@@ -38,6 +35,7 @@ import scatter from './selectorIcons/scatter.svg';
 
 // XYPlotControls
 import XYPlotControls from '@veupathdb/components/lib/components/plotControls/XYPlotControls';
+import { min, max, lte, gte } from 'lodash';
 
 export const scatterplotVisualization: VisualizationType = {
   gridComponent: GridComponent,
@@ -70,25 +68,19 @@ function FullscreenComponent(props: VisualizationProps) {
 
 function createDefaultConfig(): ScatterplotConfig {
   return {
-    enableOverlay: true,
     valueSpecConfig: 'Raw',
   };
 }
 
 type ScatterplotConfig = t.TypeOf<typeof ScatterplotConfig>;
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-const ScatterplotConfig = t.intersection([
-  t.type({
-    enableOverlay: t.boolean,
-  }),
-  t.partial({
-    xAxisVariable: Variable,
-    yAxisVariable: Variable,
-    overlayVariable: Variable,
-    facetVariable: Variable,
-    valueSpecConfig: t.string,
-  }),
-]);
+const ScatterplotConfig = t.partial({
+  xAxisVariable: Variable,
+  yAxisVariable: Variable,
+  overlayVariable: Variable,
+  facetVariable: Variable,
+  valueSpecConfig: t.string,
+});
 
 type Props = VisualizationProps & {
   fullscreen: boolean;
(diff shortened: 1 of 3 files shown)


moontrip:
This is now updated with tidy-up works by Bob. Details are listed in the following:

a) updated scatter plot and its viz to be compatible with Bob's tidy-up work
  - resolved merge conflicts, adjusting/changing/removing props, changing viz to be compatible with tidy-up work, etc.
b) fixed plot control bug for y-axis variable = date
  - always to be Raw regardless of previously selected option
c) used RadioButtonGroup directly instead of XYPlotControl component
d) introduced new prop to hide (grayed out) plot options when y-axis variable = date
  - only Raw option is active (screenshot)

Though, probably need to do another merge with Jamie's variable coverage table work.

scatter-date4

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -172,17 +205,37 @@ function ScatterplotViz(props: Props) {
     [updateVizConfig]
   );
 
+  // set valueSpec as Raw when yAxisVariable = date
+  useEffect(() => {
```
jtlong3rd, line 209: (To be addressed after Variable Coverage is merged.)

The linter is warning about missing dependencies for this effect:

```
Line 215:6:  React Hook useEffect has missing dependencies: 'findVariable' and 'updateVizConfig'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```

Simply adding `findVariable` and `updateVizConfig` to the dependency list isn't ideal, as that could cause `updateVizConfig({ valueSpecConfig: 'Raw' })` to be run unnecessarily.

I propose deleting this effect and moving the "if the yAxisVariable is of type 'date', call `updateVizConfig({ valueSpecConfig: 'Raw' })`" logic to the definition of the `handleInputVariableChange` callback.

Once Variable Coverage is merged, it will be clearer how the code will need to change.

jtlong3rd (commented):
This is looking good! I have one inline comment, which you don't have to address until after the Variable Coverage work gets merged.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: @jtlong3rd thank you for review and suggestion 👍 I missed that linter message.
I tested your approach like this at `handleInputVariableChange` (will change `findVariable` to `findEntityAndVariable` after your merge):
```
      updateVizConfig({
        xAxisVariable,
        yAxisVariable,
        overlayVariable,
        facetVariable,
        valueSpecConfig: (findVariable(yAxisVariable)?.type === 'date')
          ? 'Raw'
          : vizConfig.valueSpecConfig
      });
```

Basically it works as well, however, I still get a similar error, `React Hook useCallback has missing dependencies: 'findVariable' and 'vizConfig.valueSpecConfig'...`

bobular (approved):
Functionally looks/works great and the disabledList option to the radio group looks good too.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
jtlong3rd: Yes, you're right! It's OK add those missing dependencies to the `useCallback`. Because this is a `useCallback` and not a `useEffect`, `updateVizConfig` won't actually get called until `handleInputVariableChange` gets invoked.

moontrip: @jtlong3rd Aha I misunderstood your intention 😄 Yep I will do that after your merge. Thanks again 👍

moontrip:
merged main (e.g., variable coverage table etc.) and resolved conflicts. Here are two screenshot: one is numeric and the other is yAxisVariable = date. @jtlong3rd I wonder this shows your variable coverage table correctly so can you please check it out ? 👍 

At least one more commit is required for changing package.json to accept corresponding plot component works - will do that too

scatter1
scatter2

jtlong3rd:
> merged main (e.g., variable coverage table etc.) and resolved conflicts. Here are two screenshot: one is numeric and the other is yAxisVariable = date. @jtlong3rd I wonder this shows your variable coverage table correctly so can you please check it out ? 👍
> 
> At least one more commit is required for changing package.json to accept corresponding plot component works - will do that too
> 
> scatter1
> scatter2

The placement of the Variable Coverage table looks good!

jtlong3rd (approved):

moontrip:
thank you for your reviews and comments @jtlong3rd and @bobular 👍 👍

PR #191, opened 2021-06-29
Fix EDA to work with refactored plot components and controls

bobular:
Depends on web-components PR is https://github.com/VEuPathDB/web-components/pull/150 (`plot-props-tidy`)

Nothing really revolutionary here, however, take a look at `HistogramFilter` - the nasty double use of `HistogramControls` has been replaced by `SelectedRangeControl` and `HistogramControls`.

I could similarly update `HistogramVisualization` to use its own small set of controls instead of `HistogramControls`.  I may do this after I hit 'create' on this PR. (done!)

Before this is merged, please can @moontrip fix the ScatterplotVisualization - assigning to you too thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "dependencies": {
     "@material-ui/core": "^4.11.3",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.3.14",
+    "@veupathdb/components": "^0.3.15",
     "@veupathdb/wdk-client": "^0.1.4",
     "@veupathdb/web-common": "^0.1.2",
     "debounce-promise": "^3.1.2",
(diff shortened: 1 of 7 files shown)


moontrip:
I have looked at Scatter plot viz and they looked fine. Since scatter plot and its viz are changed a lot at my PRs, thus I will have a chance to revisit them again after your tidy-up works are merged. This is the same case for box plot and its viz.

Update: I just saw your note, but you can just think that I will do additional work at my own PRs (concerning scatter and box plots and their vizs) after your tidy-up works (this and web-components ones) are merged. There would be lots of merge conflicts but I can manage them myself :)

moontrip:
@bobular I have spent more time fixing the issue on the scatter plot viz than I expected because your approach at XYPlot component was different from my recent works, which led me to investigate incorrect parts at first :( 

Anyway, just realized that for your current work, I don't necessarily change your works at plot components so just changed the viz to work it out.

dmfalke (approved):

PR #192, opened 2021-06-29
change the behavior of Visualize tab

moontrip:
Basically, the task is to behave `Visualize` (Viz) tab to point out `/pass-through` when it is clicked from a plot viz's full screen mode. Since there is a case to click `Viz` tab from `Browse and subset` tab, I have made a function to check whether the current Url string has UUID in the end (case of plot viz's full screen).

### DIFF
--- a/src/lib/workspace/AnalysisPanel.tsx
+++ b/src/lib/workspace/AnalysisPanel.tsx
@@ -42,7 +42,7 @@ export function AnalysisPanel(props: Props) {
   const location = useLocation();
   const [lastVarPath, setLastVarPath] = useState('');
   const [lastVizPath, setLastVizPath] = useState('');
-  //DKDK use relativePath as input arg: 'Browse and Subset' do not have UUID
+  // use relativePath as input arg: 'Browse and Subset' do not have UUID
   const isUUIDUrl = useCallback(
     (relativePath: string) => {
       const lastUrlElement = relativePath.split('/').pop();
@@ -118,7 +118,7 @@ export function AnalysisPanel(props: Props) {
           },
           {
             display: 'Visualize',
-            //DKDK use relativePath here to handle between subset and viz tabs
+            // use relativePath here to handle a behavior between subset and viz tabs
             route: isUUIDUrl(location.pathname.replace(routeBase, ''))
               ? '/visualizations/pass-through'
               : `/visualizations${lastVizPath}`,
--- a/src/lib/workspace/AnalysisPanel.tsx
+++ b/src/lib/workspace/AnalysisPanel.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useCallback } from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import { cx } from './Utils';
 import { AnalysisSummary } from './AnalysisSummary';
 import { EntityDiagram, Status, useAnalysis, useStudyRecord } from '../core';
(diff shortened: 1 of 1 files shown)


src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -42,6 +42,17 @@ export function AnalysisPanel(props: Props) {
   const location = useLocation();
   const [lastVarPath, setLastVarPath] = useState('');
   const [lastVizPath, setLastVizPath] = useState('');
+  // use relativePath as input arg: 'Browse and Subset' do not have UUID
+  const isUUIDUrl = useCallback(
```
jtlong3rd, line 46: Since `isUUIDUrl` is only being called once (to determine if a fullscreen visualization is being viewed), you can replace `useCallback` with `useMemo`, and rename `isUUIDUrl` to something like `inFullscreenVisualization`.

Maybe something like:

```typescript
const inFullscreenVisualization = useMemo(() => {
  const relativePath = location.pathname.replace(routeBase, '');
  const lastUrlElement = relativePath.split('/').pop();

  // 'Browse and Subset' and the 'Visualize' selector do not have a UUID in the url
  return (
    lastUrlElement?.match(
      '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'
    ) != null
  );
}, [location, routeBase]);
  ```

**Review by jtlong3rd — CHANGES_REQUESTED** [review]
The new functionality works great! I had one small inline suggestion.

File: src/lib/workspace/AnalysisPanel.tsx (hunk 1 above)
Comment (moontrip) [reply src/lib/workspace/AnalysisPanel.tsx]: Thank you for your quick review and comment @jtlong3rd 👍 I agree with your suggestion and changed to use useMemo instead of useCallback

**Review by jtlong3rd — APPROVED** [review]

**Review by dmfalke — APPROVED** [review]

**Comment by moontrip** [comment]
thank you for your reviews @jtlong3rd and @dmfalke 👍

File: src/lib/workspace/AnalysisPanel.tsx [hunk 2]
```diff
@@ -107,7 +120,10 @@ export function AnalysisPanel(props: Props) {
           },
           {
             display: 'Visualize',
-            route: `/visualizations${lastVizPath}`,
+            // check whether user is at viz's full screen mode
+            route: inFullscreenVisualization
+              ? '/visualizations/pass-through'
```
bobular, line 125: Should `/pass-through` be hardcoded here?  I don't know how to fix it, but it seems problematic!

bobular: Actually, the following seems to work just fine (presumably due to default computation redirects)

```
            route: inFullscreenVisualization
              ? '/visualizations'
              : `/visualizations${lastVizPath}`,
```

moontrip: @bobular certainly it can be done programmatically like introducing a function to have string.split('/').pop(), then join('/') but I thought it might be overkill for this case: we can add it later, if necessary

Yeah /visualizations would redirect to /visualizations/pass-through but just left it to be more clear :)

moontrip: @bobular Just for a record, this will work for a single line approach:

`location.pathname.replace(routeBase, '').substring(0, location.pathname.replace(routeBase, '').lastIndexOf('/'))`

bobular: I'm mostly worried about `pass-through` being hardcoded into the code, when we start using other "computations".

Is your string manipulation suggestion designed to remove the UUID portion of the route, and leave behind the 'pass-through'?  Maybe a regular expression replace (with empty string) would be cleaner (no need to duplicate the `replace(routeBase, '')` )

```
location.pathname.replace(routeBase, '').replace(/\/[^/]*$/, ''); // remove everything from the last '/' to the end
```

moontrip: @bobular yep it works too 👍 I made a new PR using your approach, https://github.com/VEuPathDB/web-eda/pull/193

dmfalke:
I don’t think this is the correct behavior. When we have multiple
computations, we’ll want to go to the computations selector.

I think this logic could be simplified by checking if we’re in the viz tab,
and if so, just using `${routeBase}/visualizations`.

moontrip:
@dmfalke Thank you for your comment. Yes perhaps it can be simplified as:

`route : location.pathname.replace(routeBase, '').includes('/visualizations') ? '/visualizations' : '/visualizations${lastVizPath}'`

(I don;t think we need $routeBase?)

bobular:
I don't think your last suggestion will work, @moontrip - the fullscreen visualization paths will contain `'/visualizations'`

Maybe Dave meant something more like my suggestion
https://github.com/VEuPathDB/web-eda/pull/192#discussion_r661011428

bobular:
We should probably discuss on the new PR?
https://github.com/VEuPathDB/web-eda/pull/193

moontrip:
@bobular well actually my last approach works fine as @dmfalke suggested because as I mentioned earlier, the visualize tab will direct /visualize/pass-through only if a user is at viz tab (regardless it is full screen mode)

PR #193, opened 2021-06-30
change hard-corded part of viz tab behavior

moontrip:
Following @bobular 's opinion, I have changed the hard-corded part, `/visualize/pass-through` to use a programmatic approach using regex.

### DIFF
--- a/src/lib/workspace/AnalysisPanel.tsx
+++ b/src/lib/workspace/AnalysisPanel.tsx
@@ -122,7 +122,8 @@ export function AnalysisPanel(props: Props) {
             display: 'Visualize',
             // check whether user is at viz's full screen mode
             route: inFullscreenVisualization
-              ? '/visualizations/pass-through'
+              ? // remove everything from the last '/' to the end (UUID in this case)
+                location.pathname.replace(routeBase, '').replace(/\/[^/]*$/, '')
               : `/visualizations${lastVizPath}`,
             exact: false,
           },
--- a/src/lib/workspace/AnalysisPanel.tsx
+++ b/src/lib/workspace/AnalysisPanel.tsx
@@ -42,18 +42,6 @@ export function AnalysisPanel(props: Props) {
   const location = useLocation();
   const [lastVarPath, setLastVarPath] = useState('');
   const [lastVizPath, setLastVizPath] = useState('');
-  // check whether a user is at viz's full screen mode
-  const inFullscreenVisualization = useMemo(() => {
-    const relativePath = location.pathname.replace(routeBase, '');
-    const lastUrlElement = relativePath.split('/').pop();
-
-    // 'Browse and Subset' and the 'Visualize' selector do not have a UUID in the url
-    return (
-      lastUrlElement?.match(
-        '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'
-      ) != null
-    );
-  }, [location, routeBase]);
 
   useEffect(() => {
     const relativePath = location.pathname.replace(routeBase, '');
@@ -120,10 +108,11 @@ export function AnalysisPanel(props: Props) {
           },
           {
             display: 'Visualize',
-            // check whether user is at viz's full screen mode
-            route: inFullscreenVisualization
-              ? // remove everything from the last '/' to the end (for this case, UUID)
-                location.pathname.replace(routeBase, '').replace(/\/[^/]*$/, '')
+            // check whether user is at viz
+            route: location.pathname
+              .replace(routeBase, '')
+              .includes('/visualizations')
+              ? '/visualizations'
               : `/visualizations${lastVizPath}`,
             exact: false,
           },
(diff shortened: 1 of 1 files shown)


moontrip:
@jtlong3rd Is it better to use `useMemo` in this case as well, similar to previous `inFullScreenVisualization` ? Like:

```
  // check whether a user is at visualization
  const inVisualization = useMemo(() => {
    return (
      location.pathname
      .replace(routeBase, '')
      .includes('/visualizations')
    );
  }, [location, routeBase]);
```

moontrip:
@bobular After contemplating your comment at other PR ([here](https://github.com/VEuPathDB/web-eda/pull/192#issuecomment-871435562)), I think I got your point what it meant. Please correct me if I am wrong.

So, basically my latest change will work for the case, `/viz/pass-through/UUID`, but not work for , e.g., `/viz/computation/UUID` (not currently existing but may be added in the future). Is it what you thought?

bobular:
Sorry @moontrip - you are right - the simple version seems to work just fine.  For some reason (without testing it) I thought it would not.

bobular:
> Sorry @moontrip - you are right - the simple version seems to work just fine. For some reason (without testing it) I thought it would not.

Though is the `   .replace(routeBase, '') ` needed?

moontrip:
@bobular it is not a must as I doubt we will somehow use the string `visualizations` twice in a URL

bobular:
> @bobular it is not a must as I doubt we will somehow use the string `visualizations` twice in a URL

Aha - yes, good thinking

jtlong3rd:
> @jtlong3rd Is it better to use `useMemo` in this case as well, similar to previous `inFullScreenVisualization` ? Like:
> 
> ```
>   // check whether a user is at visualization
>   const inVisualization = useMemo(() => {
>     return (
>       location.pathname
>       .replace(routeBase, '')
>       .includes('/visualizations')
>     );
>   }, [location, routeBase]);
> ```

I think what you have is fine! `useMemo` is helpful when you want to (1) (2) preservce the 

> @jtlong3rd Is it better to use `useMemo` in this case as well, similar to previous `inFullScreenVisualization` ? Like:
> 
> ```
>   // check whether a user is at visualization
>   const inVisualization = useMemo(() => {
>     return (
>       location.pathname
>       .replace(routeBase, '')
>       .includes('/visualizations')
>     );
>   }, [location, routeBase]);
> ```

I actually like what you have right now! I realized yesterday that the `useMemo` which I suggested yesterday was overkill - I'll leave a comment in that PR when I have a moment.

jtlong3rd (approved):

src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -120,9 +108,11 @@ export function AnalysisPanel(props: Props) {
           },
           {
             display: 'Visualize',
-            // check whether user is at viz's full screen mode
-            route: inFullscreenVisualization
-              ? '/visualizations/pass-through'
+            // check whether user is at viz
+            route: location.pathname
+              .replace(routeBase, '')
+              .includes('/visualizations')
```
dmfalke, line 114: I think we should keep the replace, but the following can probably be replaced with a `startsWith`:

```suggestion
              .replace(routeBase, '')
              .startsWith('/visualizations')
```

dmfalke (changes requested):
Part of me wants to update the props api for `WorkspaceNavigation` such that you can use a function for the value of `route`, giving the consumer access to whether the tab is active or not.

E.g., 
```typescript
interface WorkspaceNavigationProps {
  display: 'Visualize',
  route: ({ isActive }) => '/visualizations' + (isActive ? lastVizPath : '')
}
```

But not sure it's worth it :slightly_smiling_face:.

I left a suggestion inline.

moontrip:
@jtlong3rd Thank you for your comment and suggestion 👍

src/lib/workspace/AnalysisPanel.tsx
moontrip: @dmfalke Thank you for your suggestion to change and I agree with that to use startsWith() 👍 May I ask your opinion on one more thing ? 😊 I am just debating myself, whether I will stick to the current approach you suggested or another approach to remove the last url element (e.g., using regEx). The former can certainly do what we want now, but the latter can handle more general cases as described follows.

For example, current approach will work for the current case, /viz/pass-through/UUID, but not work for more general case, e.g., /viz/computation/UUID (not currently existing but may probably be added in the future). The second approach (removing the last url element) would be able to handle general cases like `location.pathname.replace(routeBase, '').replace(/\/[^/]*$/, '')`

Certainly we can revisit this by sticking to the current one, though. What do you think?

dmfalke: I think this currently approach is actually the behavior we will want. In cases where there are multiple computations offered, clicking on the tab should go to the computation selector.

moontrip: @dmfalke thank you for your info 👍 I have updated with using startsWith() instead of includes()

bobular (approved):
Looks good - approved!

dmfalke (approved):

PR #200, opened 2021-07-02
changing the behavior of starred variables at MyVariable

moontrip:
when unstarring a variable at MyVariable, keep it visible

Here is an example  screenshot

myvariable1

### DIFF
--- a/src/lib/core/components/VariableList.tsx
+++ b/src/lib/core/components/VariableList.tsx
@@ -191,6 +191,26 @@ export default function VariableList(props: VariableListProps) {
 
   const starredVariablesLoading = starredVariables == null;
 
+  // moved this useState here
+  const [showOnlyStarredVariables, setShowOnlyStarredVariables] = useState(
+    false
+  );
+
+  // make visibleStarredVariables state be used at MyVariable
+  const [visibleStarredVariables, setVisibleStarredVariables] = useState(['']);
+
+  // update visibleStarredVariable for MyVariable
+  useEffect(() => {
+    setVisibleStarredVariables(starredVariables ? starredVariables : ['']);
+  }, [showOnlyStarredVariables]);
+
+  // update visibleStarredVariables when starredVariables are updated at VariableTree, not MyVariable
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


jtlong3rd (changes requested):
The new functionality is very nice, and your overall approach (maintain the visible starred variables in component state) is solid!

I have a few inline suggestions for simplifying the implementation and handling some linter warnings.

I also have a suggestion for a UX edge case, which GitHub won't let me describe in an inline comment because the relevant code is "too far away" from your diff. Currently, if "My Variables" is turned on, and you unstar every variable, "My Variables" is automatically turned off. This could be unpleasant for a user who accidentally unstars their last starred variable.

If you agree, you can make the following change: replace this block of code...

```typescript
  const starredVariableToggleDisabled = starredVariablesSet.size === 0;

  useEffect(() => {
    if (starredVariableToggleDisabled) {
      setShowOnlyStarredVariables(false);
    }
  }, [starredVariableToggleDisabled]);
```

... with this block of code:

```typescript
  const starredVariableToggleDisabled = (
    !showOnlyStarredVariables &&
    starredVariablesSet.size === 0
  );
```

(That is, (1) update the definition of `starredVariableToggleDisabled` so that the toggle is disabled when "My Variables" is turned off and there are no starred variables, and (2) delete the effect which was automatically turning off "My Variables" when there were no starred variables.)

moontrip:
@jtlong3rd sorry for late response. I am working for other project on every Friday so I didn't have time to look at your comments/suggestions carefully. In summary, I accepted all of your comments.

Some notes are left in the following. First of all, yes I knew that the toggle switch was turned off when all starred variables were removed at MyVariable. Since there was no specific requirement for that behavior, I just left it as was :) But I agree with your opinion that it is better to follow the basic approach, keep it visible. As for typing useState, I often defined types at useState so your comment is reasonable. Finally, I missed that approach to use toggleShowOnlyStarredVariables callback instead of using useEffect twice, which is much more elegant solution than mine. Thank you for suggesting better approach 👍👍

jtlong3rd (approved):

dmfalke (approved):

PR #204, opened 2021-07-05
Fix uncontrolled/controlled warning for Slider in HistogramFilter controls

### LINKED ISSUE #203: Material UI warning about controlled/uncontrolled props for Slider in HistogramFilter
I cannot reproduce this console warning in `web-components` storybook so I think it's an EDA issue with its usage of `Slider`.

Pretty sure I introduced this in the great tidy-up/controls refactor

bobular:
Fixes #203 

If there's a better way, I'd love to know.

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -460,10 +460,10 @@ function HistogramPlotWithControls({
         displayLibraryControls={displayLibraryControls}
         opacity={opacity}
         orientation={histogramProps.orientation}
-        binWidth={data?.binWidth}
+        binWidth={data?.binWidth ?? 1}
         onBinWidthChange={handleBinWidthChange}
-        binWidthRange={data?.binWidthRange}
-        binWidthStep={data?.binWidthStep}
+        binWidthRange={data?.binWidthRange ?? { min: 1, max: 2 }}
+        binWidthStep={data?.binWidthStep ?? 1}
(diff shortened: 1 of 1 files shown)


jtlong3rd (approved):

dmfalke (commented):
I'm wary of providing these "stub" values. I wonder if a better solution is to make sure the underlying widgets never use a value of `undefined`?

bobular:
> I'm wary of providing these "stub" values. I wonder if a better solution is to make sure the underlying widgets never use a value of `undefined`?

I agree they are ugly and arbitrary.

If we hid the offending widget while the value was undefined, it would be OK, but then the layout would change and be distracting.

dmfalke:
Okay, well let's go ahead with this solution for now. We can address in the future if something comes up.

dmfalke (approved):

bobular:
> Okay, well let's go ahead with this solution for now. We can address in the future if something comes up.

We could put the stub values in the widget implementations, e.g. in web-components/src/components/widgets/Slider.tsx

```
      <MUISlider
        min={minimum}
        max={maximum}
        value={localValue ?? 0}
        step={step}
        onChange={handleChange}
      />
```
This keeps the problem at the lowest possible level.  Out of sight, out of mind!

bobular:
Yeah, I've googled around some more, and everyone suggests "just use a fallback empty string" (e.g. `value={prop.foo ?? ''}`)

Not so good for number or slider inputs, but I will address this in a new PR in web-components.

PR #205, opened 2021-07-05
Issue 198 adapt HistogramFilter to use extra variable annotations

### LINKED ISSUE #198: Histogram filter: use defaults from variable metadata
Metadata for number and date variables now include the following properties, which should be used as defaults in the subsetting tab:

- `displayRangeMin`
- `displayRangeMax`
- `rangeMin`
- `rangeMax`
- `binWidthOverride`
- `binWidth`

These properties are used for two configuration properties of the histogram plugin: `binSpec` and `viewport`.

The login for `binSpec` is: use `binWidth`, unless `binWidthOverride` is provided, in which case it should be used.

The logic for `viewport` is: for `xMin` use `min(rangeMin, displayRangeMin)`, for `xMax` use `max(rangeMax, displayRangeMax)`.

It is expect that values for at lease `rangeMin`, `rangeMax`, and `binWidth` will be provided. We may need to consider what to do when this isn't the case.

More background on these properties can be found in https://github.com/VEuPathDB/plot.data/issues/16.

bobular:
Fixes #198 

Requires https://github.com/VEuPathDB/web-components/pull/163 in `web-components`

I also improved the UX I hope with this additional fix: if the user changes the x-range, then any existing binWidth settings are reset and the back-end will return something suitable.  This means that we have proper semantic zoom now enabled. I also enabled the new spinner.

Ready for testing and review but there are some known issues remaining

- [x] `studies/GEMSCC0003-1` endpoint isn't returning bin slider min, max or step - do we need to calculate it on client? Probably not, but see [#eda-backend thread](https://epvb.slack.com/archives/C01894F7P89/p1625495362391600)
- [x] when zooming in to a few weeks, and the bin width is "1 day",  then the bar width is wafer thin - I think the calculations are lacking 24 hours (especially if binStart and binEnd are the same date)
- [x] new bug found - when x-axis zoom min == max, the back end gives a 500 and with a date variable this can "kill" this variable in subsetting (no way to get the histogram back without deleting the analysis and starting again)
- [x] ~~check that data outside the curator's displayRangeMin/Max is not cut off~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.3.17",
+    "@veupathdb/components": "^0.3.18",
     "@veupathdb/wdk-client": "^0.1.4",
     "@veupathdb/web-common": "^0.1.2",
     "debounce-promise": "^3.1.2",
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -378,9 +378,6 @@ function HistogramPlotWithControls({
 
   const handleIndependentAxisRangeChange = useCallback(
     (newRange?: NumberOrDateRange) => {
-      console.log(
-        `handleIndependentAxisRangeChange newRange: ${newRange?.min} to ${newRange?.max}`
-      );
       updateUIState({
         // when the independent axis range is 'zoomed', reset the binWidth
         // so the back end provides a suitable value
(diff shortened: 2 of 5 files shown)


bobular:
Solved second sub-issue in [this web-components commit](https://github.com/VEuPathDB/web-components/commit/2fbb333970eb758989776361153f0ba6889ba692)

Also reported an issue with weekly binning here:
https://github.com/VEuPathDB/EdaDataService/issues/51

bobular:
First subtask is not really an issue.  Everything seems to be working fine with backend-provided bin slider min/max.

Third task is completed via the web-components task.

4th subtask will be a won't do - I suggest this is in the realm of https://github.com/VEuPathDB/web-eda/issues/98

PR #207, opened 2021-07-06
Variable Coverage Table

### LINKED ISSUE #121: Visualizations: add 'variable coverage' stats table
This is currently a placeholder.

I am trying to understand what the `completeCases` response really means when you have more than one variable (e.g. xVariable and overlayVariable for barplot viz), by playing around with examples, but these are not always working for me.  Could be related to https://github.com/VEuPathDB/EdaDataService/issues/38

jtlong3rd:
Resolves #121. 

**Notes**
1. In the long term, we might communicate coverage statistics using an UpSet plot. Perhaps we could use [this library](https://upset.js.org/)?
2. This implementation should be DRYed up once the UX requirements for variable coverage / missing data are finalized. Currently, there's a fair amount of per-visualization code for rendering the `OutputEntityTitle` and `VariableCoverageTable`, as well as laying out the `VariableCoverageTable` alongside the visualization's plot(s).


dmfalke (approved):

src/lib/core/hooks/variableCoverage.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Transforms a CompleteCasesTable array into a map
+ * @param completeCases
+ * @returns A map with key-value pairs of the form {entityId}.{variableId} -> {completeCaseCount}
+ */
+export function useCompleteCasesMap(
+  completeCases: CompleteCasesTable = []
+): Record<string, number> {
+  return useMemo(
+    () =>
+      completeCases.reduce((memo, { completeCases, variableDetails }) => {
```
bobular, line 100: Is this really a reduction?  If the same `variableDetails.variableId` is seen twice, the former value will be silently overwritten.

It seems that `lodash.keyBy` would do something similar, but it won't clean up the array-or-not-array completeCases mess, as shown in my toy example.

```
a = [ { name: 'john', age: 23 }, { name: 'mary', age: [ 22 ] } ]
z = _.keyBy(a, (obj) => obj.name)
{
  john: { name: 'john', age: 23 },
  mary: { name: 'mary', age: [ 22 ] }
}
```

src/lib/core/hooks/variableCoverage.ts
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Returns an array of data to be rendered in a VariableCoverageTable
+ * @param completeCases
+ * @param filteredEntityCounts
+ * @param variables
+ * @param outputEntityId
+ * @returns A VariableCoverageTableRow[]
+ */
+export function useVariableCoverageTableRows(
```
bobular, line 20: Something to discuss at tomorrow's meeting and probably not here...

What's the difference between a hook and a (memoized) utility function?

bobular (approved):
Tested functionally and made a few general comments in the code - that are more for discussion than demands for change.

Good to see findEntityAndVariable DRYed up :-)

src/lib/core/hooks/variableCoverage.ts
bobular: Though `_.keyBy` also silently overwrites!  I just expect more from `reduce` ;-)

jtlong3rd: You've kicked up two important issues:

1.  To resolve the "overwriting issue," we could update the backend so that each "row" of the `completeCases` response includes not just the id of the variable, but also the variable's role (xAxisVariable, yAxisVariable, etc). With such a change, the complete cases map could be keyed by variable role instead of variable id.

2. While you're right that replacing my `reduce` with a `keyBy` won't automatically resolve the "overwriting issue" or the "array-or-not-array completeCases mess," there's still a clear opportunity to use `keyBy` to improve readability:

Given the complete cases rows:

* `filter` away the rows that are missing `completeCases` or `variableDetails`
* `keyBy` the variable id associated to each row
* `map` each row to its complete cases count (via lodash's `mapValues`)

That means making three passes through the data (instead of one pass with the `reduce` implementation), but that's OK. If these three passes are a source of performance issues, we have much bigger issues on our hands!

src/lib/core/hooks/variableCoverage.ts
jtlong3rd: Sure - let's add that to the agenda!

src/lib/core/hooks/variableCoverage.ts
bobular: Well, it's an interesting case of readability vs conciseness and performance.

I hadn't thought of the incomplete cases part (which makes it a proper reduction, I had missed that) - so it ends up being six of one and half a dozen of the other (though I learned something in the process).

Issue #211, opened 2021-07-07
Visualization: Legend missing when overlay variable has only one value


bobular:
The legend should be shown!

Example: subset/filter for one country, then do a viz plot and overlay by country

moontrip:
It can be done something like this

`displayLegend={data.value && (data.value.series.length > 1 || typeof vizConfig.overlayVariable !== 'undefined')}`

Issue #183, opened 2021-06-22
Visualizations: remove redundant `enableOverlay` viz state


bobular:
This is a relic from my very first Histogram viz straw man.  There used to be an on/off switch to toggle the use of the overlay variable.

That is no longer the approach we're using but the code still remains (at least in Histogram and Scatterplot viz implementations) - it should be removed.

Issue #148, opened 2021-06-07
Visualizations: create legend title with overlay variable's name for all plots


bobular:
Currently any viz using an overlay variable can end up with a mysterious "yes, no" legend, as below.
Even the more self-explanatory variables, such as country, should have a title above the legend.

![image](https://user-images.githubusercontent.com/308639/121054134-f5180e80-c7b3-11eb-9ecb-bb441e9a7f29.png)

moontrip:
An example with XYPlot component at web-component is now [approved and merged](https://github.com/VEuPathDB/web-components/pull/152). Thus, other plot components and corresponding vizs would be applied with the approach.

bobular:
Hi @moontrip thank you for this.

I have moved your implementation into the new tidied/refactored PlotlyPlot so I will mark this issue as being solved by the refactoring PR.

moontrip:
Thanks Bob for your information 👍 I glanced at your commits regarding this but it seems like this part is not pushed to the PR yet: will take a look at it later. 

I have also thought of the case to move this legend title part to `Plotlyplot` component. However, I was reluctant to do that because I did not want to send additional prop to `Plotlyplot` beyond key props like `data`, `config`, and `layout`. I thought it is better to handle such a prop (e.g., legendTitle prop) at each plot component level, not `Plotlyplot`. :)

bobular:
Hi DK,

It's here in PlotlyPlot
https://github.com/VEuPathDB/web-components/blob/83784fa1cdf1aacc39d9a3b2385ee42dad43b2b1/src/plots/PlotlyPlot.tsx#L124
and used in a story here
https://github.com/VEuPathDB/web-components/blob/83784fa1cdf1aacc39d9a3b2385ee42dad43b2b1/src/stories/plots/XYPlot.stories.tsx#L38

I worked out (with some help from Dave) how to send non-Plotly props to
PlotlyPlot, so now all the general things like `title`,
`displayLibraryControls`, `displayLegend`, etc are all handled by
PlotlyPlot.

moontrip:
@bobular thank you for sharing the info 👍 

Actually I am thinking of handling the date string for scatter plot viz for my next task (after refactoring boxplot viz), so some of your changes would be helpful - I was thinking that we might not need to use Date type though: perhaps string type is sufficient for plotly purpose but I will revisit it in the near future.

[...]

PR #212, opened 2021-07-08
legend title, legend bug, interactive, enableOverlay

moontrip:
This PR addresses three small issues and one bug for vizs at once. For Mosaic plot, only d) is addressed as it does not have overlay (concerning a) and b)) and c) was already addressed. Boxplot viz was already addressed at other PR.

a) [Visualization: Legend missing when overlay variable has only one value #211](https://github.com/VEuPathDB/web-eda/issues/211)

b) [Visualizations: create legend title with overlay variable's name for all plots #148](https://github.com/VEuPathDB/web-eda/issues/148) - hope you don't mind my work instead of you, Bob ;)

c) [Visualizations: remove redundant enableOverlay viz state #183](https://github.com/VEuPathDB/web-eda/issues/183)

d) set `interactive` props for full screen mode (no ticket exists)

Some screenshots for a) and b) - filter: Bangladesh only

scatter-legend-title

histogram-legend-title

barplot-legend-title

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
@@ -228,8 +228,7 @@ function BarplotViz(props: Props) {
           barLayout={'group'}
           displayLegend={
             data.value &&
-            (data.value.series.length > 1 ||
-              typeof vizConfig.overlayVariable !== 'undefined')
+            (data.value.series.length > 1 || vizConfig.overlayVariable != null)
           }
           independentAxisLabel={
             vizConfig.xAxisVariable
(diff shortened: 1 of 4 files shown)


jtlong3rd (commented):
This looks good! I have one general suggestion:

Instead of checking whether `vizConfig.overlayVariable` is defined using `typeof vizConfig.overlayVariable !== 'undefined'`, I would use `vizConfig.overlayVariable != null`. It's a little more concise, and also guards against the possibility that `vizConfig.overlayVariable` is `null`.

(`x == null` evaluates to `true` if `x` is `undefined` or `null`. The use of `==` is typically discouraged in JavaScript, but our linter permits `== null`, because it's an extremely convenient way to check if a value is `null` or `undefined`.)

moontrip:
@jtlong3rd Thank you for your review and suggestion 👍 I have made changes to use `!= null` 😃

jtlong3rd (approved):

bobular (approved):
Looks and works great.  Thanks DK - no I'm not sad you did the legend titles ;-)

PR #216, opened 2021-07-08
Variable tree updates

### LINKED ISSUE #194: Variable tree: order variables by order number
[Depends on https://github.com/VEuPathDB/web-eda/pull/189]

Variables have an optional property called `displayOrder`. This property should be used to sort siblings within a variable tree. We should be able to do this using lodash's [orderBy](https://lodash.com/docs/4.17.15#orderBy):

```typescript
  const orderedChildren = orderBy(children, child => child.displayOrder ?? child.displayName)
```
Addressing this issue will likely require an overhaul of how we are constructing the variable tree for display. Currently, we are reusing some code from wdk-client which does not do this ordering, and which requires converting "eda variables" into "wdk fields". This is overhead we should get rid of, and this is a good time to do that.

### LINKED ISSUE #196: Variable tree: Featured variables
[Depends on https://github.com/VEuPathDB/web-eda/pull/189]

The new variable property `isFeatured` should be used to create a **Featured variables** category in the variable tree. See mock-ups [here](https://docs.google.com/presentation/d/16dVuZ5Xcyv9Ql2kSkxX4TC5X8_xtIAkY9RckrPo6A7Q/edit#slide=id.gdc07b6bbd8_0_0).

dmfalke:
This PR incorporates new variable metadata into the `VariableTree` component. It also includes some CSS updates for the scroll-able section of the variable tree, and an improvement to `io-ts` validation error messages.

The featured variables are rendered above the tree of variables. It is a separate section from the tree, which does not scroll with the rest of the variables. This allows the selected variable to be visible within the view of the tree, without losing the context of the featured variables section.

This will all require some UX review.

closes #196 
closes #194

### DIFF
--- a/src/lib/core/components/VariableList.tsx
+++ b/src/lib/core/components/VariableList.tsx
@@ -355,7 +355,7 @@ export default function VariableList(props: VariableListProps) {
                   const isActive = field.term === activeField?.term;
                   const isDisabled = disabledFields.has(field.term);
                   return (
-                    <li>
+                    <li key={field.term}>
                       <a
                         className={
                           'wdk-AttributeFilterFieldItem' +
--- a/src/lib/core/components/VariableList.tsx
+++ b/src/lib/core/components/VariableList.tsx
@@ -39,6 +39,7 @@ import {
 import { cx } from '../../workspace/Utils';
 import { Tooltip } from '@material-ui/core';
 import { HtmlTooltip } from '@veupathdb/components/lib/components/widgets/Tooltip';
+import { safeHtml } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
 
 //defining types - some are not used (need cleanup later)
 interface VariableField {
@@ -82,6 +83,7 @@ interface VariableListProps {
   disabledFieldIds?: string[];
   hideDisabledFields: boolean;
   setHideDisabledFields: (hide: boolean) => void;
+  featuredFields: VariableField[];
 }
 
 interface getNodeSearchStringType {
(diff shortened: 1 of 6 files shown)


dmfalke:
I also tweaked the variable constraints tooltip that has been annoying. I added a delay for when it shows up. Hopefully it will not get in the way as much.

src/lib/core/components/VariableList.tsx
```diff
@@ -341,11 +347,52 @@ export default function VariableList(props: VariableListProps) {
...
           )}
+          {featuredFields.length && (
+            <div className="FeaturedVariables">
+              <h3>Featured variables</h3>
+              <ul>
+                {featuredFields.map((field) => {
+                  const isActive = field.term === activeField?.term;
+                  const isDisabled = disabledFields.has(field.term);
+                  return (
+                    <li>
```
jtlong3rd, line 358: Missing "key"

jtlong3rd (changes requested):
Except for one "business-as-usual" inline suggestion, looks good!

src/lib/core/components/VariableList.tsx
dmfalke: Ugh, thanks!

jtlong3rd (approved):

bobular (approved):
Have tested (including search-by-variable/vocabulary-value) and it all works nicely.  I skimmed the code and don't see any issues.

Some questions for future: what about real-estate if there are 25 featured variables? Or are they always single-digits?

dmfalke:
I hadn't thought about real estate for that many features vars. It's a good question. We could make that section collapsible and scrollable as well. I'll go ahead and merge this and we can discuss.

PR #218, opened 2021-07-10
Limit number of decimal places shown in histogram filter stats

### LINKED ISSUE #124: subsetting: in histogram for numerics, limit precision of the reported Mean
for example, in GEMS, see the "Age at eligibility form completion" variable.  the mean is "17.7058093676607"

probably should use the (upcoming) `precision` meta-data field (and add 1).

Also take care of line wrapping (comment below)

dmfalke:
fixes #124

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -254,10 +254,12 @@ export function HistogramFilter(props: Props) {
             }}
           >
             <div className="histogram-summary-stats">
-              <b>Min:</b> {fgSummaryStats.min} &emsp; <b>Mean:</b>{' '}
-              {fgSummaryStats.mean} &emsp;
-              <b>Median:</b> {fgSummaryStats.median} &emsp; <b>Max:</b>{' '}
-              {fgSummaryStats.max}
+              <b>Min:</b> {formatStatValue(fgSummaryStats.min, variable.type)}{' '}
+              &emsp; <b>Mean:</b>{' '}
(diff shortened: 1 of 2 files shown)


src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -573,3 +575,15 @@ async function getHistogram(
...
+
+// TODO [2021-07-10] - Use variable.precision when avaiable
+function formatStatValue(
+  value: string | number,
+  type: HistogramVariable['type']
+) {
+  return type === 'date'
+    ? value
+    : Number(value).toLocaleString(undefined, {
+        maximumFractionDigits: 4,
```
bobular, line 587: This should probably go to the scientific advisory board, but I am strongly of the opinion that we display summary statistics to a number of **significant figures**.    Individual data points should be displayed an appropriate number of **decimal places** where appropriateness is related to the level of precision the datum can be measured.  Maybe we can come back to this when we have precision metadata from the back end.

As is:
image

With maximumSignificantDigits: 4
image

bobular (approved):
Happy to approve without changes.  I'm just raising some discussion of decimal places vs significant places.  It's not trivial.  Maybe it'll be fine when we have the metadata.

src/lib/core/components/filter/HistogramFilter.tsx
dmfalke: I'm fine with whatever folks want. I was under the impression that the `precision` field will describe the number of decimals to show, not _actual_ precision. We should clarify.

Issue #197, opened 2021-06-30
Selected variable view: Display definition


dmfalke:
[Depends on https://github.com/VEuPathDB/web-eda/pull/189]

In the selected-variable view, show the variable's `definition` beneath the display name, if available.

PR #219, opened 2021-07-12
add variable.definition

moontrip:
add variable.definition below variable selector at viz. Here is a screenshot

display-definition

### DIFF
--- a/src/lib/core/components/VariableTree.tsx
+++ b/src/lib/core/components/VariableTree.tsx
@@ -200,6 +200,20 @@ export function VariableTreeDropdown(props: Props) {
           />
         </div>
       </PopoverButton>
+      {/* add variable.definition */}
+      {variable && (
+        <div
+          style={{
+            position: 'absolute',
+            top: '3em',
+            right: 0,
+            color: '#000000',
+            fontSize: '1em',
+          }}
+        >
+          {variable?.definition}
+        </div>
+      )}
     </div>
   );
 }
--- a/src/lib/core/components/VariableTree.tsx
+++ b/src/lib/core/components/VariableTree.tsx
@@ -200,12 +200,6 @@ export function VariableTreeDropdown(props: Props) {
           />
         </div>
       </PopoverButton>
-      {/* add variable.definition */}
-      {variable && (
-        <div className={cx('-DisplayVariableDefinition')}>
-          {variable?.definition}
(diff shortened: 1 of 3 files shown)


moontrip:
moved style to scss

bobular (commented):
Hi DK.

I'm seeing layout of z-axis issues when a server response error is shown

image

What's the longest variable definition we are likely to get?  And is there a plan for these long ones?

Finally, is this task only concerning the `InputVariables` component?  As #197 is a little vague, I suggest these also as important
1. mouse-over in the VariableTree
2. display variable definition in subsetting along with the exisiting "Original variable name" (I would expect to see the full definition on mouse-over the (?) icon - rather than the rather boring generic mouseover currently there) - or maybe add a (i) icon ('i' for information)

moontrip:
@bobular yes I am aware of that overlapping issue with (error) messages but does not have clear answer to address. It is because the current text of variable.definition is associated with each variable selector to be shown appropriately in space (esp. width). It may be okay for demo purpose ? 🤔 
One possible solution is to use mouse-over to a simple text, like showing 'variable definition' text at the bottom of variable selector and add mouse-over event on the text.

Yes the ticket is ambiguous. I understood that the ticket was for Viz's variable selector, so perhaps it needs more clear description on its scope and work to be done.

dmfalke:
@DK I don't think this was the intended place for the variable definition. This should be in the subsetting tab, below the "Original variable name":
image

moontrip:
@dmfalke aha I see. One quick question: do we need a prefix for variable.definition (like `Variable definition: {variable.definition}` or simply show the contents of variable.definition (`{variable.definition}`) in there?

dmfalke:
This is what we're doing on the live site
image

We don't have a prefix, which I think it fine. On the live site, the definition comes before the provider label. It makes more sense to me for it to come after, so I would do that.

moontrip:
@dmfalke is this acceptable ? I set small margins at top and bottom

variabledefinition1

variabledefinition2

dk:
Hi all, please don't cc me. Thanks

dmfalke (approved):

jtlong3rd (approved):

moontrip:
merging this PR. Perhaps Bob's comment above may be discussed in the mtg or issue

Issue #222, opened 2021-07-13
Featured variables: UI improvements


dmfalke:
This feature was reviewed at the July 12, 2021 EDA UX meeting. The following changes were requested:
- [ ] make collapsible
- [ ] increase shading
- [ ] tooltip of entire _Featured variables_ (Danica and Sheena will provide a definition)

dmfalke:
I just remembered that I want to preserve the collapsed state of the featured vars panel. I will do this now.

PR #223, opened 2021-07-13
Featured vars: darken background and DRY up rendering

dmfalke:
I don't have verbiage for tooltip content yet, but I think we can go ahead with these other changes.

#222


jtlong3rd (approved):

dmfalke:
I just remembered that I want to preserve the collapsed state of the featured vars panel. I just pushed a commit for this, and to fix an issue with padding.

jtlong3rd (approved):

PR #226, opened 2021-07-14
Enable Data Restrictions

jtlong3rd:
This PR introduces [data restrictions](https://github.com/VEuPathDB/EbrcWebsiteCommon/blob/master/Client/src/App/DataRestriction/DataRestrictionUtils.js) into the dev environment. The restrictions can be overridden by specifying a `REACT_DISABLE_DATA_RESTRICTIONS=true` environment variable.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -67,10 +67,10 @@ initialize({
     Page: (DefaultComponent: React.ComponentType<Props>) => {
       return function ClinEpiPage(props: Props) {
         useEffect(() => {
-          if (process.env.REACT_APP_ENABLE_DATA_RESTRICTIONS === 'true') {
-            enableRestriction();
-          } else {
+          if (process.env.REACT_APP_DISABLE_DATA_RESTRICTIONS === 'true') {
             disableRestriction();
+          } else {
+            enableRestriction();
           }
         }, []);
 
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,6 +1,6 @@
 import './globals';
 import { endpoint, rootElement, rootUrl } from './constants';
-import React from 'react';
+import React, { useEffect } from 'react';
 import reportWebVitals from './reportWebVitals';
 
 import { initialize } from '@veupathdb/web-common/lib/bootstrap';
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -54,8 +64,29 @@ initialize({
   ],
   componentWrappers: {
     SiteHeader: () => Header,
+    Page: (DefaultComponent: React.ComponentType<Props>) => {
+      return function ClinEpiPage(props: Props) {
+        useEffect(() => {
+          if (process.env.REACT_APP_ENABLE_DATA_RESTRICTIONS === 'true') {
```
bobular, line 70: Would a fail-safe approach be better?  The restrictions are on unless devs explicitly turn them off?

I know this might cause short-term pain when we first pull the changes into our dev instance, but it's better than forgetting to turn it on in a production instance, maybe?

jtlong3rd: In fact, reverting my most recent commit will change the behavior to the fail-safe approach you described! I'll go with my first instincts, because the short-term pain isn't all that painful.

I should clarify that: 

1. This component is the `Page` that's rendered by the dev server - the production version of the `Page` currently lives [here](https://github.com/VEuPathDB/ClinEpiWebsite/blob/master/Site/webapp/js/client/component-wrappers/Page.tsx).
2. In this repo, assets which lie outside `src/lib`, such as this file, aren't published to npm.

bobular:
Aha - I see - the production version of Page.tsx has it always-on.
That wasn't immediately obvious ;-)

dmfalke (approved):

dmfalke:
@jtlong3rd do we need to add a proxy for getting permissions, or does this come from the wdk rest service?

jtlong3rd:
> @jtlong3rd do we need to add a proxy for getting permissions, or does this come from the wdk rest service?

Offline, I clarified that we will eventually need to add this proxy, but not right away. The endpoint which fetches a user's permissions (maintained [here](https://github.com/VEuPathDB/service-dataset-access)) is only exercised for registered users, and currently, the local dev environment doesn't offer WDK login.

PR #230, opened 2021-07-15
Use updated distribution endpoint

dmfalke:
This PR includes changes to use the updated distribution endpoint for filters in the subsetting tab. The updated endpoint is designed for performance, which is a requirement for subsetting.

One UX compromise is that we cannot do semantic zoom in this context. The front end must always provide range and binning info.

This depends on https://github.com/VEuPathDB/web-components/pull/165 for mouse selection in histograms to work correctly.

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -25,11 +25,10 @@ import { getOrElse } from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/function';
 import { number, partial, TypeOf, boolean, type, intersection } from 'io-ts';
 import React, { useCallback, useMemo } from 'react';
-import { DataClient, HistogramRequestParams } from '../../api/data-api';
 import { usePromise } from '../../hooks/promise';
 import { AnalysisState } from '../../hooks/analysis';
 import { useSubsettingClient } from '../../hooks/workspace';
-import { DateRangeFilter, Filter, NumberRangeFilter } from '../../types/filter';
+import { DateRangeFilter, NumberRangeFilter } from '../../types/filter';
 import { StudyEntity, StudyMetadata } from '../../types/study';
 import { TimeUnit, NumberOrDateRange, NumberRange } from '../../types/general';
 import { gray, red } from './colors';
@@ -153,13 +152,13 @@ export function HistogramFilter(props: Props) {
       );
 
       const series = [
-        histogramResponseToDataSeries(
+        distributionResponseToDataSeries(
           `All ${entity.displayName}`,
           distribution.background,
           gray,
           variable.type
         ),
-        histogramResponseToDataSeries(
+        distributionResponseToDataSeries(
           `Subset of ${entity.displayName}`,
           distribution.foreground,
           red,
@@ -566,7 +565,7 @@ function HistogramPlotWithControls({
   );
 }
 
-function histogramResponseToDataSeries(
+function distributionResponseToDataSeries(
   name: string,
   response: DistributionResponse,
   color: string,
@@ -574,8 +573,8 @@ function histogramResponseToDataSeries(
 ): HistogramDataSeries {
   const bins = response.histogram.map(
     ({ value, binStart, binEnd, binLabel }) => ({
-      binStart,
-      binEnd,
+      binStart: type === 'date' ? binStart : Number(binStart),
+      binEnd: type === 'date' ? binEnd : Number(binEnd),
       binLabel,
       count: value,
     })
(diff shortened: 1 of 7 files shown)


dmfalke:
Note, I plan to merge several commits into one, so that we can easily revert to using the histogram and barplot plugins, if needed.

bobular:
I'm not sure how you plan to do the latter, but sounds good.  I will review
now.

dmfalke:
I think this is ready

bobular (commented):
HistogramFilter:
At the moment the year/month/week/day menu is showing up for non-date variables. It would be nice to replace the `"(month)"` shown in the label of BinWidthControl's SliderWidget with the dropdown menu.  A bit of a fiddle but maybe worth it?

Default slider range for date vars could be 1-60?

`22,567 (100%) of 22,567 Participant have data for this variable` seems to have disappeared for HistogramFilter

dmfalke:
Thanks, Bob. All good points. I will address as soon as I can.

dmfalke:
I pushed commits to address these suggestions. See also the relevant commit in https://github.com/VEuPathDB/web-components/pull/165.

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -80,28 +79,32 @@ export function HistogramFilter(props: Props) {
...
     if (variable.type === 'number')
       return {
-        binWidth: variable.binWidthOverride ?? variable.binWidth,
+        binWidth: variable.binWidthOverride ?? variable.binWidth ?? 0.1,
         binWidthTimeUnit: undefined,
         independentAxisRange:
           variable.displayRangeMin != null && variable.displayRangeMax != null
             ? { min: variable.displayRangeMin, max: variable.displayRangeMax }
-            : undefined,
+            : { min: Math.min(0, variable.rangeMin!), max: variable.rangeMax! },
```
bobular, line 87: Should we make `{Date,Number}Variable.range{Min,Max}` mandatory props in types/study.ts?

It seems that if `NumberVariable.rangeMax` is missing, the request to the back end will fall over.

If they are missing from the studies/variables metadata we might get a more meaningful error message?

Same goes for date rangeMin and Max - see below

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -80,28 +79,32 @@ export function HistogramFilter(props: Props) {
...
       independentAxisRange:
         variable.displayRangeMin != null && variable.displayRangeMax != null
-          ? { min: variable.displayRangeMin, max: variable.displayRangeMax }
-          : undefined,
+          ? {
+              min: variable.displayRangeMin + 'T00:00:00',
+              max: variable.displayRangeMax + 'T00:00:00',
+            }
+          : {
+              min: variable.rangeMin! + 'T00:00:00',
```
bobular, line 105: If `variable.rangeMin` is missing, `'T00:00:00'` is likely to be an invalid date for the back end.  

See previous comment above.

bobular (approved):
Sorry, I meant to start a review but was tired and pressed the wrong button with the previous comments about rangeMin and rangeMax being missing.

If you think it's OK to keep it as-is, I'm all for that.

dmfalke:
I’ll play with making them mandatory. I’m not sure they’ll be populated for
categorical number and date vars, so might require even more factoring.

bobular:
Here's a categorical number var from GEMS1 that does have rangeMin/Max, so
hopefully happy days (but not more than 7)!

{
"63": {
"id": "EUPATH_0010238",
"parentId": "EUPATH_0010016",
"providerLabel": "[\"F4A_DRH_DAYS\"]",
"displayName": "Diarrheal episode duration",
"definition": "Assessed at enrollment for cases",
"vocabulary": [
"1",
"2",
"3",
"4",
"5",
"6",
"7"
],
"displayType": "default",
"displayOrder": 1,
"type": "number",
"dataShape": "categorical",
"units": "",
"precision": 0,
"rangeMin": 1,
"rangeMax": 7,
"binWidth": 0.3,
"distinctValuesCount": 7,
"isTemporal": false,
"isFeatured": false,
"isMergeKey": false,
"isMultiValued": false
}
}

dmfalke:
Shoot, I forgot to squash my commits! Hopefully we wont have to revert anything!

dmfalke:
If we need to revert, I can just revert individual commits. It shouldn't be too bad, since they are colocated in this PR.

dmfalke:
I also directly pushed this commit: https://github.com/VEuPathDB/web-eda/commit/9f4a72796d2594fbbc23232a4d0dd131fabf9063

PR #239, opened 2021-07-16
Use adjustBinEndToEndOfDay prop for HistogramVisualisation

bobular:
Requires https://github.com/VEuPathDB/web-components/pull/167

This super-simple change adds the `adjustBinEndToEndOfDay` prop to the Histogram component for the HistogramVisualisation (which receives date-based bin data like this `binStart: 2001-03-01, binEnd: 2001-03-31`)

The HistogramFilter uses the Distribution endpoint which returns `binStart: 2001-03-01, binEnd: 2001-04-01`

Here is the HistogramFilter which doesn't use the day-adjustment any more.  All fixed!
image

### DIFF
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -311,6 +311,7 @@ function HistogramViz(props: Props) {
             findEntityAndVariable(entities, vizConfig.overlayVariable)?.variable
               .displayName
           }
+          adjustBinEndToEndOfDay
         />
       ) : (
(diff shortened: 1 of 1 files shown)


bobular:
We're not quite out of the woods yet...

The filter in the above screenshot creates the following histogram viz

image

There's still an issue with the inclusive/exclusive nature of date filters that I will look in to

bobular:
I think it's all about how the filters are implemented on the back end

I've made a comment in https://github.com/VEuPathDB/EdaDataService/issues/54

bobular:
Closing without merging.  This is not needed now thanks to back end work.

PR #242, opened 2021-07-20
Improve histogram and barplot viz controls

bobular:
Depends on https://github.com/VEuPathDB/web-components/pull/169 

So far I have done

- fixed the binWidth control to handle date units
- added absolute/relative control to histogram
- added absolute/relative and logscale controls to barplot

I've chosen not to add zooming controls to histogram x-axis at this time.
Maybe wait until after #98 to add this, if still wanted.

PR #246, opened 2021-07-22
Adapt client to latest API changes

bobular:
Just a draft placeholder PR to make it clear I'm working on this.

Here is Danielle's list of changes:

1. fixes to binning date histos for the viz tab
1. making sure facetVariableDetails is always an array
4. updated viz descriptions
5. fixing variableId in the various variableDetails (so it stops saying entityId.variableId)
6. updates to the 'proportion' option for bar and histo (note: the request will now have to specify a barmode)
7. fix the boxplot bug where 'rawData' was out of order
8. add support for showing 'No data' as a value for strata vars (note: new showMissingness prop in request)
9. include a count of 'plottedIncompleteCases' (normally 0, unless new 'showMissingness' is on)
10. make statsTables in the responses optional (for when 'showMissingness' is on)
11. new `data.label` prop for boxplot response in place of entity.variable-named key
12. `mean: 'true' | 'false'` now capitalized `'TRUE' | 'FALSE'`

### DIFF
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -67,6 +67,7 @@ export interface HistogramRequestParams {
   config: {
     outputEntityId: string;
     valueSpec: 'count' | 'proportion';
+    barmode: 'overlay' | 'stack';
     xAxisVariable: VariableDescriptor;
     overlayVariable?: VariableDescriptor; // TO DO: should this be StringVariable??
     facetVariable?: ZeroToTwoVariables; // ditto here
@@ -143,6 +144,7 @@ export interface BarplotRequestParams {
     outputEntityId: string;
     // add proportion as it seems to be coming
     valueSpec: 'count' | 'identity' | 'proportion';
+    barmode: 'group' | 'stack';
     xAxisVariable: VariableDescriptor;
     // barplot add prop
     overlayVariable?: VariableDescriptor;
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -49,7 +49,7 @@ const sampleSizeTableArray = array(
 export type CompleteCasesTableRow = TypeOf<typeof completeCases>;
 const completeCases = partial({
   // set union for size as it depends on the presence of overlay variable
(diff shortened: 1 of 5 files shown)


moontrip:
@bobular thank you for your efforts 👍 I am not quite sure if you already noticed and was working, there are unknown changes for boxplot, which Danielle just confirmed. Previously mean option was lowercase but now uppercase. At `BoxplotRequestParams` of `data-api.ts`

from 
`    mean: 'true' | 'false';`

to
```
    mean: 'TRUE' | 'FALSE';
    computeStats: 'TRUE' | 'FALSE';
```

I realized there is new input option, `computeStats`, so I have added it too.

Accordingly, `getRequestParams` at `BoxplotVisualization` needs to be changed from

from
`      mean: 'true',`

to
```
      mean: 'TRUE',
      computeStats: 'TRUE',
```

Certainly there are more cases to be changed at both `data-api.ts` (e.g., overlayVariableDetails now returns empty that affects legend text for scatterplot and boxplot (asking Danielle if it will be empty or bug), new label for boxplot, etc.)

jtlong3rd:
@bobular Thanks for doing this! @moontrip brought to my attention that the Variable Coverage Table is broken now - the counts aren't appearing in the table. This appears to be due to the change of variableId from "entityId.variableId" to "variableId".

Since you're "in the neighborhood," could you update the [Variable Coverage hooks ](https://github.com/VEuPathDB/web-eda/blob/800546a2f184e8e34b43b6bfedf8163eff39aab3/src/lib/core/hooks/variableCoverage.ts) to support the new variableId format? This is also a good opportunity to change the implementation of `useCompleteCasesMap` back to `reduce`, like you wanted. 😉

bobular:
I'm going to stop working on this and put it up for review.
All visualisations are working (with coverage table too).

There's possibly some work still to do with histogram date bin responses (now exclusive binEnd, was inclusive) but that's easier to test in the branch with superior histogram controls #242 

So a quick review and merge would be helpful if poss please.

dmfalke (approved):

moontrip (approved):

jtlong3rd (approved):

PR #247, opened 2021-07-22
Add route to create a new analysis

dmfalke:
This is needed for links to create a new analysis that will appear on the ClinEpiDB homepage and menu.

### DIFF
--- a/src/lib/workspace/NewAnalysis.tsx
+++ b/src/lib/workspace/NewAnalysis.tsx
@@ -11,8 +11,9 @@ export function NewAnalysis(props: Props) {
   const { analysisClient: analysisStore, studyId } = props;
   const history = useHistory();
   useEffect(() => {
-    async function run() {
-      const { id } = await analysisStore.createAnalysis({
+    let cancelled = false;
+    analysisStore
+      .createAnalysis({
         name: 'Unnamed Analysis',
         studyId,
         filters: [],
@@ -21,14 +22,19 @@ export function NewAnalysis(props: Props) {
         visualizations: [],
         computations: [],
         variableUISettings: {},
+      })
+      .then(({ id }) => {
+        if (!cancelled) {
+          const newLocation = {
+            ...history.location,
+            pathname: `./${id}`,
+          };
+          history.push(newLocation);
+        }
       });
-      const newLocation = {
-        ...history.location,
-        pathname: `./${id}`,
-      };
-      history.push(newLocation);
-    }
-    run();
+    return () => {
+      cancelled = true;
+    };
   }, [analysisStore, history, studyId]);
   return <div style={{ fontSize: '3em' }}>Creating new analysis...</div>;
 }
--- a/src/lib/workspace/NewAnalysis.tsx
+++ b/src/lib/workspace/NewAnalysis.tsx
@@ -1,0 +1,34 @@
+import React, { useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+import { AnalysisClient } from '../core';
+
+interface Props {
+  analysisClient: AnalysisClient;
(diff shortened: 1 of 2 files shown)


src/lib/workspace/NewAnalysis.tsx
```diff
@@ -0,0 +1,34 @@
...
+        derivedVariables: [],
+        visualizations: [],
+        computations: [],
+        variableUISettings: {},
+      });
+      const newLocation = {
+        ...history.location,
+        pathname: `./${id}`,
+      };
+      history.push(newLocation);
```
jtlong3rd, line 29: I think we should "cancel" this redirect if the user navigates away from the "new analysis" route before the analysis creation has finished. What do you think?

jtlong3rd (commented):
Looks good! I had one inline comment about the current UX.

src/lib/workspace/NewAnalysis.tsx
dmfalke: Yep, I agree with that. Do we have any utilities that would help in this case?

dmfalke: I forgot about this utility class: https://github.com/VEuPathDB/WDKClient/blob/af2099baf3048f65d759023c2c7837d1b359d70b/Client/src/Utils/Task.ts. I might try it out

jtlong3rd: Neat! If that doesn't pan out, I suppose you could always... 

1. Do the "usual" trick of adding a `cancelled` or `ignore` variable to the effect

OR

2. Use `usePromise` to create the analysis and conditionally render a `Redirect`

dmfalke: I do think (1) is the easiest to read.

This is what it looks like with `Task`:

```typescript
  useEffect(
    () =>
      Task.fromPromise(
        analysisStore.createAnalysis({
          name: 'Unnamed Analysis',
          studyId,
          filters: [],
          starredVariables: [],
          derivedVariables: [],
          visualizations: [],
          computations: [],
          variableUISettings: {},
        })
      ).run(({ id }) => {
        const newLocation = {
          ...history.location,
          pathname: `./${id}`,
        };
        history.push(newLocation);
      }),
    [analysisStore, history, studyId]
  );
```

jtlong3rd: Let's go with (1) then!

dmfalke: For the sake of comparison
```typescript
  useEffect(() => {
    let cancelled = false; 
    analysisStore.createAnalysis({
      name: 'Unnamed Analysis',
      studyId,
      filters: [],
      starredVariables: [],
      derivedVariables: [],
      visualizations: [],
      computations: [],
      variableUISettings: {},
    }).then(({ id }) => {
      if (!cancelled) {
        const newLocation = {
          ...history.location,
          pathname: `./${id}`,
        };
        history.push(newLocation);
      }
    });
    return () => { cancelled = true };
  }, [analysisStore, history, studyId]);
```

dmfalke: The thing I like about the version with `Task` is that there are fewer variables floating around, thus less potential for something to break in the future (not likely here, but in general).

jtlong3rd: Using Task is fine by me! (And I'll definitely have to keep in mind that we have this utility at our disposal.)

dmfalke: Okay, I pushed a version using `Task` :shrug:

jtlong3rd (approved):

PR #252, opened 2021-07-26
API-related fix barmode to barMode

bobular:


bobular:
Bad-boy merging without approval

Issue #229, opened 2021-07-15
var completeness table jumps around on the page


d-callan:
sometimes its to the right of the scatter plot (gems 1 case control) and sometimes below.

moontrip:
@d-callan (and @dmfalke)
I have tested this but cannot reproduce it. Thus, below is just my general thought.

The plot and the table are placed in the left and right, respectively and they are dynamically laid out depending on one's display resolution and/or browser's width (designed by @jtlong3rd) like a modern website does: this is the same for other vizs beyond scatter plot. I don't know exactly how you observe the behavior, but one possible case I can guess is that you might have reduced your browser's size, which might lead to top/bottom display automatically instead of left/right one - you can test this by increase/decrease your browser's width.

Thus I think this is a non-issue. Any thought @d-callan, @dmfalke, and @jtlong3rd ?

d-callan:
im usually on my laptop, and it seems to depend on what variable(s) im plotting. so my best guess is that something on screen is changing width significantly based on label lengths.

moontrip:
@d-callan Thank you for sharing your experience. Now I see what you meant. The table itself is also dynamic, especially "Label" column that shows variable names, so e.g., X-axis of "Age" and "Sleeping rooms in dwelling count" would have different table width, which may go over the limit of left and right arrangement (so place them as top and bottom). 

I am not quite sure whether leaves it as is or the Label column should have a fixed width, which the latter may not be looking  good with long variable name that may be comprised of multiple lines instead of single one

moontrip:
one of difficulty is that this also depends on the zoom level of user's browser

jtlong3rd:
Similar to @moontrip's idea of giving the label column a fixed width:

Maybe we could just change [this style](https://github.com/VEuPathDB/web-eda/blob/800546a2f184e8e34b43b6bfedf8163eff39aab3/src/lib/core/components/visualizations/Visualizations.scss#L161) from `max-width: 500px` to `width: 500px`? That would give the table a fixed width.

Another possibility would be to put the table inside a `div` with fixed width.

moontrip:
@jtlong3rd perhaps setting fixed width: 500px would not cause the sudden change of table location from right to bottom per variable selection at least 👍

moontrip:
we have decided to set a fixed width which would make the table not change its size(width) per variable selection

PR #253, opened 2021-07-26
set variable coverage table width

moontrip:
As discussed in https://github.com/VEuPathDB/web-eda/issues/229, Jamie and I thought we should set a variable coverage table width so that the table does not change with the variable selection. This will alleviate the issue Danielle experienced.


jtlong3rd (approved):
Glad this fix worked!

PR #254, opened 2021-07-26
Ordering of categorical variables for table filter and vizualisations

bobular:
Requires https://github.com/VEuPathDB/web-components/pull/174 

I've started with the `TableFilter` in subsetting.
The fix was relatively easy but I note that the use of `indexOf` is not going to be efficient for variables with very many values.  But is the coding overhead worth it to optimize this?

- [x] Subsetting/TableFilter
- [x] Barplot incl overlay
- [x] Boxplot incl overlay
- [x] Mosaic & 2x2
- [x] Scatter+Lineplot overlay
- [x] Histogram overlay


dmfalke (commented):
I am suspicious of the approach being taken here. I would think we could just order the data going into the plot, and that the plot itself would not need to know anything about `variable.vocabulary`.

dmfalke:
Regarding use of `indexOf`, I think we should not worry about performance until it's a proven issues. The implementation of `indexOf` could very well be optimized by the underlying engine.

bobular:
> I am suspicious of the approach being taken here. I would think we could just order the data going into the plot, and that the plot itself would not need to know anything about `variable.vocabulary`.

I suppose the approach here breaks down for MosaicPlot which has two dependent axes.

Regarding (not using) the plotly-based approach. Sorting all the different data shapes will be a pain. Can discuss in meeting.

bobular:
Hi @dmfalke - reordering the data for one plot type was relatively easy.

I'm torn how to do boxplot next.
1. copy-paste reorderData and add more mappings for q1, q3, mean, median, etc..
2. try to generalise a helper function that will handle BarplotData and BoxplotData

Bearing in mind that MosaicData will need a completely different reordering function, I'm tempted to go for 1!

Would appreciate your opinion

bobular:
Nearly done here.

The boxplot, unlike the barplot, does not have an "empty orange space" where the is no data for "15-29 minutes".  I will see if this can be fixed.

image

bobular:
Solved the "gaps between boxes for missing overlay variables" problem over in the `web-components` PR.

All that remains now is to sort out (pun intended) Mosaic.

image

bobular:
Mosaic done.
image

I'll just merge main and then open up for reviews.

bobular:
Just realised Scatter, Line and Histogram need overlayVariable ordering.

bobular:
Histogram and Scatterplot done (Lineplot done too, but I'm not sure we'll be releasing that in phase 1 - it never makes sense to me)

One small issue with Scatterplot - the legend marker is different for a series with no data.  I'm sure this can be fixed easily - I just ran out of time today. The PR is still good to review!

image

bobular:
All sorted now. Easy fix.

image

bobular:
Oh dear - the overlay sorting completely kills the fitted line modes for ScatterPlot Viz :-(

bobular:
OK, **now** it's sorted - I hope

image

moontrip:
@bobular I have tested boxplot and scatter plot vizs with a couple of examples that I have often used and they worked just fine 👍

moontrip (approved):
I approve barplot, boxplot and scatter plot vizs :)

moontrip:
@bobular By chance I found a potential bug for histogram: could check that it happened at both histogramfilter and histogram viz. When I use `Sleeping rooms in dwelling count` at GEMS1 study, y-axis tick label over 10,000 remained the same without change (See the screenshots below). I also tried to use other variable with more data at the first bin, `Persons sleeping in dwelling count`, and 10,000 labels remain the same.

histogram-issue1

histogramfilter-issue1

histogramfilter-issue2

dmfalke:
I can confirm the bug that @moontrip reported. The count of the first bin is reported as `13638` by the `/distribution` endpoint. The tooltip when mousing over indicates the count is `10000`. It's not clear to me what it going on here.

dmfalke:
I can further confirm that this issue is on `main`. I believe this is happening with other variables as well. The feature site is currently using https://github.com/VEuPathDB/web-eda/releases/tag/0.6.13, so we can probably pinpoint the set of commits that introduced this bug by checking out tags starting with 0.6.14, and testing until we find the tag that introduced it.

It's also possible (and more likely) that this was introduced in https://github.com/VEuPathDB/web-components. We should keep an eye on the version being installed for each of the tags above.

This will need a new issue.

bobular:
Hi @dmfalke and @moontrip - I know what caused this.

There was a request in a meeting to remove the tiny 2 and 5 tickmarks that are shown in log scale.  I did it with a formatting directive in web-components.  I'll make an issue and probably also solve it too.  It's likely that the formatting should only be applied when log scale is on.

dmfalke:
That is some aggressive rounding! Glad it was an easy fix.

dmfalke (approved):
This looks good!

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -145,6 +140,23 @@ function BoxplotViz(props: Props) {
...
+      xAxisVariable: xAxisVariable ? xAxisVariable.variable : undefined,
+      yAxisVariable: yAxisVariable ? yAxisVariable.variable : undefined,
+      overlayVariable: overlayVariable ? overlayVariable.variable : undefined,
+    };
+  }, [
+    findEntityAndVariable,
+    vizConfig.xAxisVariable,
+    vizConfig.yAxisVariable,
+    vizConfig.overlayVariable,
+  ]);
```
moontrip, line 158: @bobular I just wonder if this approach at boxplot viz will be used for all other vizs?

bobular: Yes it might remove some repetitive use of `findEntityAndVariable` in other viz's - but boxplot had the most, so I only did it there so far.

moontrip: @bobular Thank you for your explanation. Since my axis label unit work is also related to this in part, perhaps I can do that for you if you want - the unit work needs to change all vizs as well :) I am waiting for this PR is merged so that I can avoid possible conflicts.

Issue #250, opened 2021-07-23
Completeness table could use consistent and less misleading percentage representation


bobular:
VariableCoverageTable could use the same `toPercentage()` function as used in HistogramFilter header (UnknownCount component)

`import { toPercentage } from 'wdk-client/Components/AttributeFilter/AttributeFilterUtils';`

The screenshot below would then say something like >99.99% instead of 100.00% and <0.01% instead of 0.00%

![image](https://user-images.githubusercontent.com/308639/126841574-9a06de3b-aa95-4ff7-a4e8-f6b90cef6859.png)

PR #256, opened 2021-07-26
changing percentage style at variable coverage table

moontrip:
Following #250 issue, `toPercentage()` is used for variable coverage table. But one question remains that the function, `toPercentage`, is actually returned either `> 99` or `< 1` (see screenshot that was tested at scatterplot viz), not decimal points like `> 99.99` or `< 0.01`. If we need to support more precise decimal points like Bob mentioned in the issue, then we can probably make a function that is similar to `toPercentage()`. Just let me know @bobular :)

scatter-percent1

### DIFF
--- a/src/lib/core/components/VariableCoverageTable.tsx
+++ b/src/lib/core/components/VariableCoverageTable.tsx
@@ -33,10 +33,10 @@ export interface VariableCoverageTableRow {
   completeCount?: number;
   /* absolute count of entities which are missing the variable */
   incompleteCount?: number;
-  /* percent of subset entities which have the variable */
-  completePercent?: number;
-  /* percent of subset entities which are missing the variable */
-  incompletePercent?: number;
+  /* percent of subset entities which have the variable: allowing string like '< 1', '> 99' */
+  completePercent?: number | string;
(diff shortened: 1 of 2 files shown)


jtlong3rd (approved):

moontrip:
@bobular After testing some cases, I think that >99% and <1% seems to be quite reasonable values we may use in general situations: two decimal points are too small to satisfy. For example, the exact Data percentage of above screenshot (scatter plot one I have uploaded) is 99.969 and 99.978. If we use two decimal points like 99.99 in this case, which would have a criterion of larger than 99.994, then it would be shown as 100%.

Alternatively, I think that we may use Math.floor, which would prevent the issue of showing 100% with very small number of 'No Data'?

moontrip:
sorry mistakenly pressed Close with comment :(

bobular:
Sorry DK - I mistakenly remembered the function returning decimal places (e.g. 99.99%).  You are right that the current behaviour should suit us just fine.

bobular (approved):
Works well (sorry I should have said that my example was from PRISM1)
And the code looks great.
Thanks

Issue #240, opened 2021-07-19
Visualizations: Use first element of `dataElementDependencyOrder` as output variable


dmfalke:
If `dataElementDependencyOrder` is not provided, use the first variable displayed. This should be the only variable, but we don't have a way to enforce it, so we will need to be defensive. It might be a good idea to have a small utility function to encapsulate this logic.

See https://github.com/VEuPathDB/EdaDataService/issues/41 for more context.

Issue #238, opened 2021-07-16
show viz descriptions


d-callan:
the apps endpoint of the data service provides a description, in a single sentence currently, of each viz. in a slack conversation @dmfalke said wrt display "it could be in a tooltip in the viz selector, and inline in the viz itself"

moontrip:
@d-callan I have addressed the two cases, tooltip at viz selector and viz itself. Please see below two screenshots where description at viz was added next to viz name, e.g., "Scatter plot: Visualize the relationship between two continuous variables" under changeable title (Unnamed Visualization). Are they alright for you?

![scatter-thumbnail0](https://user-images.githubusercontent.com/12802305/127326929-bec05fcc-412e-4fef-8aec-f38daf79b1ec.png)

![scatter-depend0](https://user-images.githubusercontent.com/12802305/127326980-38219517-5eb7-40be-81d6-998bf4b4b76e.png)

d-callan:
Looks great. Thanks!

Issue #227, opened 2021-07-15
show var completeness table when plot data is empty


d-callan:
just tried to plot `sleeping rooms in dwelling count` against `axillary temperature` in a scatter plot for gems1 case control. the response indicates all values are missing for axillary temperature (not sure yet if thats the data or a back end bug, but thats besides the point currently). i think we should show the completeness of the vars even when there is no plottable data.

moontrip:
With the update by Danielle, now I can test this case however it is not clear to me what the issue is exactly. The table is always shown regardless of data or no data, so perhaps your question is to show values in the table (Data and No Data). But your example, sleeping rooms in dwelling count vs axillary temparature throws 500 server error (backend issue) and no proper response is given to the frontend side (like CompleteCasesTable response). Thus it is not clear whether no data will work properly.

@jtlong3rd who made the nice table can answer this more accurately but my gut feeling is that if the response is properly given to the frontend without 500 error, the table would show values, e.g., Data = 0 (0 %), No Data = 10,000 (100 %). By the way, @jtlong3rd, your nice table seems to stop working after recent backend change. Please see [this draft PR](https://github.com/VEuPathDB/web-eda/pull/246) made by Bob whether it would be helpful for you to identify an issue

jtlong3rd:
> By the way, @jtlong3rd, your nice table seems to stop working after recent backend change. Please see [this draft PR](https://github.com/VEuPathDB/web-eda/pull/246) made by Bob whether it would be helpful for you to identify an issue

I'll take a look! I suspect that this is related to the change in the variable id format.

moontrip:
I think this is closely related to #240 so I will also check this as well.

moontrip:
@d-callan After addressing dataElementDependencyOrder (#240), I could test your case, `sleeping rooms in dwelling count` against `axillary temperature` at scatter plot viz. Yes I could see there are many no data for `axillary temperature`, about 30 %, but plot and table are showed accordingly. Thus, as I presumed, they seem to work just fine if the backend responds with data (without server error). Do you think your issue is addressed ?

![scatter-depend0](https://user-images.githubusercontent.com/12802305/127327466-6b632b0f-b3fe-48c2-9198-7c5024e0e788.png)

d-callan:
For now maybe.. I have some lingering concerns but I can open a new issue if I find a good example of one of them.

PR #258, opened 2021-07-28
following dependency order and showing viz description

moontrip:
This PR includes dependency order issue (#240) and viz description (#238): possibly addressing another issues like #227 and [EDADataService one](https://github.com/VEuPathDB/EdaDataService/issues/41) in this work. 

For dependency order, as of now four vizs that have X/Y variables such as scatter, line, mosaic, and box plot vizs are the targets to be addressed. Also realized that grid view/thumbnail did not have `dataElementDependencyOrder` props, thus it was required to pass it to the grid view/thumbnail (at `VisualizationsContainer`). I have known two cases at least (see Figures 1 and 2) that threw errors for specific combination of variables, which can be reproduced at feature site. After following the dependency order, it seems to work properly now.

Regarding viz description, `VisualizationsContainer` component was changed to allow the features such as tooltip for variable picker (see Figure 3) and description at viz. As of now, I have added viz description at the next of each viz's name, e.g., for scatter plot viz, it was implemented as "Scatter plot: Visualize the relationship between two continuous variables": refer to Figures 1 (scatter plot) and 2 (mosaic plot). 

**Figure 1**
scatter-depend0

**Figure 2**
mosaic-depend0

**Figure3**
scatter-thumbnail0

### DIFF
--- a/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -95,6 +95,7 @@ function ConfiguredVisualizations(props: Props) {
     [computationId, visualizations]
   );
   if (computation == null) return <div>Computation not found</div>;
+
   return (
     <Grid>
       <Link replace to={`${url}/new`} className={cx('-NewVisualization')}>
@@ -105,6 +106,8 @@ function ConfiguredVisualizations(props: Props) {
         .map((viz) => {
           const type = visualizationTypes[viz.type];
           const meta = visualizationsOverview.find((v) => v.name === viz.type);
+          // dataElementDependencyOrder should be passed for grid view/thumbnail
+          const dataElementDependencyOrder = meta?.dataElementDependencyOrder;
           return (
             <div key={viz.id}>
               <div className={cx('-ConfiguredVisualization')}>
@@ -154,6 +157,8 @@ function ConfiguredVisualizations(props: Props) {
                     filters={filters}
                     starredVariables={starredVariables}
                     toggleStarredVariable={toggleStarredVariable}
+                    // dataElementDependencyOrder should be passed for grid view/thumbnail
+                    dataElementDependencyOrder={dataElementDependencyOrder}
                   />
                 ) : (
                   <div>Visualization type not implemented: {viz.type}</div>
@@ -203,27 +208,30 @@ function NewVisualizationPicker(props: Props) {
               className={cx('-PickerEntry', vizType == null && 'disabled')}
               key={`vizType${index}`}
             >
-              <button
-                type="button"
-                disabled={vizType == null}
-                onClick={async () => {
-                  const id = uuid();
-                  addVisualization({
-                    id,
-                    computationId: computationId,
-                    type: vizOverview.name!,
-                    displayName: 'Unnamed visualization',
-                    configuration: vizType?.createDefaultConfig(),
-                  });
-                  history.replace(`../${computationId}/${id}`);
-                }}
-              >
-                {vizType ? (
-                  <vizType.selectorComponent {...vizOverview} />
-                ) : (
-                  <PlaceholderIcon name={vizOverview.name} />
-                )}
-              </button>
+              {/* add viz description tooltip for viz picker */}
+              <Tooltip title={<>{vizOverview.description}</>}>
+                <button
+                  type="button"
+                  disabled={vizType == null}
+                  onClick={async () => {
+                    const id = uuid();
+                    addVisualization({
+                      id,
+                      computationId: computationId,
+                      type: vizOverview.name!,
+                      displayName: 'Unnamed visualization',
+                      configuration: vizType?.createDefaultConfig(),
+                    });
+                    history.replace(`../${computationId}/${id}`);
+                  }}
+                >
+                  {vizType ? (
+                    <vizType.selectorComponent {...vizOverview} />
+                  ) : (
+                    <PlaceholderIcon name={vizOverview.name} />
+                  )}
+                </button>
+              </Tooltip>
               <div className={cx('-PickerEntryName')}>
                 <div>{vizOverview.displayName}</div>
               </div>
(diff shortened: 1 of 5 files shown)


dmfalke:
This looks great. Let's discuss the UI at our dev meeting today. I'll review the code soon.

moontrip:
@dmfalke Thank you for your quick view 😃 I know that there are repetitive parts, one is at variableCoverageTable's prop and the other is at getRequestParams function. The former may be used as a util function (with using bracket []) but the latter is not possible (as far as I know) as it needs to access object name (if it is converted to use a util function). Since they are just three vizs/cases, I just made them verbose and repetitive...

jtlong3rd (changes requested):
Looks good so far! You'll also need to pass the output entity (the entire entity, not just its id) to the `OutputEntityTitle`. Since the output entity is being used in three different places for these visualizations, I would consider deriving the `outputEntityId` and `outputEntity` in a single spot per visualization, perhaps in some sort of `useMemo` with `dataElementDependencyOrder`, `vizConfig`, and `findEntityAndVariable` as dependencies.

moontrip:
@jtlong3rd Thank you for your review and suggestion 👍 In the following commit, I have only worked for scatter plot viz with your suggestion and want to make sure if it is what you thought. If right, then I will work for other vizs too. 

https://github.com/VEuPathDB/web-eda/pull/258/commits/6675a1339ff0fe899eddade2ec4cee342ab9b0c3

jtlong3rd:
> @jtlong3rd Thank you for your review and suggestion 👍 In the following commit, I have only worked for scatter plot viz with your suggestion and want to make sure if it is what you thought. If right, then I will work for other vizs too.
> 
> [6675a13](https://github.com/VEuPathDB/web-eda/commit/6675a1339ff0fe899eddade2ec4cee342ab9b0c3)

I like it! One suggestion: we actually *don't* need a `useMemo` for `outputEntityId`. The reason is that the `outputEntity` has an `id` property, so for the two place where you need the `outputEntityId`, you can just pass `outputEntity.id`.

moontrip:
> > @jtlong3rd Thank you for your review and suggestion 👍 In the following commit, I have only worked for scatter plot viz with your suggestion and want to make sure if it is what you thought. If right, then I will work for other vizs too.
> > [6675a13](https://github.com/VEuPathDB/web-eda/commit/6675a1339ff0fe899eddade2ec4cee342ab9b0c3)
> 
> I like it! One suggestion: we actually _don't_ need a `useMemo` for `outputEntityId`. The reason is that the `outputEntity` has an `id` property, so for the two places where you need the `outputEntityId`, you can just pass `outputEntity.id`.

@jtlong3rd 😮 yes you are absolutely right! I was too focused on the two things separately 😄 I will remove that `outputEntityId` 👍

moontrip:
In conjunction with updates of @jtlong3rd 's suggestion to all relevant vizs, a couple of updates are additionally made: 

a) fixed scatter plot viz bug for y-axis range in some cases (#262): changed some scheme for computing yMin/yMax
b) removed html tags from scatter plot viz's legend texts for [legend ellipsis with tooltip](https://github.com/VEuPathDB/web-components/pull/170)  

Screenshots of the above two are attached here:

scatter-range-fix1

scatter-legend-tag

src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -328,7 +336,10 @@ function FullScreenVisualization(props: Props & { id: string }) {
               }
             />
           </h3>
-          <div className="Subtitle">{overview?.displayName}</div>
+          {/* add viz description */}
+          <div className="Subtitle">
+            {overview?.displayName}: {overview?.description}
+          </div>
```
dmfalke, line 342: We agreed to revert this today.

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -144,6 +144,16 @@ function BoxplotViz(props: Props) {
...
 
+  // outputEntity for OutputEntityTitle's outputEntity prop and outputEntityId at getRequestParams
+  const outputEntity = useMemo(() => {
+    const outputEntityVariableName =
+      dataElementDependencyOrder != null &&
+      dataElementDependencyOrder[0] === 'yAxisVariable'
+        ? vizConfig.yAxisVariable
+        : vizConfig.xAxisVariable;
+    return findEntityAndVariable(outputEntityVariableName)?.entity;
+  }, [dataElementDependencyOrder, vizConfig, findEntityAndVariable]);
```
dmfalke, line 155: I think this is the repetitive code you were referring to. Maybe it can be generalized into a hook like this:

```typescript
function useFindOutputEntity(
  dataElementDependencyOrder?: string[],
  dataElementVariables: Record<string, VariableDescriptor | undefined>,
  defaultVariableName?: string,
  entities: StudyEntity[]
): EntityAndVariable | undefined {
  const findEntityAndVariable = useFindEntityAndVariable(entities);
  return useMemo(
    () => {
      const variableName = (dataElementDependencyOrder == null || dataElementDependencyOrder.length === 0)
        ? defaultVariableName
        : dataelementDependencyOrder[0];
      const variable = dataElementVariables[variableName];
      return findEntityAndVariable(variable)?.entity;
    },
    [dataElementDependencyOrder, dataElementVariables, defaultValue, entities ]
  );
}
```

It could be used like this:
```typescript
const outputEntity = useFindOutputEntity(
  dataElementDependencyOrder,
  vizConfig,
  'xAxisVariable',
  entities
);
```

Keep in mind that this is untested code, so some modifications may be necessary.

dmfalke (commented):
The code looks good. I added a suggestion for how to reduce some of the repetition, and a change based on UI feedback we received today.

jtlong3rd (approved):
It's definitely worth giving Dave's `useFindOutputEntity` suggestion a shot, but everything else looks good!

src/lib/core/components/visualizations/VisualizationsContainer.tsx
moontrip: thank you for reminding me and removed :)

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
moontrip: Thank you for suggestion! 👍 After battling ts issues (haha), I could update it

dmfalke:
@moontrip feel free to skip my suggesting for extracting that logic. We can always address later.

moontrip:
> @moontrip feel free to skip my suggesting for extracting that logic. We can always address later.

@dmfalke too late as I already addressed your suggestion in recent commits (an hour ago) 😄

dmfalke (approved):

PR #267, opened 2021-08-03
adding unit at axis label

moontrip:
made a function to return axis label with unit if exists: for now, only variable with number type will show unit(s) in axes. As an example, only scatter plot viz is used to test: if acceptable then I will apply this to other vizs. A screenshot is attached for reference. Also, variable selector for the facet variable is removed as shown in the screenshot.

axislabelunit1
.


src/lib/core/utils/axisLabelUnit.ts
```diff
@@ -0,0 +1,15 @@
+import { Variable } from '../types/study';
```
jtlong3rd, line 1: Nit: name this file `axis-label-unit`, to be consistent with the names of the other files in `utils`.

jtlong3rd (commented):
Aside from one small inline comment, I think this is good to apply to the other viz"es"!

src/lib/core/utils/axisLabelUnit.ts
moontrip: Thank for you review 👍 I have changed the file name

dmfalke (commented):
This looks good to me. Feel free to proceed with other visualizations!

src/lib/core/utils/axis-label-unit.ts
```diff
@@ -0,0 +1,15 @@
+import { Variable } from '../types/study';
+
+export function axisLabelUnit(
```
bobular, line 3: Would you mind renaming this `axisLabelWithUnit` ?

moontrip: @bobular No problem! I can change the function name :)

moontrip: @bobular I have changed the name as suggested

moontrip:
@jtlong3rd and @dmfalke I have updated this PR by addressing a couple of things as follows:

a) changed function name following Bob's suggestion
b) simplified the use of `findEntityAndVariable` function for all vizs by referring to what Bob did for boxplot viz. Tested several cases to ensure that they are working fine with changes
c) added axis label unit to all vizs
d) facet variable selector is removed at all vizs

jtlong3rd (approved):

dmfalke (approved):

PR #281, opened 2021-08-06
Fully resolve path to new anlaysis to prevent incorrect parsing

dmfalke:
Prior to this PR, when in an analysis, the **New analysis** button would link to a route such as `/eda/DS_841a9f5259/34/../new`. **react-router** would  match on this exact string, rather than the resolved path of `/eda/DS_841a9f5259/new`. The fix is to fully resolve the path before passing to the `Link` component.


dmfalke:
I looked to see if there was an issue for this in https://github.com/ReactTraining/react-router, but couldn't find one.

jtlong3rd (approved):

PR #293, opened 2021-08-10
All analyses table

### LINKED ISSUE #292: "All Analyses" table
The root route for the EDA workspace should display a table listing a user's analyses. The table should include the ability to delete analyses, and to pin analyses to the top of the table.

dmfalke:
### Summary
This PR introduces the **All Analyses** table, including the ability to "pin" analyses to the top of the table, and to remove unpinned analyses.

### Detailed changes
* Methods for getting/setting Analyses preferences
* Method to delete multiple analyses
* `useAnalysisList` React hook
* `usePinnedAnalyses` React hook
* `AllAnalyses` React Component
* Modified root route to render `AllAnalyses`

### Related issues
- closes #292


bobular:
Hi @dmfalke 
I appreciate that the entry point to each study might be from the home page cards or other site navigation (menus), but if we merge this branch, we (developers) will have no way to start a new analysis for a study that's not already listed in the My Analyses table.

Can we bring back the list of studies?  Even if just for devs.

dmfalke:
@bobular, yes, that's a good idea. I will add that.

src/lib/workspace/AllAnalyses.tsx
```diff
@@ -0,0 +1,338 @@
...
+  FormControlLabel,
+  Icon,
+  makeStyles,
+  Switch,
+  ThemeProvider,
+  Tooltip,
+} from '@material-ui/core';
+import { Loading, Mesa } from '@veupathdb/wdk-client/lib/Components';
+import { ContentError } from '@veupathdb/wdk-client/lib/Components/PageStatus/ContentError';
+import { useSessionBackedState } from '@veupathdb/wdk-client/lib/Hooks/SessionBackedState';
```
bobular, line 14: What's the difference between this and the localforage thing we're using in Mocks.ts ?

src/lib/workspace/WorkspaceRouter.tsx
```diff
@@ -38,6 +47,11 @@ export function WorkspaceRouter({
           <Redirect to={location.pathname.replace(/\/\/+/g, '/')} />
         )}
       />
+      <Route
+        path={`${path}/studies`}
+        exact
+        render={() => <StudyList subsettingServiceUrl={subsettingServiceUrl} />}
```
bobular, line 53: When called from this route, StudyList is rendering links with URLs like `/eda/studies/DS_841ea82`, but they just need to be `/eda/DS_841ea82`

src/lib/workspace/AllAnalyses.tsx
dmfalke: `localforage` uses a more persistent storage by default (typically [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API)). `useSessionBackedState` uses a storage that gets cleared when the tab/window is closed ([sessionState](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)).

src/lib/workspace/WorkspaceRouter.tsx
dmfalke: Ah, right. I will fix.

src/lib/workspace/AnalysisList.tsx
```diff
@@ -14,16 +20,11 @@ export function AnalysisList(props: Props) {
...
-  const updateAnalysisList = React.useCallback(async () => {
-    const list = await analysisStore.getAnalyses();
-    setAnalysisList(list.filter((a) => a.studyId === studyId));
-  }, [analysisStore, studyId]);
-  React.useEffect(() => {
-    updateAnalysisList();
-  }, [updateAnalysisList]);
+  const { analyses: analysisList, deleteAnalyses } = useAnalysisList(
+    analysisStore
+  );
```
bobular, line 27: Need to reinstate the filter on studyId 
```suggestion
  const { analyses, deleteAnalyses } = useAnalysisList(
    analysisStore
  );
  const analysisList = analyses ? analyses.filter((analysis) => analysis.studyId === studyId) : undefined;
```

src/lib/workspace/AllAnalyses.tsx
```diff
@@ -0,0 +1,338 @@
...
+            [
+              (analysis) => (isPinnedAnalysis(analysis.id) ? 0 : 1),
+              (analysis) => {
+                const columnKey = tableSort[0];
+                switch (columnKey) {
+                  case 'study':
+                    return (
+                      datasets?.records.find(
+                        (d) => d.id[0].value === analysis.studyId
+                      )?.displayName ?? 'Unknown study'
```
bobular, line 124: There's a second or so's delay when I reload this page, and "Unknown study" is a bit disconcerting.

Could this work?
```suggestion
                      )?.displayName ?? 'Loading...'
```

bobular: Actually this is the wrong place to do this (see below)

src/lib/workspace/AllAnalyses.tsx
```diff
@@ -0,0 +1,338 @@
...
+        },
+        {
+          key: 'study',
+          name: 'Study',
+          sortable: true,
+          renderCell: (data: { row: Analysis }) => {
+            const dataset = datasets?.records.find(
+              (d) => d.id[0].value === data.row.studyId
+            );
+            if (dataset == null) return 'Unknown study';
```
bobular, line 297:
```suggestion
            if (dataset == null) return 'Loading...';
```

src/lib/workspace/AllAnalyses.tsx
```diff
@@ -0,0 +1,338 @@
...
+            return (
+              <Link to={`${url}/${dataset.id[0].value}`}>
+                {dataset.displayName}
```
bobular, line 300: SCORE S. mansoni italics tags are currently showing.  The approach in StudyList can be used here to clean the HTML.

```
import { safeHtml } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
```

```suggestion
                {safeHtml(dataset.displayName)}
```

bobular (approved):
Did some thorough testing and found the bug with the study-page's AnalysisList.  Fixes and suggestions provided above/below.

src/lib/workspace/AllAnalyses.tsx
dmfalke: I'd prefer to defer rendering of the table until the datasets are loaded. I will update in a bit.

src/lib/workspace/AllAnalyses.tsx
dmfalke: Thanks, this is good! I will update soon.

jtlong3rd (approved):

dmfalke:
I've made some updates to this branch:
- Updated datetime formatting to be consistent with other ClinEpiDB pages
- Rearranged layout of table actions
- Added search box for table
- Added confirmation modal for delete actions
- Only render after datasets have been loaded to prevent **Unknown study** in all rows

jtlong3rd (approved):

jtlong3rd (approved):

jtlong3rd:
> Just pushed a fix for some type bugs. This is a result of `Mesa` being up-typed 🙁,

Story of my (VEuPathDB) life!

dmfalke:
> > Just pushed a fix for some type bugs. This is a result of `Mesa` being up-typed slightly_frowning_face,
> 
> Story of my (VEuPathDB) life!

word

bobular:
> Only render after datasets have been loaded to prevent **Unknown study** in all rows

This is great, but it now needs a progress spinner IMO!

bobular:
> > Only render after datasets have been loaded to prevent **Unknown study** in all rows
> 
> This is great, but it now needs a progress spinner IMO!

Hmm, I see there's this
```
        {(loading || datasets == null) && <Loading />}
```
It's nested inside the `{analyses && datasets && ( ...` so needs to go outside of this.

src/lib/workspace/AllAnalyses.tsx
```diff
@@ -0,0 +1,423 @@
...
+              />
+              <span>
+                Showing {filteredAnalysesAndDatasets?.length} of{' '}
+                {analyses.length} analyses
+              </span>
+            </div>
+            {(loading || datasets == null) && <Loading />}
+          </Mesa.Mesa>
+        )}
+      </div>
```
bobular, line 402:
```suggestion
      <div className={classes.root}>
        <h1>My Analyses</h1>
        {error && <ContentError>{error}</ContentError>}
        {(loading || datasets == null) && <Loading />}
        {analyses && datasets && (
          <Mesa.Mesa state={tableState}>
            <div
              style={{
                display: 'flex',
                alignItems: 'center',
                gap: '1ex',
              }}
            >
              <TextField
                variant="outlined"
                size="small"
                label="Search analyses"
                inputProps={{ size: 50 }}
                value={searchText}
                onChange={(e) => setSearchText(e.target.value)}
              />
              <span>
                Showing {filteredAnalysesAndDatasets?.length} of{' '}
                {analyses.length} analyses
              </span>
            </div>
          </Mesa.Mesa>
        )}
      </div>
```

bobular: Sorry, that's a much bigger-looking diff than it really is!

I moved the `<Loading />` line up to the top :-)

bobular (approved):
Looks great - just one tiny fix suggested.

dmfalke:
@bobular I made a change close to what you suggested. I'm going to go ahead and merge this now :slightly_smiling_face: .

PR #296, opened 2021-08-10
truncated X-axis

moontrip:
@bobular I have made a draft for truncated X-axis but need your opinions as you know much better than me for histogram related stuff 😃 Temporarily the truncated X-axis area is filled with gray color: it is not a good choice of color though. A screenshot is attached in the end. Please note that this PR replies on [another PR at web-components repo](https://github.com/VEuPathDB/web-components/pull/186).

As of now, there are two major issues that blocked me to go further.

a) there are the backend information for X-axis (independent axis) like data's min/max/display.min/max, etc. thus those info can be used for setting both plot and data ranges separately. Basically the approach I am taking is that layout.xaxis.range is set to data's min/max (from response) thus data response from X-axis control can be used to show truncated data. However, this is not the case for Y-axis (dependent axis) as it is simply using layout.yaxis.range for Y-axis control: no info/stats from backend response nor specific data request for y-axis. Any thought?

b) hatching area: I tried several ways to fill those truncated X-axis area with a pattern like hatching but it is quite a daunting task from my understanding: layout.image was also attempted but it was not that feasible for our purpose. Plotly does not allow to use a pattern for those area natively, thus perhaps a possible solution is to use onAfterPlot event to make patterns manually but it is quite tricky to do that with limited access to plotly's plot. Any idea?

An additional note is that I tried to show an info message below the range control whenever X-axis range control changes, something similar to your green warning message. 

histogramfilter1

### Related issues
- https://github.com/VEuPathDB/web-eda/issues/98 (closes)


dmfalke:
@moontrip please update the description of this PR so that issues are not automatically linked. This causes github to close those issues when the PR is merged, which is behavior we do not want, given the new project flow.

I recommend adding a section at the bottom of the description, like this:
----
### Related issues
- https://github.com/VEuPathDB/web-eda/issues/98 (closes)

moontrip:
new works are made in the new PR, #346 , thus closing this PR

Issue #286, opened 2021-08-06
Visualizations: "Only show compatible variables" should hide non-compatible featured variables


SheenaTomko:
When choosing variables to plot, toggling on "Only show compatible variables" should hide all non-compatible featured variables

aaronwlsong:
**QA Checking:**

On QA website, the toggle of "only show compatible variables" works, when click it, all non-compatible variables are hidden. 

Before click the toggle:

<img width="922" alt="WeChat160a43f0ebfad130659f72b29b89ce80" src="https://user-images.githubusercontent.com/61077845/130120274-23dc2794-5ff2-4725-94e6-fe9ce1dea3c4.png">

After click the toggle:

<img width="976" alt="WeChata3deed07ac0c84ef91a42fe4ce132072" src="https://user-images.githubusercontent.com/61077845/130120366-54584522-ea98-420c-b7cf-a27607ad4101.png">

Issue #268, opened 2021-08-03
Featured variables: Move above variable tree


dmfalke:

danicahelb:
looks good, closing ticket

PR #299, opened 2021-08-10
Move featured vars and compat vars toggle to above tree

dmfalke:
The PR moves feature variables and the compatible variables toggle to above the variable tree. It also simplifies some styling used to make the variables in the tree scrollable.

### Screenshot
image

### Related issue
#268
#286

### DIFF
--- a/src/lib/core/components/VariableList.tsx
+++ b/src/lib/core/components/VariableList.tsx
@@ -301,121 +301,11 @@ export default function VariableList(props: VariableListProps) {
     ]
   );
 
-  const wrapTreeSection = useCallback(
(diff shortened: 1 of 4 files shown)


jtlong3rd (approved):

bobular (approved):
Looks good/works well.

I'm feeling that the UI could still be tweaked a bit.

The viz drop-downs are clear that they are variable pickers, but in subsetting, the left hand panel is less obviously *all* about choosing variables.  Possibly have a "Choose a variable" heading and then two subheadings: "Featured variables" and "All variables".

bobular:
Much cleaner with the new spacing 👍

PR #307, opened 2021-08-13
Switch for showing missing data for overlay variable [DEPRECATED]

### LINKED ISSUE #294: Visualizations: Allow users to show missing data for stratification variables
See [mockups](https://docs.google.com/presentation/d/14WwD95etPuj6MS0wHQZ9_2X8CbEQrdhw/edit#slide=id.ge5d990b9bf_0_158). Need to allow users to toggle on/off missing data for the stratification variables. The statement "Include points with no data for selected stratification variables" should update to reflect what entity level is being plotted.

bobular:
Fixes #294 

Requires https://github.com/VEuPathDB/web-components/pull/189

Not quite following the spec [mocked up here](https://docs.google.com/presentation/d/14WwD95etPuj6MS0wHQZ9_2X8CbEQrdhw/edit#slide=id.gb82e9eac95_0_121) in order to get something out more quickly for phase 1 (no faceting).

I added a `wrapper` prop to the inputs given to InputVariables, so that a box and a switch could be added to **one** of the input variable widgets.  In future if we group overlay+facet together we'll need to do something a bit more sophisticated.

As usual, I'm not pretending to be a styling and layout expert.

image

To do still:

- [ ] missing data needs to be grey (will talk to @chowington !)
- [ ] roll out to other viz's
- [ ] style/layout

### DIFF
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -17,6 +17,7 @@ import {
 interface InputSpec {
   name: string;
   label: string;
+  wrapper?: React.FunctionComponent;
 }
 
 export interface Props {
@@ -180,22 +181,27 @@ export function InputVariables(props: Props) {
   return (
     <div>
       <div className={classes.inputs}>
-        {inputs.map((input, index) => (
-          <div key={input.name} className={classes.input}>
-            <div className={classes.label}>{input.label}</div>
-            <VariableTreeDropdown
-              rootEntity={entities[0]}
-              disabledVariables={disabledVariablesByInputIndex[index]}
-              starredVariables={starredVariables}
-              toggleStarredVariable={toggleStarredVariable}
-              entityId={values[input.name]?.entityId}
-              variableId={values[input.name]?.variableId}
-              onChange={(variable) => {
-                handleChange(input.name, variable);
-              }}
-            />
-          </div>
-        ))}
+        {inputs.map((input, index) => {
+          const Wrapper = input.wrapper ?? (({ children }) => <>{children}</>);
+          return (
+            <Wrapper>
+              <div key={input.name} className={classes.input}>
+                <div className={classes.label}>{input.label}</div>
+                <VariableTreeDropdown
+                  rootEntity={entities[0]}
+                  disabledVariables={disabledVariablesByInputIndex[index]}
+                  starredVariables={starredVariables}
+                  toggleStarredVariable={toggleStarredVariable}
+                  entityId={values[input.name]?.entityId}
+                  variableId={values[input.name]?.variableId}
+                  onChange={(variable) => {
+                    handleChange(input.name, variable);
+                  }}
+                />
+              </div>
+            </Wrapper>
+          );
+        })}
       </div>
       {/* <div className={`${classes.label} ${classes.dataLabel}`}>Data inputs</div> */}
     </div>
--- a/src/lib/core/components/visualizations/MissingDataToggleWrapper.tsx
+++ b/src/lib/core/components/visualizations/MissingDataToggleWrapper.tsx
@@ -1,0 +1,46 @@
+import React, { useCallback } from 'react';
+import { StudyEntity, Variable } from '../../types/study';
+import { makeEntityDisplayName } from '../../utils/study-metadata';
+import Switch from '@veupathdb/components/lib/components/widgets/Switch';
+
+interface MissingDataToggleWrapperProps {
+  state: boolean | undefined;
+  onStateChange: (newState: boolean) => void;
+  outputEntity: StudyEntity | undefined;
+  overlayVariable: Variable | undefined;
+}
+
+export default function useMissingDataToggleWrapper({
+  state,
+  onStateChange,
+  outputEntity,
+  overlayVariable,
+}: MissingDataToggleWrapperProps) {
+  return useCallback(
+    ({ children }) => {
+      return (
+        <div
+          style={{
+            border: '2px solid #f0f0f0',
+            padding: '3px',
+            display: 'flex',
+            flexDirection: 'row',
+          }}
+        >
+          {children}
+          <Switch
+            label={`Include ${
+              outputEntity
+                ? makeEntityDisplayName(outputEntity, true)
+                : 'points'
+            } with no data for selected overlay variable`}
+            state={state}
+            onStateChange={onStateChange}
+            disabled={overlayVariable == null}
+          />
+        </div>
+      );
+    },
+    [outputEntity, state, overlayVariable, onStateChange]
+  );
+}
--- a/src/lib/core/api/data-api.ts
+++ b/src/lib/core/api/data-api.ts
@@ -80,6 +80,7 @@ export interface HistogramRequestParams {
       xMin: string;
       xMax: string;
     };
+    showMissingness?: 'TRUE' | 'FALSE';
   };
 }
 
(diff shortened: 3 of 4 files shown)


src/lib/core/components/visualizations/InputVariables.tsx
```diff
@@ -180,22 +181,27 @@ export function InputVariables(props: Props) {
...
-              entityId={values[input.name]?.entityId}
-              variableId={values[input.name]?.variableId}
-              onChange={(variable) => {
-                handleChange(input.name, variable);
-              }}
-            />
-          </div>
-        ))}
+        {inputs.map((input, index) => {
+          const Wrapper = input.wrapper ?? (({ children }) => <>{children}</>);
```
bobular, line 185: Is there a more elegant way to do the 'no op' wrapper?

src/lib/core/components/visualizations/InputVariables.tsx
```diff
@@ -17,6 +17,7 @@ import {
 interface InputSpec {
   name: string;
   label: string;
+  wrapper?: React.FunctionComponent;
```
jtlong3rd, line 20: I think this approach of adding a "wrapper" to the `InputSpec` could work!

It might be more idiomatic to express the wrapping as `InputComponent: React.ComponentType<InputProps>` or `renderInput: (props: InputProps) => React.ReactNode` on the `InputSpec`.

(Here, `InputProps` is the current props interface for the `VariableTreeDropdown`.)

One benefit of this approach is a wrapper will have access to the `InputProps` passed via `InputVariables`, which the wrapper might need as part of its custom rendering.

src/lib/core/components/visualizations/InputVariables.tsx
jtlong3rd: If you adopt my suggestion of giving `InputSpec` an `InputComponent`, this line could be changed to something like this:

```suggestion
          const InputComponent = input.InputComponent ?? VariableTreeDropdown;
```

src/lib/core/components/visualizations/InputVariables.tsx
```diff
@@ -180,22 +181,27 @@ export function InputVariables(props: Props) {
...
+                  starredVariables={starredVariables}
+                  toggleStarredVariable={toggleStarredVariable}
+                  entityId={values[input.name]?.entityId}
+                  variableId={values[input.name]?.variableId}
+                  onChange={(variable) => {
+                    handleChange(input.name, variable);
+                  }}
+                />
+              </div>
+            </Wrapper>
```
jtlong3rd, line 202: If you adopt my suggestion of giving `InputSpec` an `InputComponent`, you would do something like this:

```suggestion
            <InputComponent
              key={input.name}
              rootEntity={entities[0]}
              disabledVariables={disabledVariablesByInputIndex[index]}
              starredVariables={starredVariables}
              toggleStarredVariable={toggleStarredVariable}
              entityId={values[input.name]?.entityId}
              variableId={values[input.name]?.variableId}
              onChange={(variable) => {
                handleChange(input.name, variable);
              }}
            />
```

src/lib/core/components/visualizations/MissingDataToggleWrapper.tsx
```diff
@@ -0,0 +1,46 @@
...
+                ? makeEntityDisplayName(outputEntity, true)
+                : 'points'
+            } with no data for selected overlay variable`}
+            state={state}
+            onStateChange={onStateChange}
+            disabled={overlayVariable == null}
+          />
+        </div>
+      );
+    },
```
jtlong3rd, line 43: If you adopt my suggestion of giving `InputSpec` an `InputComponent`, you could do something like this:
```suggestion
    (props: InputProps) => {
      return (
        <div
          style={{
            border: '2px solid #f0f0f0',
            padding: '3px',
            display: 'flex',
            flexDirection: 'row',
          }}
        >
          <VariableTreeDropdown {...props} />
          <Switch
            label={`Include ${
              outputEntity
                ? makeEntityDisplayName(outputEntity, true)
                : 'points'
            } with no data for selected overlay variable`}
            state={state}
            onStateChange={onStateChange}
            disabled={overlayVariable == null}
          />
        </div>
      );
    },
```

src/lib/core/components/visualizations/InputVariables.tsx
bobular: Thanks Jamie - I'm having a go at this now.

When `InputComponent !== VariableTreeDropdown`, how do I pass the extra props that the extended component needs?

e.g. when `InputComponent` is `VariableTreeDropdownWithToggle` I need to pass these extra props:
```
interface Props extends VariableTreeDropdownProps {
  /** the text label to show with the toggle switch */
  toggleLabel: string;
  /** the controlled state of the toggle switch */
  toggleState: boolean | undefined;
  /** handler for toggle switch state change */
  onToggleStateChange: (newState: boolean) => void;
  /** optional flag to disble the toggle switch */
  disableToggle?: boolean
}
```
It seems that maybe
```
interface InputSpec {
  name: string;
  label: string;
  component?: React.FunctionComponent;
}
```
should include a `componentExtraProps: any` but that gets nasty...

dmfalke (commented):
This is a good start. However, I'm not sure I favor the wrapper approach. I was thinking `InputSpec` could include a property called `hasMissingDataToggle`.

The associated [`ValueByInputName`](https://github.com/VEuPathDB/web-eda/blob/5649a0da4a40f1a78ebe234a6e263d331d00d54d/src/lib/core/utils/data-element-constraints.ts#L75) would need to be modified to include a `boolean` property called `includeMissingData`.

My thinking is that this would be sufficient to update the data request to include the `showMissingness` property, and to inform the plot that there is missing data.

---

A related question: I think we decided that missing data in the plot would use a gray color. Do you have thoughts on how to handle this? Two possibilities that come to mind are:
1. Declare the color for that single data series in the `data` props passed to the plot.
2. Have a dedicated `missingData` prop for each plot.

I'm leaning towards (2) since it will make it easier to be consistent about how we represent missing data in various plots. Thoughts?

src/lib/core/components/visualizations/InputVariables.tsx
bobular: I've got as far as I can.
I don't know if we need to get into Generics, but I'm finding it hard to type the props correctly.  The branch below compiles, but fails at runtime because of the `: any` typing and the lack of "extra props" passed to the "special" component.

```
interface InputSpec {
  name: string;
  label: string;
  component?: (props: any) => JSX.Element; // ? extends VariableTreeDropdownProps
}
```


https://github.com/VEuPathDB/web-eda/tree/missing-data-button-test1

bobular: Update - might abort the wrapper approach!  (See Dave's comments below.)

bobular:
Re: colors. I was thinking along the lines of 2. also.

bobular:
Here's a new branch with a new PR: https://github.com/VEuPathDB/web-eda/pull/308

I will close this PR unmerged early next week.  Thank you for the input!

src/lib/core/components/visualizations/InputVariables.tsx
jtlong3rd: No worries - Dave's suggestion of adding a `hasMissingDataToggle`  should do the job. 😄 

For future reference, the `InputComponent` approach that I was suggesting wouldn't have involved generics - the `InputComponent` would always receive the props the default `VariableTreeDropdown` expects, and these props would always be passed in `InputVariables`. If the `InputComponent` needs additional "props," they could be passed implicitly through whatever `useCallback` was used to derive the `InputComponent`.

(You did something similar for your initial wrapper work - there, the only "explicit" prop you were passing to the `Wrapper` was its `children`, and for the missing data wrapper, you passed the additional `MissingDataToggleWrapperProps` implicitly through the `useCallback` which defined that wrapper.)

bobular: Thanks Jamie - it's good to know I was barking up the right tree!

PR #308, opened 2021-08-14
Switch for showing missing data for overlay variable

### LINKED ISSUE #294: Visualizations: Allow users to show missing data for stratification variables

bobular:
Fixes #294 
Requires https://github.com/VEuPathDB/web-components/pull/189 (especially if greyed out switch is required)

Here's a new approach.

I added a new `InputSpec` prop: `role : 'primary' | 'stratification'`

Now `InputVariables` renders the dropdown inputs as usual, but adds an extra CSS class to any stratification input (in future it will be overlay and faceting).  Then it renders a toggle switch with the same class/styling (currently a lovely pink border).

At the moment the toggle is only shown when an overlay variable is selected.  But this could be easily changed to grey out/disable the toggle (like I had in the previous attempt).

I'm assuming that some CSS magic can visually group/lay out the 'stratification' input components appropriately.

image
(from GEMS1)

### DIFF
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -74,6 +74,7 @@ const useStyles = makeStyles(
   {
     inputs: {
       display: 'flex',
+      flexWrap: 'wrap',
       marginBottom: '0.5em',
     },
     input: {
@@ -93,9 +94,16 @@ const useStyles = makeStyles(
       fontSize: '1.35em',
       fontWeight: 500,
     },
-    primary: {},
+    primary: {
+      order: 0,
+    },
+    showMissing: {
+      order: 1,
+      flexBasis: '100%',
+      marginTop: '1em',
+    },
     stratification: {
-      border: '1px solid pink',
+      order: 2,
     },
   },
   {
@@ -222,7 +230,8 @@ export function InputVariables(props: Props) {
         ))}
         {onShowMissingnessChange &&
           inputs.filter((input) => input.role === 'stratification').length && (
-            <div className={[classes.input, classes.stratification].join(' ')}>
+            <div className={[classes.input, classes.showMissing].join(' ')}>
+              <h4>Stratification (optional)</h4>
               <Switch
                 label={`Include ${
                   outputEntity
@@ -232,6 +241,10 @@ export function InputVariables(props: Props) {
                 state={showMissingness}
                 onStateChange={onShowMissingnessChange}
                 disabled={!enableShowMissingnessToggle}
+                labelPosition="after"
+                containerStyles={{
+                  marginLeft: '20px',
+                }}
               />
             </div>
           )}
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -13,10 +13,13 @@ import {
   mapStructure,
   preorder,
 } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
+import Switch from '@veupathdb/components/lib/components/widgets/Switch';
+import { makeEntityDisplayName } from '../../utils/study-metadata';
 
 interface InputSpec {
   name: string;
   label: string;
+  role?: 'primary' | 'stratification'; // TO DO: make mandatory when all viz are updated??
 }
 
 export interface Props {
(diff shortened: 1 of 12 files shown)


jtlong3rd (commented):
This new approach looks good to me!

dmfalke (commented):
This approach looks good to me! I think the UI needs some work. I recall some mock-ups that move stratification inputs to a different section. That might be a good starting point.

bobular:
Looks good to review now.  I will have a quick look at the Mosaic viz's and see what we should do there.

image

bobular:
Just noticed that the "x,xxx Participants" text is showing the same number with and without missing data.  Need to sort this out!

image

dmfalke (commented):
I just started testing this, and the first thing I noticed is that we include a legend entry for "No data", even if there is no missing data (e.g., scatter plot w/ x = Enrollment date, y = Hospitalization duration, overlay = Country). In this case, I think we should not have an entry in the legend. Of course, this should be raised to UX.

I am also not crazy about the styling. I recall seeing a mock-up that moved stratification vars into their own line (w/ a **Stratification variables** heading). I think we should consider doing this and putting the toggle in that section. If we do this, then I don't think we need the pink boxes.

dmfalke:
[Here is the mock-up](https://docs.google.com/presentation/d/14WwD95etPuj6MS0wHQZ9_2X8CbEQrdhw/edit#slide=id.ge5d990b9bf_0_158) I was referring to.

src/lib/core/components/visualizations/InputVariables.tsx
```diff
@@ -13,10 +13,13 @@ import {
   mapStructure,
   preorder,
 } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
+import Switch from '@veupathdb/components/lib/components/widgets/Switch';
+import { makeEntityDisplayName } from '../../utils/study-metadata';
 
 interface InputSpec {
   name: string;
   label: string;
+  role?: 'primary' | 'stratification'; // TO DO: make mandatory when all viz are updated??
```
dmfalke, line 22: Can this be required now?

dmfalke (approved):
The code looks good to me. I don't have a problem with you merging this. As I mentioned above, I do think we need some UI/UX feedback. That can happen before or after this is merged - your call.

jtlong3rd (approved):

src/lib/core/components/visualizations/InputVariables.tsx
bobular: Ooh yes, good catch.  I just had to update MosaicViz

bobular:
> I just started testing this, and the first thing I noticed is that we include a legend entry for "No data", even if there is no missing data (e.g., scatter plot w/ x = Enrollment date, y = Hospitalization duration, overlay = Country). In this case, I think we should not have an entry in the legend. Of course, this should be raised to UX.
> 
> I am also not crazy about the styling. I recall seeing a mock-up that moved stratification vars into their own line (w/ a **Stratification variables** heading). I think we should consider doing this and putting the toggle in that section. If we do this, then I don't think we need the pink boxes.

You don't like the thin pink outline @dmfalke ??? That was always just a placeholder. Supposed to be too ugly to keep for too long.  Thanks for being polite though!

Now then, when there is actually no missing data - yes, I think that should disable the new switch and it shouldn't show the empty category in the legend - it's actually misleading because people will look for the grey points/bars.  I had thought of this too.  I'll implement something in the morning.

dmfalke:
> You don't like the thin pink outline @dmfalke ??? That was always just a placeholder. Supposed to be too ugly to keep for too long. Thanks for being polite though!

:upside_down_face:

moontrip:
I have tested scatter and box plot viz for several cases and they seem to work well 👍

bobular:
Thanks DK.

I'll see if I can sort out the layout/styling in the morning, then merge.

bobular:
Here's the new styling - not perfect (text doesn't line up properly) but OK for now.

(The facet variables are just a temporary mock-up for this screenshot)

image

PR #312, opened 2021-08-16
Allow user to browse variables without creating an anlysis resource.

### LINKED ISSUE #283: Only create a new analysis on backend when user interacts with filters, viz, etc
We are currently creating a new analysis on the backend as soon as the user clicks on the **New analysis** link. A more ideal behavior would to defer creating the analysis on the backend until the user makes a change that would normally cause an update to be made to the backend (e.g., modify subset, add an analysis, star a variable, etc).

If a user is simply looking at variables without creating filters, a new analysis will not be created. This will prevent lots of "empty" analyses from being created.

dmfalke:
When the user clicks on "New Analysis", we will no longer eagerly create an analysis on the backend. Instead, we will wait until the user makes a change to the analysis to create the resource on the backend. When this happens, the url will be replaced with an analysis id.

If the user changes binning, sorting, or other `variableUISettings`, a new analysis will not be created, but those settings will be carried over if a new analysis is created.

closes #283

### DIFF
--- a/src/lib/core/components/EDAWorkspaceContainer.tsx
+++ b/src/lib/core/components/EDAWorkspaceContainer.tsx
@@ -15,7 +15,7 @@ import { workspaceTheme } from './workspaceTheme';
 const theme = createMuiTheme(workspaceTheme);
 export interface Props {
   studyId: string;
-  analysisId: string;
+  analysisId?: string;
   children: React.ReactChild | React.ReactChild[];
   className?: string;
   analysisClient: AnalysisClient;
--- a/src/lib/core/components/EDAWorkspaceContainer.tsx
+++ b/src/lib/core/components/EDAWorkspaceContainer.tsx
@@ -15,7 +15,6 @@ import { workspaceTheme } from './workspaceTheme';
 const theme = createMuiTheme(workspaceTheme);
 export interface Props {
   studyId: string;
-  analysisId?: string;
   children: React.ReactChild | React.ReactChild[];
   className?: string;
   analysisClient: AnalysisClient;
--- a/src/lib/core/__test__/hooks/useAnalysis.test.tsx
+++ b/src/lib/core/__test__/hooks/useAnalysis.test.tsx
@@ -129,7 +129,6 @@ describe('useAnalysis', () => {
   it('should update store on copy', async () => {
     const { result, waitFor } = render();
     await waitFor(() => result.current.status === Status.Loaded);
-    if (result.current.copyAnalysis == null) return;
     const res = await result.current.copyAnalysis();
     const analyses = await analysisClient.getAnalyses();
     const newAnalysis = analyses.find((analysis) => analysis.id === res.id);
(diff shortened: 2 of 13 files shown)


dmfalke:
One glitch that I'm aware of is, when a NewAnalysis becomes a SavedAnalysis, the analysis view blanks out and any data is reloaded. This is happening because when the route transitions from new to saved, the SavedAnalysis component loads the analysis resource, and while it is loading, renders a blank/loading state. This causes all child components to unmount.

A possible solution is to employ an Analysis cache, and to preload the cache with the new analysis before transitioning. This would likely require a change to how an analysis is loaded in the `useAnalysis` hook, and to expose a function to preload an analysis. I think this would be a small amount of work, which I will attempt shortly.

dmfalke:
> A possible solution is to employ an Analysis cache, and to preload the cache with the new analysis before transitioning. This would likely require a change to how an analysis is loaded in the `useAnalysis` hook, and to expose a function to preload an analysis. I think this would be a small amount of work, which I will attempt shortly.

This does help, but the counts for both the entity diagram and the active filter are reset/reloaded. I'm not sure why yet, and I'm not sure it will be easy to address. I will keep investigating, but would appreciate a review and/or testing from others, if possible. Thanks!

bobular:
I'm testing now.  It would be good to avoid that prominently blank reload if possible.

I seem to have found a glitch where every new analysis is given a number (not 'new') - I'll try to figure out how to replicate this!

bobular:
So, the "Start a new analysis" button needs the new behaviour.

What about the "copy analysis" button?  Aha - I see that's only visible when the analysis has been saved, and it makes sense to save the copy immediately.

Regarding the blank screen/reload.  Maybe we just embrace this?  Show a big "Saving analysis" thingy?

src/lib/core/components/EDAWorkspaceContainer.tsx
```diff
@@ -15,7 +15,7 @@ import { workspaceTheme } from './workspaceTheme';
 const theme = createMuiTheme(workspaceTheme);
 export interface Props {
   studyId: string;
-  analysisId: string;
+  analysisId?: string;
```
bobular, line 18: I don't see `analysisId` used in `EDAWorkspaceContainer` nor do I see `props` passed on to another component, so can this be removed completely?

dmfalke:
Oh, I forgot to push my changes addressing the blank screen issue. One moment.

bobular (approved):
Looks good!
I get the general idea how it all works.

src/lib/core/components/EDAWorkspaceContainer.tsx
dmfalke: Yes! I will remove

bobular:
> Oh, I forgot to push my changes addressing the blank screen issue. One moment.

Oh - just tested this and it seems a bit snappier.

dmfalke:
> So, the "Start a new analysis" button needs the new behaviour.

I just removed this button, FYI.

dmfalke:
Some large branch of the tree of React Components is being unmounted, when switching from new to saved. This is the root issue. I will spend a little bit of time trying to figure out the cause.

dmfalke:
Oh, duh, it's because we are using two different components: `NewAnalysisPage` and `SavedAnalysis`. I did this so that I could simplify the hooks code. I will look into combining them.

bobular:
All very interesting!

dmfalke:
Before I spend too much effort on this, I think we should ask about the UX of what should happen when a new analysis becomes a saved analysis. As you alluded, it might be a better UX to have an overlay indicating that the analysis is being saved/created. I plan to merge this soon and will raise this question once it is deployed.

jtlong3rd (approved):

PR #315, opened 2021-08-18
Subsetting histogram: selected range bounds enforcement is now handled by the client, fix single-day subsets, tweak button widths

bobular:
Depends on https://github.com/VEuPathDB/web-components/pull/190

The components PR simplifies some of the internals of `Histogram` by removing its handling of `selectedRangeBounds`

Here we reinstate the bounds check on the EDA side - see `enforceBounds`.

There were some very minor bugs in single-day resolution range subset selections - these have now all been resolved.

Edit to add: Actually one of the "minor date-related bugs" was not minor at all.  If you selected a single bar, at any resolution (month, year, day etc), in a date-variable histogram (subsetting), you would see that the size of the subset is not the same as the number of entities represented by the bar.  This is because the filter created is including the end day.  So a bin that is labelled "01 Jan 2018 to 01 Feb 2018" needs to create a filter "01 Jan 2018 to 31 Jan 2018".  This has now been fixed, but at the Histogram component level, which means at some point we should document exactly what the bin and filter specifications are.  I will add a technical debt ticket.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.6.0",
+    "@veupathdb/components": "^0.7.0",
     "@veupathdb/wdk-client": "^0.2.0",
     "@veupathdb/web-common": "^0.1.5",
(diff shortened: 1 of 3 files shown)


bobular:
Ready for review.  A bug I thought I had, was actually already in `main` - about to report it

dmfalke (approved):

moontrip (approved):

PR #320, opened 2021-08-19
dependentAxisLabel always Count in subset histogram

asizemore:
_Addresses issue #288_

Set the y axis label of the histogram in the subsetting tab to always read "Count". See attached image.

Tested locally with GEMS and PRISM2.

Screen Shot 2021-08-19 at 2 56 26 PM

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -532,7 +532,7 @@ function HistogramPlotWithControls({
         displayLibraryControls={displayLibraryControls}
         onSelectedRangeChange={handleSelectedRangeChange}
         barLayout={barLayout}
-        dependentAxisLabel={`Count of ${entityName}`}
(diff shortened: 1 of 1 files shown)


dmfalke (approved):
Looks great, @asizemore!

dmfalke:
Hi @bobular, would you mind quickly reviewing this? Thanks!

bobular:
On it!

bobular (changes requested):
Looks good!

Although my emacs IDE does not highlight this (which is strange, because I think it should) - `entityName` is now an unused "variable" in `HistogramPlotWithControls` and there is now no need to pass `entityName` through into this component.  The `entityName` prop can be removed from this component.

asizemore:
Hi @bobular ! Thank you for the feedback! I found three instances of `entityName` and removed them all so that the component does not use it anymore. Let me know if there's anything else you see that I can fix!

bobular (approved):

PR #321, opened 2021-08-20
keep hide toggle button state at variable selector

moontrip:
During this work, I found a small bug on the hiding (toggle-on) that hides variables including the Featured Variable. The issue was that if the Featured Variable is non-existent, i.e., `allowedFeaturedFields.length === 0` (at `VariableList` component), then erroneous zero (0) character is shown (see attached screenshot). It was solved by simply adding larger than 0, i.e., `allowedFeaturedFields.length > 0`.

Anyway, I have tested several case and this implementation seems to work fine - only applied to scatter plot viz for now. However, I noticed that this had a side effect: whenever toggle the show/hide button, then plot is redrawn because I used vizualization.config to store the status of the buttons (on/off). It would be great if you (@dmfalke, @jtlong3rd, or anyone) can suggest a way to avoid this.

variablelist-smallbug1

### DIFF
--- a/src/lib/core/components/VariableList.tsx
+++ b/src/lib/core/components/VariableList.tsx
@@ -85,6 +85,8 @@ interface VariableListProps {
   hideDisabledFields: boolean;
   setHideDisabledFields: (hide: boolean) => void;
   featuredFields: VariableField[];
+  //DKDK onHideToggleChange
+  onHideVariableToggleChange?: (value: boolean) => void;
 }
 
 interface getNodeSearchStringType {
@@ -115,6 +117,8 @@ export default function VariableList(props: VariableListProps) {
     toggleStarredVariable,
     hideDisabledFields,
     setHideDisabledFields,
+    //DKDK onHideVariableToggleChange
+    onHideVariableToggleChange,
   } = props;
   const [searchTerm, setSearchTerm] = useState<string>('');
   const getPathToField = useCallback(
@@ -379,6 +383,9 @@ export default function VariableList(props: VariableListProps) {
               type="button"
               onClick={() => {
                 setHideDisabledFields(!hideDisabledFields);
+                //DKDK onHideVariableToggleChange
+                if (onHideVariableToggleChange)
+                  onHideVariableToggleChange(!hideDisabledFields);
               }}
             >
               <Toggle on={hideDisabledFields} /> Only show compatible variables
@@ -386,7 +393,8 @@ export default function VariableList(props: VariableListProps) {
           </HtmlTooltip>
         </div>
       )}
-      {allowedFeaturedFields.length && (
+      {/* DKDK allowedFeaturedFields.length should have > 0 condition: otherwise, it prints 0 when .length = 0 */}
+      {allowedFeaturedFields.length > 0 && (
         <div className="FeaturedVariables">
           <details
             open={Options.featuredVariablesOpen}
--- a/src/lib/core/components/VariableTree.tsx
+++ b/src/lib/core/components/VariableTree.tsx
@@ -23,6 +23,9 @@ export interface Props {
   onChange: (variable?: VariableDescriptor) => void;
   hideDisabledFields?: boolean;
   setHideDisabledFields?: (hide: boolean) => void;
+  //DKDK onHideVariableToggleChange
+  hideVariableToggleStatus?: boolean;
+  onHideVariableToggleChange?: (value: boolean) => void;
 }
 export function VariableTree(props: Props) {
   const {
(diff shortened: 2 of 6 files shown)


moontrip:
perhaps I make a new object key under `visualization` (beyond `visualization.configuration`) and its update function, like `visualization.buttonConfig` and `updateVizButtonConfig` function, and then place button states in there. I will test this approach.

moontrip:
tried to make new object property and updated function, `visualization.buttonConfig` and `updateVizButtonConfig` function, which makes the side effect be resolved.

dmfalke:
@moontrip I think you may be making this more complicated than it needs to be. Perhaps the requirements we not clearly stated.

The scope of the state of the toggle should be application-wide, rather than per-visualization. In other words, the state of the toggle should be the same everywhere. E.g., if it's toggled on in one place, then it should be on in all places.

I think the best place to store this state is in [AnalysisPanel.tsx](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/workspace/AnalysisPanel.tsx). Eventually, this should be persisted to the UserStateService, but that has not been implemented yet, so we can defer on that. For now, I think it's okay if it is just component state.

Let me know if you have questions.

moontrip:
> @moontrip I think you may be making this more complicated than it needs to be. Perhaps the requirements we not clearly stated.
> 
> The scope of the state of the toggle should be application-wide, rather than per-visualization. In other words, the state of the toggle should be the same everywhere. E.g., if it's toggled on in one place, then it should be on in all places.
> 
> I think the best place to store this state is in [AnalysisPanel.tsx](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/workspace/AnalysisPanel.tsx). Eventually, this should be persisted to the UserStateService, but that has not been implemented yet, so we can defer on that. For now, I think it's okay if it is just component state.
> 
> Let me know if you have questions.

@dmfalke thank you for your detailed explanation! 👍 Yeah I might have overthought of the thing 😓 Perhaps we may do a show-hide-variable button toggle to affect other buttons at visualization app level (not per-plot-viz) - I mean, all true or false regardless of X/Y/Main/OverlayVariable: given the situation that each viz plot have different variable selector, this is reasonable I suppose (unless we use per-viz-plot level control).

Please correct me if I am wrong in the following. Off the top of the head, perhaps we use `useContext` hook approach at `AnalysisPanel` (wrapping with `<SomeContext.Provider>`). Certainly I need to make a Context/Context.Provider component beforehand: the Context will have button state variable (boolean) and its handler function to change the button state. Then, use the Context at relevant places (`VariableTree `and `VariableList`), similar to what I have done in the previous approach (changing initial value of useState of toggle button).

Since I am doing for other works today (Friday as usual), I will probably test this approach later.

dmfalke:
I'm not sure we really need to use context for this. It might be simpler to just pass the state and callback as props. I'm finding that using context adds a bit of indirection and would like to try to avoid it. Other than that, I think this sounds good.

moontrip:
> I'm not sure we really need to use context for this. It might be simpler to just pass the state and callback as props. I'm finding that using context adds a bit of indirection and would like to try to avoid it. Other than that, I think this sounds good.

@dmfalke Thank you for sharing your thought. Yes, from a logic perspective, passing state & callback may be simpler, however.  it would be quite complicated for implementation in some sense due to the following two-fold: a) if I have to pass state & callback props from `AnalysisPanel` to `VariableTreeDropdown` (residing in `VariableTree` component) (defining toggle button state) and `VariableList` (button onClick event), it means they pass through lots of components: (I might miss something here but) AnalysisPanel -> ComputationRoute -> PassThroughComputation -> VisualizationsContainer -> FullScreenVisualization -> Plot viz component (histogram, scatter vizs etc.) -> InputVariables ->  -> VariableTree -> VariableList; b) considering the route at a), if I remember correctly, there is no specific way to pass state & function props directly to VariableTreeDropdown without stopping by Plot viz component (histogram, scatter vizs etc.). This means that the state & callback props should be defined at all plot vizs. 

I glanced at your new comments at hooks cleanup and some concerns regarding useContext, so probably I have no choice to use  the directly-passing-props approach 😄: at some components in the long route probably I don't need to change much but either way I need to add types for those state & callback props haha.

dmfalke:
@moontrip I've been ruminating on this a bit more, and I think the potential performance issues I cited with React.Context are unfounded. Why don't you go ahead with your original plan, and we can revisit if need be.

moontrip:
> @moontrip I've been ruminating on this a bit more, and I think the potential performance issues I cited with React.Context are unfounded. Why don't you go ahead with your original plan, and we can revisit if need be.

@dmfalke Thank you for checking it! 👍 Yeah I will try my idea 🤞

moontrip:
With the new approach using useContext, #329, this PR is closed for now.

PR #329, opened 2021-08-23
using useContext for remembering show/hide variable toggle at variable selector

moontrip:
After discussing with @dmfalke at the previous PR, #321, we have decided to simplify the behavior. In this PR, useContext is used to remember the state of show/hide variable toggle at variable selector. Now the behavior is: a) at a plot viz, show/hide toggle is reserved for each variable (X, Y, Overlay variables etc.); b) but if leaving a plot viz, only the last state of show/hide toggle is reserved via useState and it would be used as a new state for a new plot viz: i.e., all true or all false regardless of variable selectors (X, Y, Overlay variables etc.).

### DIFF
--- a/src/lib/core/components/VariableTree.tsx
+++ b/src/lib/core/components/VariableTree.tsx
@@ -11,8 +11,6 @@ import { edaVariableToWdkField } from '../utils/wdk-filter-param-adapter';
 import VariableList from './VariableList';
 import './VariableTree.scss';
 import { useStudyEntities } from '../hooks/study';
-// import ShowHideVariableContext
-import { ShowHideVariableContext } from '../utils/show-hide-variable-context';
 
 export interface Props {
   rootEntity: StudyEntity;
@@ -23,8 +21,6 @@ export interface Props {
   disabledVariables?: VariableDescriptor[];
   /** term string is of format "entityId/variableId"  e.g. "PCO_0000024/EUPATH_0000714" */
   onChange: (variable?: VariableDescriptor) => void;
-  hideDisabledFields?: boolean;
-  setHideDisabledFields?: (hide: boolean) => void;
 }
 export function VariableTree(props: Props) {
   const {
@@ -35,8 +31,6 @@ export function VariableTree(props: Props) {
     entityId,
     variableId,
     onChange,
-    hideDisabledFields = false,
-    setHideDisabledFields = () => {},
   } = props;
   const entities = useStudyEntities(rootEntity);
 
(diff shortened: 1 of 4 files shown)


src/lib/core/utils/show-hide-variable-context.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { createContext, useState } from 'react';
+
+export const ShowHideVariableContext = createContext({
+  toggleShowHideVariable: false,
+  setToggleShowHideVariableHandler: (toggleShowHideVariable: boolean) => {},
```
dmfalke, line 5: I think these could be named a little better. The current names are a bit ambiguous. How about `showOnlyCompatibleVariables`/`setShowOnlyCompatibleVariables`?

src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -175,10 +177,13 @@ export function AnalysisPanel(props: Props) {
...
-        path={`${routeBase}/visualizations`}
-        render={() => <ComputationRoute analysisState={analysisState} />}
-      />
+      {/* use ShowHideVariableContextProvider here for handling the state of show/hide variable toggle */}
+      <ShowHideVariableContextProvider>
+        <Route
+          path={`${routeBase}/visualizations`}
+          render={() => <ComputationRoute analysisState={analysisState} />}
+        />
+      </ShowHideVariableContextProvider>
```
dmfalke, line 186: The provider should be the root node returned by this component. Otherwise the state will be reset every time the user switches between the subset/viz tabs.

dmfalke (changes requested):
Hi @moontrip, this is looking good. I've made some suggestions for changes, if you don't mind taking a look. Thanks!

src/lib/core/components/VariableTree.tsx
```diff
@@ -169,7 +171,16 @@ export function VariableTree(props: Props) {
...
-  const [hideDisabledFields, setHideDisabledFields] = useState(false);
+
+  // use ShowHideVariableContext for setting initial hideDisabledFields boolean
+  const {
+    toggleShowHideVariable,
+    setToggleShowHideVariableHandler,
+  } = useContext(ShowHideVariableContext);
+  const [hideDisabledFields, setHideDisabledFields] = useState(
+    toggleShowHideVariable
+  );
```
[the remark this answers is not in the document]
moontrip: @dmfalke Thank you for your reviews and suggestions. 
As for this comment, setHideDisabledFields (state action) is required to make click event of toggling button so it is necessary.

src/lib/workspace/AnalysisPanel.tsx
moontrip: @dmfalke actually I wrapped whole returning div with this provider at first and then changed it to a local region (viz only). Will change this back 😄

src/lib/core/components/VariableTree.tsx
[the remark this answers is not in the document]
moontrip: @dmfalke Ah... I got your point. So, instead of hideDisabledFields, using the context would work as well. I will check this case - there are several places to be changed then.

moontrip:
@dmfalke updated with your suggestions. By using `showOnlyCompatibleVariables `instead of current `hideDisabledFields`, the **local** state of the toggle button is also changed immediately for all variable selectors: all true or false. I think it is alright, perhaps better than keeping local state of the toggle button per variable selector.

dmfalke:
> @dmfalke updated with your suggestions. By using `showOnlyCompatibleVariables `instead of current `hideDisabledFields`, the **local** state of the toggle button is also changed immediately for all variable selectors: all true or false. I think it is alright, perhaps better than keeping local state of the toggle button per variable selector.

Great! This sounds like what we want 🙂.

jtlong3rd (approved):

dmfalke (approved):

moontrip:
@dmfalke @jtlong3rd Thank you for your reviews and suggestions 👍

PR #346, opened 2021-08-25
truncated axis for histogram filter

moontrip:
This PR relies on the [PR at web-components](https://github.com/VEuPathDB/web-components/pull/195). Detailed description is discussed in there.

Clean-up is not done yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.7.3",
+    "@veupathdb/components": "^0.7.4",
     "@veupathdb/wdk-client": "^0.2.0",
     "@veupathdb/web-common": "^0.1.5",
     "debounce-promise": "^3.1.2",
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -24,7 +24,7 @@ import UnknownCount from '@veupathdb/wdk-client/lib/Components/AttributeFilter/U
 import { getOrElse } from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/function';
 import { number, partial, TypeOf, boolean, type, intersection } from 'io-ts';
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback, useMemo, useEffect } from 'react';
 import { usePromise } from '../../hooks/promise';
 import { AnalysisState } from '../../hooks/analysis';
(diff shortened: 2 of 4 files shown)


bobular (commented):
This is very nice, thanks.  I would like to have a look on Monday at the logic behind the dependent axis truncation flag. (I know you've been concerned about this for a week or so!)  I wonder if it's worth doing a max(bin.counts) to at least detect overflow at the top of the plot.  IIRC Danica and co are not too bothered about the underflow warning.  Or we could change the wording in the warning to "may be truncated" and leave things as they are.

bobular:
Hi @moontrip  - I have pushed a commit that I think leaves this new feature in a functionally/scientifically acceptable state for feedback from the outreach team.

What's new is that the dependent axis lower truncation is only shown if the user-entered lower bound is > 0.  Histograms always show counts, so we never expect the dependent axis to have negative values.  (For this reason, I think `truncationConfig()` isn't going to be multi-purpose and could be moved out of the utils directory.)

The dependent axis warning has softer language than the independent axis too.

We could check the dependent axis more carefully for truncated data in the HistoViz component, but I think we should only do that if advised by ClinEpi folk, as it will be a bit messy making another pass on the data.

bobular:
I wonder, how easy would it be to make sure that the grey bars were a similar thickness on both axes? Might be fiddly?  A Phase 2 thing, I guess.

moontrip:
> Hi @moontrip - I have pushed a commit that I think leaves this new feature in a functionally/scientifically acceptable state for feedback from the outreach team.
> 
> What's new is that the dependent axis lower truncation is only shown if the user-entered lower bound is > 0. Histograms always show counts, so we never expect the dependent axis to have negative values. (For this reason, I think `truncationConfig()` isn't going to be multi-purpose and could be moved out of the utils directory.)
> 
> The dependent axis warning has softer language than the independent axis too.
> 
> We could check the dependent axis more carefully for truncated data in the HistoViz component, but I think we should only do that if advised by ClinEpi folk, as it will be a bit messy making another pass on the data.

@bobular Thank you for your changes 👍 As for the truncationConfig(), it may also be utilized for other plot vizs like passing additional argument, e.g, plot name like 'histogram', I suppose.

moontrip:
> I wonder, how easy would it be to make sure that the grey bars were a similar thickness on both axes? Might be fiddly? A Phase 2 thing, I guess.

@bobular Well I think it has inherent difficulties due to: a) plot width and height are quite different (not square but wide rectangle); b) the range of X is quite different from Y; c) there is log scale as well; d) the width of plot is basically automatic, i.e., depending on browser's size, while plot height (Y) is fixed. Also, different browser's zoom level also affects plot width. Thus, it would be a daunting task to find equal bar widths, I suppose.

jtlong3rd (approved):

PR #347, opened 2021-08-25
Add multiFilter support to front end

### LINKED ISSUE #260: Subsetting: Implement multi-filter support
We should be able to reuse the multifilter component from wdk-client.

dmfalke:
### Summary
This PR adds multiFilter support to the EDA Workspace. MultiFilters are a feature from the currently production ClinEpiDB website.

image

### Notes
* The filter chip has been simplified, as compared to the production site. This will need to be discussed in a UX meeting.
* The backend API could be updated to handle requesting distributions for multiple variables in one go.
* ~~We unnecessarily refresh counts when changes are made to the active multiFilter. This will be addressed shortly.~~

---

~~depends on https://github.com/VEuPathDB/WDKClient/pull/148~~
fixes #260

### DIFF
--- a/src/lib/core/components/filter/FilterContainer.tsx
+++ b/src/lib/core/components/filter/FilterContainer.tsx
@@ -3,17 +3,18 @@ import { AnalysisState } from '../../hooks/analysis';
 import {
   StudyEntity,
   StudyMetadata,
-  VariableTreeNode,
   Variable,
+  MultiFilterVariable,
 } from '../../types/study';
 import { isHistogramVariable, isTableVariable } from './guards';
 import { HistogramFilter } from './HistogramFilter';
+import { MultiFilter } from './MultiFilter';
 import { TableFilter } from './TableFilter';
 import UnknownFilter from './UnknownFilter';
 
 interface Props {
   studyMetadata: StudyMetadata;
-  variable: Variable;
+  variable: Variable | MultiFilterVariable;
   entity: StudyEntity;
   analysisState: AnalysisState;
   totalEntityCount: number;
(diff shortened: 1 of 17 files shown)


dmfalke:
I almost forgot about a necessary update to [wdk-client](https://github.com/VEuPathDB/WDKClient). I will update soon.

dmfalke:
Per the UX meeting today:

- [ ] Update counts for sub-variables based on selection. Take operation into account.
- [ ] Add "Update" button for the above.

dmfalke:
Spoke w/ Steve. Will emulate the live site, which is to update sibling counts only if operation is **all**. The case for **any** is more challenging. I will open a new issue for this, with details.

bobular (approved):
Looks good and works well for me.

Was it intentional to not have the "22,566 (>99%) of 22,567 Household have data for this variable"  above each variable like in the live clinepi site?

src/lib/core/components/filter/FilterContainer.tsx
```diff
@@ -25,17 +26,20 @@ export function FilterContainer(props: Props) {
     <HistogramFilter {...props} />
   ) : narrowProps(isTableVariable, props) ? (
     <TableFilter {...props} />
+  ) : narrowProps(MultiFilterVariable.is, props) ? (
```
bobular, line 29: the `IOTSType.is` seems cool :+1:

bobular (approved):
> Looks good and works well for me.
> 
> Was it intentional to not have the "22,566 (>99%) of 22,567 Household have data for this variable" above each variable like in the live clinepi site?

Ah from MultiFieldFilter.jsx - it looks like if there is 100% data, this is skipped.

```
      : unknownCount > 0 && (
        <div style={{ fontWeight: 300 }}>
          <b>{toPercentage(row.summary.internalsCount,this.props.dataCount).toLocaleString()}% of {this.props.dataCount.toLocaleString()}</b> {this.props.displayName} have data  
        </div>
```

src/lib/core/components/filter/FilterContainer.tsx
dmfalke: :smile:

dmfalke:
This should be good to go now. Once I merge the latest from `main`, I will likely merge this PR.

PR #354, opened 2021-08-31
Viz thumbnails

### LINKED ISSUE #272: Visualizations: Use thumbnail image to speed up grid view
We are currently fetching data and dynamically generating miniature plots in the grid view. This performs badly with large datasets (umsp) and produces unnecessary compute work.

The proposed alternative is to grab a image of the primary plot of a viz, and to store it as a base64 string as a part of the [`Visualization` type](https://github.com/VEuPathDB/web-eda/blob/a4cb7512e8a1099f475e124dd0f65a0367f6c21c/src/lib/core/types/visualization.ts#L20-L30).

When rendering the grid view, if the property is not populated, we can use a placeholder (possibly the same svg used in the picker view, with overlay text "Preview not available").

The base64 string should be updated anytime the viz config is changed in the fullscreen view. Also, note that this can only be populated once the fullscreen view has been loaded and rendered.

dmfalke:
This PR adds support for using an image in the visualizations grid view. It takes advantage of changes introduced in https://github.com/VEuPathDB/web-components/pull/201.

### Overview

- Added browserify loader to webpack config
- Updated AnalysisState to be more immutable-friendly
- Fixed some minor type issues
- Added thumbnail property to Visualization
- Removed `gridComponent` from `VisualizationType` and replaced with `thumbnail` in overview

### Remaining TODOs
- [x] Delete thumbnails if subset changes
- [x] Document webpack config needs
- [x] Determine default image

---
depends on https://github.com/VEuPathDB/web-components/pull/201
fixes #272


moontrip:
@dmfalke For your information, I have tested plot vizs and they mostly worked great! 👍 I found a bug at Scatter plot viz: it worked fine for the default Raw plot, but either Smoothed mean or Best fit line threw an error (Error compiling fragment shader). However, your recent commit resolved this bug so no error concerning thumbnails found yet 😄

dmfalke:
This s ready for review. There are a couple of remaining TODOs, which I will address before merging.

moontrip (approved):

jtlong3rd (approved):

dmfalke:
For the default image (when the client hasn't generated one yet), I was thinking we could use a grayscale version of the selector icon with the text **Preview not available** overlaid. Thoughts?

jtlong3rd:
> For the default image (when the client hasn't generated one yet), I was thinking we could use a grayscale version of the selector icon with the text **Preview not available** overlaid. Thoughts?
> 
> ETA: This will happen when the subset changes.

Makes sense to me!

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -382,12 +331,20 @@ function BarplotWithControls({
...
   onValueSpecChange,
+  updateThumbnail,
   ...barPlotProps
 }: BarplotWithControlsProps) {
+  const ref = useRef<PlotRef>(null);
+  useEffect(() => {
+    ref.current
+      ?.toImage({ format: 'svg', ...plotDimensions })
+      .then((src) => updateThumbnail(src));
+  }, [data, updateThumbnail]);
```
bobular, line 342: Non-back-end controls need to be passed too
```suggestion
  }, [data, updateThumbnail, dependentAxisLogScale]);
```

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -406,12 +360,21 @@ function HistogramPlotWithControls({
...
   const displayLibraryControls = false;
   const opacity = 100;
 
+  const ref = useRef<PlotRef>(null);
+
+  useEffect(() => {
+    ref.current
+      ?.toImage({ format: 'svg', ...plotDimensions })
+      .then(updateThumbnail);
+  }, [updateThumbnail, data]);
```
bobular, line 376:
```suggestion
  }, [updateThumbnail, data, histogramProps.dependentAxisLogScale]);
```

bobular (changes requested):
It's great and looks much better than the mini versions of the plots we had before.  I just noticed some dependencies to update the thumbnail when front-end-only controls are used.

However, it seems not backwards compatible with old saved analyses - the error below occurs when maximising an old visualization.  This isn't really a problem I guess?  It will be a pain to diagnose (check out `main` - make viz's, check out this branch, etc). 

```
Unhandled Rejection (Error): No DOM element with id 'plotly_plot_div_1' exists on the page.
toImage/<
src/plots/PlotlyPlot.tsx:234

  231 |   toImage: async (imageOpts: ToImgopts) => {
  232 |     await sharedImport.promise;
  233 |     await new Promise((resolve) => setTimeout(resolve, 0));
> 234 |     return toImage(plotId, imageOpts);
      | ^ 
  235 |   },
  236 | }));
  237 | 
```

dmfalke:
@bobular I've applied your suggestions. Thanks!

bobular (approved):

Issue #340, opened 2021-08-18
General: add units where needed on subset & visualize tabs


danicahelb:
Units that are associated with each variable (if any) are annotated in the .owl files under `unitLabel`

If available, the unit should be added in parenthesis to the following places:

**Browse/subset:**

1. Variable header
2. Histogram axis
3. Histogram Subset selector (ex: SUBSET ON WEIGHT (KG))

**Visualizations:**

1. All plot axes
2. All statistics tables

NOTE: in July 26, 2021 EDA WG meeting, we decided to not show units in variable picker trees since once people can change units, could get confusing there

SheenaTomko:
Units appear correctly in every place Danica mentioned

PR #356, opened 2021-08-31
adding units

moontrip:
This PR addresses #340, General: add units where needed on subset & visualize tabs.

A couple of screenshots are attached:

weight1

scatter-unit1

histogram-unit1

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -36,6 +36,8 @@ import { HistogramVariable } from './types';
 import { fullISODateRange, padISODateTime } from '../../utils/date-conversion';
 import { getDistribution } from './util';
 import { DistributionResponse } from '../../api/subsetting-api';
+// import axis label unit util
+import { axisLabelWithUnit } from '../../utils/axis-label-unit';
 
 type Props = {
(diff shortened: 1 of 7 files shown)


moontrip:
@jtlong3rd Thank you for your review 👍 It is quite fast, so I wonder you keep waiting for a PR to review 😄

bobular (changes requested):
I'll message Danielle in the Slack channel to double check my comments about unnecessary units code are valid

bobular:
Discussion here
https://epvb.slack.com/archives/C012ZK4Q5CZ/p1630533616005900

Maybe we should keep it.  There could be a variable with limited values: 1 kg, 2 kg, 3 kg, 4 kg.

moontrip:
> Discussion here
> https://epvb.slack.com/archives/C012ZK4Q5CZ/p1630533616005900
> 
> Maybe we should keep it. There could be a variable with limited values: 1 kg, 2 kg, 3 kg, 4 kg.

@bobular thank you for your review! 👍 Since I can't check all possible whether a variable would have unit or not, I just wrapped up all variables' name with a function to add units if exists: if not, then same to variable name. This is a safe/conservative approach 😄

dmfalke (approved):

PR #377, opened 2021-09-03
pass MembershipField  bar colors as props

### LINKED ISSUE #301: General: grey and red colors in browse/subset histogram do not match colors in entity diagram
image

The grey & red colors in the browse/subset histogram should match the grey & red used in the entity diagram.

Before adjusting, wait for a final decision on what colors we will use in the birds eye view bar summary, as this will impact the red color that we should use for the browse/subset & entity diagram

asizemore:
*Depends on WDKClient [PR #151](https://github.com/VEuPathDB/WDKClient/pull/151)*

Resolves #301 

Goal: mute the colors in the histogram and table in the subsetting tab so that they match the entity diagram.

Implementation notes: Now both the table and histogram read the `red`, `gray` colors defined in `colors.ts`. For the table, we needed to pass these two colors to the `MembershipField` component. See WDKClient PR for details on updates to this component.

### DIFF
--- a/src/lib/core/components/filter/TableFilter.tsx
+++ b/src/lib/core/components/filter/TableFilter.tsx
@@ -14,6 +14,7 @@ import { fromEdaFilter } from '../../utils/wdk-filter-param-adapter';
 import { TableVariable } from './types';
 import { getDistribution } from './util';
 import { DistributionResponse } from '../../api/subsetting-api';
+import { gray, red } from './colors';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


bobular (approved):
Looks good and works fine.

I'll leave the WDKClient review to Dave because I'm not familiar with non-Typescript React - maybe the new props need declaring somewhere?

dmfalke (approved):

Issue #338, opened 2021-08-18
Visualisations: showMissingness toggle should be disabled if there's no missing data to show


bobular:
This deserves a new ticket/task as the implementation is non-trivial.

Depends on https://github.com/VEuPathDB/web-eda/pull/308

We want to avoid situations like this (GEMS1):
![image](https://user-images.githubusercontent.com/308639/129883478-c55e9fbf-fadc-498f-9847-0cde220969ed.png)
where clicking on "show no data" makes zero difference to the plot.

The implementation challenge is to avoid performing all the `count` requests and calculations in `VariableCoverageTable` twice.

Or to put it another way: we need numbers that are calculated in `VariableCoverageTable` at a higher level in the component tree, but don't want to call it twice.

aaronwlsong:
Tested several variables in different study, looks good now, closing ticket.

PR #378, opened 2021-09-06
Move/improve stratification variables in visualizations variable picker

### LINKED ISSUE #295: Visualizations: Change location of stratification variables
Reviewed [mockups](https://docs.google.com/presentation/d/14WwD95etPuj6MS0wHQZ9_2X8CbEQrdhw/edit#slide=id.ge5d990b9bf_0_158) in EDA UX meeting on Aug 9th. Want to move or otherwise emphasize how stratification variables differ from X- and Y-axis variables. Mockups suggest moving stratification vars under the X- and Y-axis vars. Dan and others suggested an alternative of using subtle shading or other mechanisms to distinguish them to save vertical space. Dev team to come up with an idea and implement it

bobular:
Fixes #295 

I'm not 100% happy about the layout and responsiveness of this.  It should definitely be revisited (by someone else) after the ALL-CAPS is removed.

The logic and basic structure is all good though IMO, so please review as it is kind-of blocking #338 (unless I work from this branch)

image

### DIFF
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -104,6 +104,7 @@ const useStyles = makeStyles(
     },
     primary: {},
     stratification: {},
+    showMissingness: {},
   },
   {
     name: 'InputVariables',
@@ -270,7 +271,7 @@ export function InputVariables(props: Props) {
                 </div>
               ))}
             {onShowMissingnessChange && (
-              <div className={classes.fullRow}>
+              <div className={classes.showMissingness}>
                 <Switch
                   label={`Include ${
                     outputEntity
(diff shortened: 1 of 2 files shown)


moontrip:
@bobular I was a bit confused initially but I realized Danielle suggested other layout. Anyway, the location of the toggle button for stratification changes per plot viz: e.g., see histogram viz below. I think that It is/would be not that easy to locate the toggle in this manner, especially when we will have facet selector(s) in the future. 

histogramviz1

jtlong3rd (approved):

dmfalke:
I agree with @moontrip that the current placement of the toggle is not ideal. I was expecting something that looks more like this:
image

I'm happy to make this work, or to provide tips on how to make it work. @bobular let me know what you prefer.

bobular:
I've done what you both suggested - though haven't done anything about the precise vertical positioning.  Thanks!

Issue #325, opened 2021-08-21
Visualizations: Update text under visualization picker icons


danicahelb:
1. Update label for "RxC Contingency Table" to "RxC Contingency"
2. Update label for "2x2 Contingency Table" to "2x2 Contingency"
3. Make text below all icons larger

danicahelb:
https://epvb.slack.com/archives/C0132PM1MT5/p1629987169001100

danicahelb:
Can you also update the name of the line plot to "Time series"
Thank you!

asizemore:
Viz names updated and font size increased (from 13px to 16px). @danicahelb is that about the right size for the text below the icons?
<img width="1768" alt="Screen Shot 2021-09-03 at 3 09 27 PM" src="https://user-images.githubusercontent.com/11710234/132056041-a676f0e5-1499-4160-8caa-e739c2f61b7e.png">

Note these viz names are used elsewhere, for example as subtitles when viewing the full screen version or as thumbnails. For these other places we may want to use a one-line version, for example "Mosaic plot, RxC table". See screenshots below
<img width="647" alt="Screen Shot 2021-09-03 at 2 39 13 PM" src="https://user-images.githubusercontent.com/11710234/132056325-2b00df8c-adea-46cb-9a1d-b34f2d849904.png">
<img width="381" alt="Screen Shot 2021-09-03 at 2 39 22 PM" src="https://user-images.githubusercontent.com/11710234/132056332-35457a88-cc4d-432f-b13e-94247c7f00c9.png">

danicahelb:
I think 16px looks good, lets push it and see what it looks like on the site!

I agree that the full screen and thumbnail versions should be on one line. @asizemore do we need a new ticket for this?

asizemore:
Great will do!

We don't need a new ticket for having these versions in one line, the code I have for this issue will solve both this as well.

asizemore:
Closed with web-eda PR #379 and EdaDataService [PR #81](https://github.com/VEuPathDB/EdaDataService/pull/81)

SAK22:
Text still has "Table" under the icons

asizemore:
Just checked locally with all of the latest code and the names are correct. The name changes had to be made in EdaDataService, which had a build fail last time before the qa site was rebuilt, which I suspect is the reason on qa they are not showing up correctly. Should be fixed next time the qa site is built - I will keep checking in on this!

SAK22:
[...]

Issue #81, opened 2021-05-07
General UI: Button styling needs to be updated


dmfalke:
Styling from `web-common` is taking precedent over `material-ui` styling. This needs to be addressed in `web-common`.

bobular:
Is this also #55 ?

dmfalke:
Yes it is. Thanks for spotting that!

PR #379, opened 2021-09-06
Increase viz name size

asizemore:
With data service [PR #81](https://github.com/VEuPathDB/EdaDataService/pull/81) , addresses issue #325.

This PR makes the following updates:
- In the viz display name contains a comma, split into two lines in the create a new viz view (see screenshot using updated names)
- Increase viz display name from 13pt to 16pt.

Note: The data service PR renames the vizs, this PR affects how the names are displayed. 
Screen Shot 2021-09-06 at 4 49 22 PM

### DIFF
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -110,6 +110,7 @@
       color: #444;
       font-weight: 500;
       text-align: center;
+      font-size: 16px;
     }
   }
   &-FullScreenContainer {
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -110,7 +110,7 @@
       color: #444;
       font-weight: 500;
       text-align: center;
-      font-size: 16px;
+      font-size: 1.2em;
     }
   }
   &-FullScreenContainer {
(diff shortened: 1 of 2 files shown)


d-callan:
looks like that issue requests time series be disabled for phase 1 as well? is that already done?

src/lib/core/components/visualizations/Visualizations.scss
```diff
@@ -110,6 +110,7 @@
       color: #444;
       font-weight: 500;
       text-align: center;
+      font-size: 16px;
```
dmfalke, line 113: Could you change this to use `em` units? They are proportional (whereas `px` is fixed), which makes it easier to scale entire sections of the page, if needed. This is a **big** area of debate. Here is a [good introduction to the topic](https://engageinteractive.co.uk/blog/em-vs-rem-vs-px), if you're interested :slightly_smiling_face: .
```suggestion
      font-size: 1.2em;
```

dmfalke (changes requested):
This looks good. I did request a change (see inline comment). Thanks!

src/lib/core/components/visualizations/Visualizations.scss
asizemore: Done! Oh fascinating. Thank you for the article it was really helpful!

dmfalke (approved):

moontrip (approved):
thank you for changing it 👍

src/lib/core/components/visualizations/Visualizations.scss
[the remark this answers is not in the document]
asizemore: Thank you for the explanation @moontrip! I really appreciate it!

PR #397, opened 2021-09-07
change notification for truncated axis

moontrip:
This issue replies on the [corresponding PR](https://github.com/VEuPathDB/web-components/pull/206) at web-components. Details can be found at the PR.

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -548,7 +548,7 @@ function HistogramPlotWithControls({
       truncationConfigIndependentAxisMax
     ) {
       setTruncatedIndependentAxisWarning(
-        ' Data has been truncated (light gray area) by range selection'
+        'Data has been truncated by range selection, as indicated by the light gray shading'
       );
     }
   }, [truncationConfigIndependentAxisMin, truncationConfigIndependentAxisMax]);
@@ -556,7 +556,7 @@ function HistogramPlotWithControls({
   useEffect(() => {
     if (truncationConfigDependentAxisMin || truncationConfigDependentAxisMax) {
       setTruncatedDependentAxisWarning(
-        ' Data may have been truncated (light gray area) by range selection'
+        'Data may have been truncated by range selection, as indicated by the light gray shading'
       );
     }
   }, [truncationConfigDependentAxisMin, truncationConfigDependentAxisMax]);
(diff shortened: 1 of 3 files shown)


jtlong3rd (approved):

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -548,15 +548,15 @@ function HistogramPlotWithControls({
...
+        ' Data has been truncated (light gray area) by range selection'
       );
     }
   }, [truncationConfigIndependentAxisMin, truncationConfigIndependentAxisMax]);
 
   useEffect(() => {
     if (truncationConfigDependentAxisMin || truncationConfigDependentAxisMax) {
       setTruncatedDependentAxisWarning(
-        'Data may have been truncated (light gray area) by range selection'
+        ' Data may have been truncated (light gray area) by range selection'
```
bobular, line 559: It would be nice not to need the leading space.  Can the Notification component take care of this?

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -548,15 +548,15 @@ function HistogramPlotWithControls({
       truncationConfigIndependentAxisMax
     ) {
       setTruncatedIndependentAxisWarning(
-        'Data has been truncated (light gray area) by range selection'
+        ' Data has been truncated (light gray area) by range selection'
```
bobular, line 551: I don't really like how these warnings read.  I'd prefer something like

```
Data has been truncated by range selection, as indicated in the plot by the light gray shading.
```
What do @danicahelb and @SheenaTomko think?

I know that's very long, but there is space for two lines of text in the blue notification area.  I tried a longer message and it will need some CSS changes so that it wraps and doesn't make the blue box much wider.

SheenaTomko:
I might change it slightly to "Data has been truncated by range selection, as indicated by the light gray shading."

src/lib/core/components/filter/HistogramFilter.tsx
moontrip: @bobular thank you for your review and comments 👍 I will try to make this at Notification component level.

src/lib/core/components/filter/HistogramFilter.tsx
moontrip: @bobular will try. It would be a bit tricky to change CSS though

moontrip: @bobular As suggested I have made single space at Notification component so the single space at histogram filter is removed

moontrip: @bobular also adjusted notification size with the new message by setting style from histogram filter component. Like,

histogramfilter-update1

bobular (approved):
That all looks great now, thanks!

Issue #310, opened 2021-08-12
Visualizations: do not infer dataShape for numeric & date variables to constrain at viz level not at data loading


danicahelb:
We need it to be possible to choose continuous variables that have <10 values (or whatever the cutoff for inferring categorical) as the x or y axis variable in scatterplots (and also line plots). For example, we would like to select "Study timepoint" (which has values of 1, 2, 3, 4, &5) for the x-axis in scatterplots in the SCORE S. mansoni Cluster Randomized Trial.

dmfalke:
Another possibility: instead of defining `dataShape` to drive the plot used in subsetting, maybe it would make more sense to add yet another annotation. It could be called something like `subsettingDataShapeOverride`.

danicahelb:
@steve-fischer-200 can you please cost this out & let the outreach team know if you'd like us to add a new annotation prop

d-callan:
im still really confused by this.. how does `subsettingDataShapeOverride` help the scatter plot viz? and whats the point of overriding the var constraints? they exist to prevent from plotting things that dont make sense logically. in this case, they prevent you from putting something categorical on the axis of an unjittered scatter plot. to me the options before us for phase 1 would be:

1. do nothing, and let plotting all points in the box plot be enough for now. could consider adding a proper beeswarm for a future phase. 
2. add jittering on the backend, rather than the front. this way we can jitter points in scatter as well as box. then update the var constraints for scatter to allow categorical vars on the xaxis. this option practically means that the scatter plot viz is technically either a scatter plot or beeswarm plot depending on the chosen vars, which is something i thought we were trying to avoid?  (conditional behavior based on logic not exposed to the user, rather than an explicit decision of theirs)

dmfalke:
@d-callan my suggestion is that, for cases where we want to use the table filter for continuous vars, we can use that annotation property as a directive for the subsetting tab. The `dataShape` property would still be continuous, so it can be used as such in visualizations. Does that make sense?

d-callan:
I thought in Danica's example the shape is categorical.

danicahelb:
@d-callan the `subsettingDataShapeOverride` makes sense to me because there are instances where the data is actually continuous but is being inferred as categorical because there are less than some threshold # of possible values. Users will commonly want to plot "Study timepoint" (with possible values of 1, 2, 3, 4, and 5) for longitudinal studies.

I'd love to add jittering to the scatterplots. For phase 1, opacity is good enough to help with overplotting

[...]

PR #406, opened 2021-09-09
Fix/402/max num values var constraint

### LINKED ISSUE #402: Constrain variables based on `maxValues`
The backend provides this property, but the frontend is not utilizing it.

This is needed to prevent variables with > 8 values from being used as overlay variables, and other similar cases.

d-callan:
should resolve #402, and is related to #310

this appears to be minimally functional to me, though there may be better ways to do things, or silly mistakes ive made. i want all of those thoughts. 

also @dmfalke had a comment in a side conversation earlier that we likely want to apply the constraints to only the Variables rather than the whole tree (including categories/ VariableTreeNodes).


d-callan:
having got this working locally in combination w veupathdb/edadataservice#88 ive now realized theres some work to do in plot.data for this as well. i mean to do that first, then return to potentially change VariableTreeNode -> Variable. Im off next week though, so if this is needed before that, please feel free.

src/lib/core/types/study.ts
```diff
@@ -57,13 +57,13 @@ export const VariableTreeNode_Base = t.intersection([
...
     displayType: VariableDisplayType,
     dataShape: VariableDataShape,
+    distinctValuesCount: t.number,
   }),
 ]);
 
 const Variable_Base = t.intersection([
   VariableTreeNode_Base,
   t.type({
-    distinctValuesCount: t.number,
```
dmfalke, line 66: I don't think this is necessary. When you are using this property in `src/lib/core/utils/data-element-constraints.ts`, the type has been narrowed to a non-category variable, due to the prior `variable.type === 'category'` condition. That condition also means that you don't need to change the signature of the function.

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -62,6 +63,7 @@ export function excludedVariables(
  */
 function variableConstraintPredicate(
   constraint: DataElementConstraint,
+  //maybe this should be variable: Variable - from dave
```
dmfalke, line 66: We don't need to do this, since we are testing against variables that have `type = 'category'`. See comment above.

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -71,6 +73,10 @@ function variableConstraintPredicate(
       constraint.allowedShapes.includes(variable.dataShape)) &&
       (constraint.allowedTypes == null ||
         constraint.allowedTypes.includes(variable.type)) &&
+      (constraint.maxNumValues == null ||
+        //this case should be category anyway, but is that a strong enough thing to assume here?
+        variable.distinctValuesCount === undefined ||
```
dmfalke, line 78: This can also be removed, due to the type narrowing, if the changes in `src/lib/core/types/study.ts` are undone.

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -133,9 +140,18 @@ export function flattenConstraints(
       const shapeIsValid =
         isEmpty(constraint.allowedShapes) ||
         constraint.allowedShapes?.includes(variable.dataShape!);
+      const passesMaxValuesConstraint =
+        constraint.maxNumValues === undefined ||
+        variable.distinctValuesCount === undefined ||
```
dmfalke, line 145: This can also be removed, due to the type narrowing, if the changes in `src/lib/core/types/study.ts` are undone.

dmfalke (changes requested):
This looks pretty good! I made some suggestions to simplify some of the logic. Let me know what you think.

src/lib/core/types/study.ts
d-callan: i saw `variable.type === 'category'`, and was slightly confused by it tbh. maybe i was interpretting 'variable' overly literally. but i didnt understand how a variable could be of type 'category'. seemed the condition should always be false, if variable were really a variable.
im mostly guessing really, but i suspect `excludedVariables` handles more than variables, and the narrowing of scope is actually doing something there. and that in `flattenVariables` were seeing only variables, and the narrowing of scope isnt necessary.
in any case i suppose you're right, and i did try that first actually. think vscode started giving me red squigglies though, and my ocd wouldnt allow that.

src/lib/core/utils/data-element-constraints.ts
d-callan: is there any reason to prefer one method over the other? I would have guessed, possibly naively, that just explicitly handling only Variables rather than VariableTreeNodes would be preferred to testing for the condition of meeting a category. what if we introduce some new thing, other than categories, than can be members of VariableTreeNode and not Variable?

src/lib/core/types/study.ts
dmfalke: There is not a large distinction made between "variables" and "categories" in the front end or back end code. I tried to make a stronger distinction in the front end code with the types `Variable` and `VariableCategory`. As far as seeing errors in VSCode, I played with these changes and didn't get any errors. Maybe there was something else causing it?

src/lib/core/utils/data-element-constraints.ts
dmfalke: They are really equivalent. In both cases, we would have to deal with categories. The way we have it now simplifies the calling code, and gives us an opportunity to apply constraints to categories, if we decide to in the future.

src/lib/core/types/study.ts
d-callan: very possible. hopefully ill get to this soon.

dmfalke (approved):

PR #431, opened 2021-09-13
update analysisCache when an analysis is saved

### LINKED ISSUE #331: Name is not preserved when previously saved analysis is reopened
When I choose a previously made analysis (QA3; screenshot 1), when it opens it does not have the name QA3; it opens as unnamed analysis (screenshot 2). Later it gets saved as "Unnamed analysis" unless I give it a name (screenshot 3). Analysis name should be preserved when working with a previously made analysis. 

Screen Shot 2021-08-19 at 2 39 34 PM (2)
Screen Shot 2021-08-19 at 2 42 51 PM (2)
Screen Shot 2021-08-19 at 2 43 19 PM (2)

### LINKED ISSUE #366: General: work on an analysis is not being saved
My work on an analysis is being saved when i close out of qa.clinepidb.org, but it is not being saved as i move through the website using either the  "Search a study" drop-down or navigating to the homepage and clicking on the explore icon on the card. Something also seems weird with previous analyses, so for this ticket I first deleted all of my previous analyses.

I now have no analyses in the "my analyses" table, and i open a GEMS1 analysis from the study card. I apply 2 filters (Country & Case or Control Participant) to select 1394 Bangladesh cases. the analysis is labeled "unnamed analysis"

I click on clinepidb (top left) to get back to the homepage. I next click immediately back on the explore icon for GEMS1. Now I have 3859 participants and the filter for Cases has been removed. 

The same thing happens when i am in my unnamed GEMS1 analysis and have my selected population of 1394 Bangladesh cases and click on GEMS1 analysis icon in the search a study dropdown.

dmfalke:
An in-memory cache was recently added for analyses, to prevent excessive requests for analysis objects. The cache was not being updated when an analysis is saved, which would cause changes to be lost if a user navigated between pages on the website.

This PR ensures the cached analysis is updated when saved.

fixes #366
fixes #331

### DIFF
--- a/src/lib/core/hooks/analysis.ts
+++ b/src/lib/core/hooks/analysis.ts
@@ -118,6 +118,7 @@ export function useAnalysis(analysisId: string): AnalysisState {
     if (analysis == null)
       throw new Error("Attempt to save an analysis that hasn't been loaded.");
     await analysisClient.updateAnalysis(analysis);
+    analysisCache[analysis.id] = analysis;
     setHasUnsavedChanges(false);
   }, [analysisClient, analysis]);
 


bobular (approved):
I'd somehow missed the original implementation of the cache so spent a bit of time understanding it.

This fix looks good too.  Managed to reproduce the issue (in `main`) using browser history forward/back and this fixes it.

jtlong3rd (approved):

PR #433, opened 2021-09-13
"Quantized" P-value formatting

### LINKED ISSUE #410: p values need to be fixed
For p values on plots/tables, we want to use <0.0001, <0.001, and <.01 and only if it’s above .01 then write it out. 

The way it currently appears is shown below. 
Screen Shot 2021-08-26 at 4 30 17 PM

bobular:
Fixes #410 

Ready for code review.

Question for @chowington and @d-callan 
I notice in the [client's data-api](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/core/api/data-api.ts#L365), that the expected type for pvalue responses from the back end is `number | string` - I can't find any information about why/when it might return a string.  In my tests it has always been a number. I've kept the option of a string response, but the new quantize function will not attempt to process a string value - and it will be displayed to the user as-is.  Is this satisfactory, or should we update the client's data-api rules (to number-only)?

image

### DIFF
--- a/src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
@@ -31,6 +31,7 @@ import Tabs from '@veupathdb/components/lib/components/Tabs';
 // import axis label unit util
 import { axisLabelWithUnit } from '../../../utils/axis-label-unit';
 import { PlotRef } from '@veupathdb/components/lib/plots/PlotlyPlot';
+import { quantizePvalue } from '../../../utils/analysis';
 
 const plotDimensions = {
   width: 750,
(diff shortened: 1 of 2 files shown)


jtlong3rd (approved):

chowington:
@bobular I'm not sure at this point why the frontend has `number | string`. The API spec has it as always `number`. It could be that early on, I was getting `'NA'` responses or something similar under particular circumstances (which I don't currently remember). I'd be okay with making it just `number`.

bobular:
Thanks for the info @chowington - since Danielle is on leave this week, let's not rock the boat. 

I've made a [tidy-up ticket](https://github.com/VEuPathDB/web-eda/issues/442) in this repo directly which can be handled in Phase 2 or beyond.

PR #434, opened 2021-09-13
Plural entity names

chowington:


chowington:
Merging due to triviality

PR #444, opened 2021-09-16
Fix logic for constraining variables based on dataElementDependencyOrder

### LINKED ISSUE #318: Visualizations: Some stratification vars produce a 400 response
The service response indicates an issue with entities. See an example below.

image

dmfalke:
The previous logic used was incorrect. Entities were not properly excluded based on ancestor lineage. The comments were also a bit off. This PR fixes the logic and comments. 

fixes #318

### DIFF
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -6,7 +6,7 @@ import {
   DataElementConstraintRecord,
   excludedVariables,
   flattenConstraints,
-  ValueByInputName,
+  VariablesByInputName,
 } from '../../utils/data-element-constraints';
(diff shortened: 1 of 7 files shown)


jtlong3rd (approved):

bobular:
Gonna review in a minute!

bobular (approved):
I did extensive testing and couldn't break it :-) 

I read through the code but not with a fine toothed comb.

My only comment really is that the `values` prop and `prevValue` and `nextValue` is confusing because really the values are variables.  `prevSelectedVariable` and `nextSelectedVariable` maybe?

dmfalke:
Thanks for the feedback, @bobular. I think you're right that the variable names I chose are not very good. I will change them before merging.

PR #466, opened 2021-09-21
adding axis range for both X/Y axis at Scatter plot

moontrip:
Added X/Y axes' ranges for Scatter plot viz. Also changed util's name concerning margin so that it can be used for both X and Y axes. 5% margin is utilized for both axes not to trim marker(s).

### DIFF
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -59,8 +59,10 @@ import {
 } from '@veupathdb/components/lib/types/plots/addOns';
 // import variable's metadata-based independent axis range utils
 import { defaultIndependentAxisRange } from '../../../utils/default-independent-axis-range';
-import { independentAxisRangeMargin } from '../../../utils/independent-axis-range-margin';
+import { axisRangeMargin } from '../../../utils/axis-range-margin';
 import { VariablesByInputName } from '../../../utils/data-element-constraints';
+// util to find dependent axis range
+import { defaultDependentAxisRange } from '../../../utils/default-dependent-axis-range';
 
 const plotDimensions = {
   width: 750,
@@ -70,8 +72,9 @@ const plotDimensions = {
 // define PromiseXYPlotData
 interface PromiseXYPlotData extends CoverageStatistics {
   dataSetProcess: XYPlotData;
-  yMin: number;
-  yMax: number;
+  // change these types to be compatible with new axis range
+  yMin: number | string | undefined;
+  yMax: number | string | undefined;
 }
 
 // define XYPlotDataResponse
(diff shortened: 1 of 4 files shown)


jtlong3rd (approved):

src/lib/core/utils/default-dependent-axis-range.ts
```diff
@@ -0,0 +1,98 @@
+import { Variable } from '../types/study';
+import { NumberOrDateRange } from '@veupathdb/components/lib/types/general';
+
+export function defaultDependentAxisRange(
+  variable: Variable | undefined,
+  plotName: string,
```
bobular, line 6: I noticed that `plotName` is unused.
Well actually my editor did, lol...

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -291,12 +294,26 @@ function ScatterplotViz(props: VisualizationProps) {
...
 
+  // find deependent axis range and its margin
+  const defaultDependentRangeMargin = useMemo(() => {
+    //K set yMinMaxRange using yMin/yMax obtained from processInputData()
+    const yMinMaxRange =
+      data.value != null
+        ? { min: data.value.yMin, max: data.value?.yMax }
+        : undefined;
+
+    const defaultDependentRange = defaultDependentAxisRange(
```
bobular, line 308: Do we have time to refactor `defaultDependentAxisRange` and `defaultIndependentAxisRange` into one function? It seems they do similar things.  The former has an extra range input (the [yMin, yMax] range from the actual data) - but the logic is the same: "given a bunch of ranges, calculate the union of them".  Also, instead of passing the plot type, we could pass a boolean directive saying whether or not zero should be including in the range-min calculations.

@dmfalke - should we add this to Technical Debt or do it now?  I think probably the former?

bobular (approved):
Code looks good, and tested with GEMS1 - the TAC results have annotated ranges for sure.  It looks like the dates no longer all have ranges from 2003?

image

src/lib/core/utils/default-dependent-axis-range.ts
moontrip: @bobular first of all thank you for your reviews! I just leave plotName props in there as it may be possibly used for other plot, e.g., line plot, density plot, etc.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: @bobular Yes you are correct. I also thought of a combined function like you said, but it is a bit complicated to check all possible cases at once so I just separately made it at different function for faster implementation. In addition, perhaps a prop like isZeroMin or something like that may replace plotName prop as you said. 

I think we may refactor these two items later.

dmfalke: Let's add it to Technical Debt. Thank!

moontrip: @dmfalke Thank you for your suggestion! I will merge this PR and make a technical debt one as well

PR #474, opened 2021-09-22
Handle `integer` dataType

### LINKED ISSUE #472: Error message appearing for data with data_shape = continuous, data_type = integer
Steve updated the services so that integer data won't appear with a decimal place, and Jay updated the loading pipeline so type = integer when the type is numeric and precision = 0. 

However, now that Jay has reloaded PRISM2, the histograms for a number of variables has disappeared and been replaced by the error message "This variable type is not currently handled." Jay checked the variables and reported that the data appear to be loaded correctly with data_shape = continuous, data_type = integer. 

Here's a list of some of the variables where this is occurring: Health facility distance, Market distance, Time since last fumigation, Female Anopheles gambiae count, Gravid Anopheles gambiae count, Bloodfed Anopheles gambiae count, Unfed Anopheles gambiae count, Unable to assess Anopheles gambiae count, Female Anopheles count, Anopheles dissected for parity count, Height, Weight, Temperature, Time since enrollmentr
Screen Shot 2021-09-20 at 1 11 17 PM

dmfalke:
For now, there is nothing special to do on the front end for integer types. They are just being treated as `NumberVariable`. We can update as needed in the future.

fixes #472

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -29,7 +29,7 @@ import { usePromise } from '../../hooks/promise';
 import { AnalysisState } from '../../hooks/analysis';
 import { useSubsettingClient } from '../../hooks/workspace';
 import { DateRangeFilter, NumberRangeFilter } from '../../types/filter';
-import { StudyEntity, StudyMetadata } from '../../types/study';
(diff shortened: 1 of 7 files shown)


dmfalke:
I'm going to merge, sans approval. Outreach is eager to QA the new data type. Feel free to post comments anyway.

bobular:
Thanks - had a quick look. No comments :+1:

PR #476, opened 2021-09-22
Fix multifilter checkbox unchecking bug

bobular:
To reproduce:

Go to a multifilter and check two boxes in two different sub-variables.

Then uncheck either of those.  It will nix the entire filter.

The solution might be a bit verbose!  Happy for alternative suggestions.


dmfalke (approved):
Good catch!

Issue #358, opened 2021-08-31
Subsetting: update distributions for variables within multifilters


danicahelb:
As discussed in 8/20 EDA WG:

Update the distributions for variables within the multifilter, for both ANY and ALL

Note that current functionality on live site (any vs all) is NOT correct: 

- If ANY is selected: distributions do NOT update
- If ALL is selected: distributions update to show subset vs entire dataset

Both ANY and ALL should update for EDA

![image](https://user-images.githubusercontent.com/25450900/131589761-aeda7c1e-e909-45bc-8c69-8aeded4bfa85.png)

![image](https://user-images.githubusercontent.com/25450900/131589791-f27c7621-9a24-49eb-ba1b-08e6d5c37de6.png)

![image](https://user-images.githubusercontent.com/25450900/131589816-cde0e4b8-5f75-4346-8bd5-29cd7066feb6.png)

![image](https://user-images.githubusercontent.com/25450900/131589835-4fa0172f-ddea-4a04-bc19-b627be8fb12c.png)

danicahelb:
I'm ok with implementing the functionality of the live site for EDA Phase 1 (so supporting distribution updates for **ALL** but not for **ANY**

I will make a phase 2 ticket for **ANY**

bobular:
If any work needs doing for Phase 1 - I think this needs making clearer.  I'm confused by this whole ticket.

Why update the red/grey bars at all?

If users need to be able to explore which are the common combinations, is what we need here (or as a viz) actually an UpSet plot?

bobular:
Notes from discussion with Danica:

1. Live site behaviour means no updates when "ANY" option is used.  
    - Some feedback for users would be good (note: absent from live site)
    - Danica suggests graying out the entire Distribution column and maybe adding some text in the heading
2. Try to remove update button, and have the counts refresh automatically (in ALL mode), but with throttling/debouncing or other measures to prevent excessive requests going to back end.

bobular:
Hi @danicahelb - I hope you are feeling better.

I have addressed 2. above - the distribution bars update automatically in "all" mode. The button is removed, except for >15 sub-variables.  Even with the button, the update is automatic, but you can make it happen faster by clicking the button.

[...]

PR #477, opened 2021-09-23
Testing out serial MultiFilter distribution requests

bobular:
With this draft PR, I demonstrate the speed of serialising the back end calls to update the pink/grey bars in the MultiFilter table.  See the issue here: https://github.com/VEuPathDB/web-eda/issues/358 

It takes 10 seconds to complete the process for GEMS1A Cooking fuel 
 
distribution-calls-in-serial

Note: most of the diff is due to indentation changes - I really just swapped out the `Promise.all ... leaves.map()` for `serialArrayPromise(leaves, ...)` - it's not a big change at all.

### DIFF
--- a/src/lib/core/components/filter/MultiFilter.tsx
+++ b/src/lib/core/components/filter/MultiFilter.tsx
@@ -171,64 +171,62 @@ export function MultiFilter(props: Props) {
   const leafSummariesPromise = usePromise(
     useCallback(() => {
       const otherFilters = JSON.parse(otherFiltersJson);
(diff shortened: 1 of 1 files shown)


bobular:
That 10 seconds would be less when not using the dev server's back end proxy, but still we've decided to abort this approach in favour of [debouncing the initial triggering of these multiple requests](https://github.com/VEuPathDB/web-eda/pull/480).

The poorly named `serialArrayPromise` function might be of interest/use in the future though.

PR #480, opened 2021-09-23
Debounced entity count and multifilter distribution updates

### LINKED ISSUE #358: Subsetting: update distributions for variables within multifilters

bobular:
Fixes #358 

I kept the update button (text changed a bit) when there are more than 15 variables.  This allows the user to force the update if they can't wait for the debouncer to time out.

Entity counts are debounced centrally, but we still have the problem of calling them from three different places in the visualization tab (ED, birds eye plot, coverage table).  I won't address that here!

Keeping it in draft for now until I address the rest of the issue (communication of difference between `all` and `any` modes on the distribution bars; it's a tricky one)


dmfalke:
Nice work! Some preliminary thoughts/comments:

1. We should think about making a reusable `useDebouncedEffect` hook.
2. I don't understand why we need to increase the time based on the number of variables. Can you explain this more?
3. While debouncing is great for preventing lots of UI updates, it does not cause the underlying network requests to be cancelled. In this particular case, that would be ideal, since we are trying to avoid a cascade of blocked requests. Making that happen will take some thought and work. It will involve utilizing a `Fetch` API primitive ([example usage](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort#examples), which you can see is quite verbose). It's not clear what is the best way to expose this. I've been tempted to use something akin to https://github.com/VEuPathDB/WDKClient/blob/master/Client/src/Utils/Task.ts as our primitive for API requests. The `Task` class has the notion of cancellation built in, so it would be relatively straightforward to hide the verbosity in the example above, but at the expense of dispensing of the `Promise`-based API. Sorry to ramble on. I'm hoping to stimulate more conversation around this.

bobular:
>     1. We should think about making a reusable `useDebouncedEffect` hook.

That would be good. I've thought about it. ;-) So this would just be for the entityCounts debouncing (for now)?
 
>     2. I don't understand why we need to increase the time based on the number of variables. Can you explain this more?

This is to avoid a second request being made while the first one is still in progress (because we can't cancel requests - see 3.)

>     3. While debouncing is great for preventing lots of UI updates, it does not cause the underlying network requests to be cancelled. In this particular case, that would be ideal, since we are trying to avoid a cascade of blocked requests. Making that happen will take some thought and work. It will involve utilizing a `Fetch` API primitive ([example usage](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort#examples), which you can see is quite verbose). It's not clear what is the best way to expose this. I've been tempted to use something akin to https://github.com/VEuPathDB/WDKClient/blob/master/Client/src/Utils/Task.ts as our primitive for API requests. The `Task` class has the notion of cancellation built in, so it would be relatively straightforward to hide the verbosity in the example above, but at the expense of dispensing of the `Promise`-based API. Sorry to ramble on. I'm hoping to stimulate more conversation around this.

dmfalke:
I'm not saying that we shouldn't debounce the multifilter requests. I'm just saying that it still has deficiencies.

@jtlong3rd you might be interested in and have thoughts about cancelling network requests and how we can approach that generically. It's something we've been punting on, but we should consider addressing soon.

That being said, if we do end up with an endpoint that can deliver distributions for multiple variables at once, this may be less of a priority.

bobular:
@dmfalke and I discussed remaining UI/UX issues, e.g. greying out or other explanations for any/all mode as suggested by @danicahelb .  These don't exist on the live site and we think more discussion, mock-ups etc are required than we have time for in Phase 1.

So we will push this out for testing as-is.

dmfalke (approved):

jtlong3rd (approved):
One small comment for our future selves - it may be worth checking out useHooks' [useDebounce](https://usehooks.com/useDebounce/), as well as the [library](https://github.com/xnimorz/use-debounce) it was inspired by.

PR #481, opened 2021-09-24
Quick and dirty fix for multifilter provider label

### LINKED ISSUE #380: For multifilters, have the provider label space show the variable labels
At the moment, multifilters say "No Provider Label available" above the data table. Instead, we'd like this to show the names of all the variables that are displayed as part of the multifilter. This would be an improvement beyond the live site (which just removes that line altogether), so we can move this to a later phase if warranted. 

Screen Shot 2021-09-03 at 3 07 04 PM

dmfalke:
fixes #380

### DIFF
--- a/src/lib/workspace/Variable.tsx
+++ b/src/lib/workspace/Variable.tsx
@@ -37,7 +37,8 @@ export function VariableDetails(props: Props) {
         <h3>{axisLabelWithUnit(variable)}</h3>
         <div className={cx('-ProviderLabel')}>
           <div className={cx('-ProviderLabelPrefix')}>
-            Original variable name:
+            Original variable{' '}
+            {MultiFilterVariable.is(variable) ? 'names' : 'name'}:
           </div>
           &nbsp;
           {MultiFilterVariable.is(variable)
@@ -81,11 +82,14 @@ function findMultifilterVariableLeaves(
   variable: MultiFilterVariable,
   variablesByParentId: Record<string, VariableTreeNode[] | undefined>
 ): Variable[] {
-  if (variable.parentId == null) return [];
+  // Find all children of `variable`
   const variables = variablesByParentId[variable.id] ?? [];
-  for (const child of variables) {
-    for (const grandchild of variablesByParentId[child.id] ?? []) {
-      if (grandchild) variables.push(grandchild);
+  // For each item in `variables`, find its children and push to `variables`.
+  // These items will then be included in the iteration of `variables`, thus
+  // we will find all descendents of `variable` provided to this function.
+  for (const variable of variables) {
+    for (const child of variablesByParentId[variable.id] ?? []) {
+      if (child) variables.push(child);
     }
   }
   return variables.filter(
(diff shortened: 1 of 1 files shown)


jtlong3rd (approved):

src/lib/workspace/Variable.tsx
```diff
@@ -37,7 +39,18 @@ export function VariableDetails(props: Props) {
           <div className={cx('-ProviderLabelPrefix')}>
             Original variable name:
           </div>
-          &nbsp;{variable.providerLabel}&nbsp;
+          &nbsp;
```
bobular, line 42: It's a minor pain, but I'd pluralise "Original variable name" for the multifilter.

src/lib/workspace/Variable.tsx
```diff
@@ -63,3 +76,19 @@ export function VariableDetails(props: Props) {
     </ErrorBoundary>
   );
 }
+
+function findMultifilterVariableLeaves(
+  variable: MultiFilterVariable,
+  variablesByParentId: Record<string, VariableTreeNode[] | undefined>
+): Variable[] {
+  if (variable.parentId == null) return [];
```
bobular, line 84: What's this check on `variable.parentId` for?

src/lib/workspace/Variable.tsx
```diff
@@ -63,3 +76,19 @@ export function VariableDetails(props: Props) {
...
+  const variables = variablesByParentId[variable.id] ?? [];
+  for (const child of variables) {
+    for (const grandchild of variablesByParentId[child.id] ?? []) {
+      if (grandchild) variables.push(grandchild);
```
bobular, line 88: I'm not familiar with the layout of the data structure, so I'll take your word for it that we only want grandchildren (and not the children).

bobular (changes requested):
I'd just change the plural thing (don't change the help text).

For GEMS1A bacteria in stool - it takes up a lot of screen space.  Maybe add 
```
.EDAWorkspace-ProviderLabel {
  max-height: 5em;
  overflow: scroll;
}
```
or similar?

The other comments are just due to me not understanding!

src/lib/workspace/Variable.tsx
dmfalke: Oh, that's not needed. It was left over from an incorrect attempt. I will remove it.

src/lib/workspace/Variable.tsx
dmfalke: The variable name `grandchild` is not really accurate. What this loop is doing is: for each variable, find its children (those whose `parentId` is the same as the variable's `id`), and add it to the array `variables`. As the array is being iterated, elements that are `push`ed to it will be included in the iteration.

dmfalke:
@bobular I addressed your suggestion about pluralization. I also added some comments and changed some variable names in the help function.

Regarding scrolling the provider names: I'm going to leave as-is for now. If anything, I think the section could be collapsible, but I'm not convinced it's better.

PR #485, opened 2021-09-28
thumbnail gray-out, styling, etc

moontrip:
This PR includes several issues, including the latest update of issues. The most tricky work was to make thumbnail have gray-out box per filters. Tried several approaches to satisfy the following various use cases:

use case 1: switching between Browse and subset and a full screen plot
  - filters are applied to the plot immediately, so no gray-out for the corresponding thumbnail when minimized (moved to thumbnail view)

use case 2: under a full screen plot with filters, filters are updated through Global filter dialog (i.e., removed)
  - for the plot, this is the same to scenario 1's result as the changes of filter(s) is/are immediately applied to the plot
  - however, the change of filter(s) through the dialog affects other pre-existing plots, so need to check the status of filters for other plots' thumbnails
  - in addition, the behavior of each thumbnail's status (grayed out or not) is not the same each other
    - for example, some viz plots may be made without filters, so removing all filters at a full screen plot via Global filter dialog would not gray out those plots made without filters

use case 3: switching between Browse and subset and Visualization (thumbnail view) tab
  - if filter(s) is/are applied Browse and subset, then thumbnail(s) should be grayed out immediately
    - however, similar to the last case in the Scenario 2, there may be a case that some plots might already be plotted with the same filter condition, so it would not be grayed out

use case 4: under Visualization tab (thumbnail view)
  - removing filter would immediately make all plots be grayed out

Finally settled down to use new vizConfig props, so-called, `currentPlotFilters`, at each plot viz which is compared to global `filters` props. From my tests, this approach works for all above use cases.

### DIFF
--- a/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -130,21 +130,34 @@ function ConfiguredVisualizations(props: Props) {
                       </button>
                     </Tooltip>
                   </div>
-                  <div>
-                    <Tooltip title="View fullscreen">
-                      <Link replace to={`${url}/${viz.visualizationId}`}>
-                        <i className="fa fa-arrows-alt"></i>
-                      </Link>
-                    </Tooltip>
-                  </div>
                 </div>
-                {viz.descriptor.thumbnail ? (
-                  <img alt={viz.displayName} src={viz.descriptor.thumbnail} />
-                ) : (
-                  <div className={cx('-ConfiguredVisualizationNoPreview')}>
-                    Preview unavaiable
-                  </div>
-                )}
+                {/* add the Link of thumbnail box here to avoid click conflict with icons */}
+                <>
+                  <Link replace to={`${url}/${viz.visualizationId}`}>
+                    {viz.descriptor.thumbnail ? (
+                      <img
+                        alt={viz.displayName}
+                        src={viz.descriptor.thumbnail}
+                      />
+                    ) : (
+                      <div className={cx('-ConfiguredVisualizationNoPreview')}>
+                        Preview unavaiable
+                      </div>
+                    )}
+                    {/* make gray-out box on top of thumbnail */}
+                    {!arraysEqual(
+                      props.filters,
+                      (viz.descriptor.configuration as any).currentPlotFilters
+                    ) ? (
+                      <div className={cx('-ConfiguredVisualizationGrayOut')}>
+                        Open to sync with
+                        <br /> current subset
+                      </div>
+                    ) : (
+                      ''
+                    )}
+                  </Link>
+                </>
               </div>
               <div className={cx('-ConfiguredVisualizationTitle')}>
                 {viz.displayName ?? 'Unnamed visualization'}
@@ -326,7 +339,7 @@ function FullScreenVisualization(props: Props & { id: string }) {
         </div>
         <Tooltip title="Minimize visualization">
           <Link replace to={`../${computationId}`}>
-            <i className="fa fa-window-restore"></i>
+            <i className="fa fa-window-minimize"></i>
           </Link>
         </Tooltip>
       </div>
@@ -369,3 +382,27 @@ function FullScreenVisualization(props: Props & { id: string }) {
     </div>
   );
 }
+
+/**
+ * functions to perform deep comparison between two array of objects (filters here)
+ * @param o1: object
+ * @param o2: object
+ * @param a1: array of objects
+ * @param a2: array of objects
+ * @returns boolean
+ * usage: arraysEqual(props.filters, viz.configuration.currentPlotFilters)
+ */
+
+const arraysEqual: any = (
+  a1: Record<string, string | string[]>[],
+  a2: Record<string, string | string[]>[]
+) => a1.length === a2.length && a1.every((o, idx) => objectsEqual(o, a2[idx]));
+
+const objectsEqual: any = (
+  o1: Record<string, string | string[]>,
+  o2: Record<string, string | string[]>
+) =>
+  typeof o1 === 'object' && Object.keys(o1).length > 0
+    ? Object.keys(o1).length === Object.keys(o2).length &&
+      Object.keys(o1).every((p) => objectsEqual(o1[p], o2[p]))
+    : o1 === o2;
--- a/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useEffect } from 'react';
+import React, { useCallback, useMemo, useEffect, useRef } from 'react';
 import {
   Route,
   Switch,
@@ -273,9 +273,17 @@ function FullScreenVisualization(props: Props & { id: string }) {
     [updateVisualizations, id]
   );
 
-  // update currentPlotfilters with the latest filters at fullscreen mode
+  // store a ref to the latest version of updateVisualizations
+  const updateVisualizationsRef = useRef(updateVisualizations);
+
+  // whenever updateVisualizations changes, update the updateVisualizationsRef
   useEffect(() => {
-    updateVisualizations((visualizations) =>
+    updateVisualizationsRef.current = updateVisualizations;
+  }, [updateVisualizations]);
+
+  // update currentPlotFilters with the latest filters at fullscreen mode
+  useEffect(() => {
+    updateVisualizationsRef.current((visualizations) =>
       visualizations.map((v) =>
         v.visualizationId !== id
           ? v
--- a/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useMemo, useEffect } from 'react';
 import {
   Route,
   Switch,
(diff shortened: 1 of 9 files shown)


moontrip:
Here is the list of issues/works covered in this PR

- gray-out thumbnail per filter change
- clickable thumbnail that leads to its full screen mode
- hide thumbnail icons but show them when mouseover to a thumbnail
- arrange/change icons (thumbnail and full screen icons)

Below screenshot shows a thumbnail view. Note that the second thumbnail shows its icons as it was captured when mouseover the second thumbnail - mouse point was not captured though.

thumbnail-gray-out1

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -124,6 +126,13 @@ function BarplotViz(props: VisualizationProps) {
     [updateConfiguration, vizConfig]
   );
 
+  // update currentPlotfilters with the latest filters at fullscreen mode
+  useEffect(() => {
+    updateVizConfig({
+      currentPlotFilters: filters,
+    });
+  }, [filters]);
```
jtlong3rd, line 134: Have you tried putting this effect into the root `FullscreenVisualization` component? If doing that works, that would save us the hassle of needing to include this effect in every new visualization.

(Sidenote: we should consider a similar DRY-up of the `updateThumbnail` effects. That will be trickier because currently, each viz's `updateThumbnail` effect has different dependencies.)

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -86,6 +86,8 @@ const BarplotConfig = t.intersection([
     overlayVariable: VariableDescriptor,
     facetVariable: VariableDescriptor,
     showMissingness: t.boolean,
+    // new config props to store filters at fullscreen mode
+    currentPlotFilters: t.unknown,
```
jtlong3rd, line 90: Instead of adding `currentPlotFilters` to each viz config, you could make `currentPlotFilters` a top-level property of the `VisualizationDescriptor`:

```typescript
/**
 * Metadata for the visualization object stored in user's analysis
 */
export type VisualizationDescriptor = TypeOf<typeof VisualizationDescriptor>;
export const VisualizationDescriptor = intersection([
  type({
    type: string,
    configuration: unknown,
  }),
  partial({
    thumbnail: string,
    currentPlotFilters: t.array(Filter)
  }),
]);
```

src/lib/core/hooks/findOutputEntity.ts
```diff
@@ -8,7 +8,7 @@ export function useFindOutputEntity(
   // need to add string at Record's Type due to valueSpecConfig
   dataElementVariables: Record<
     string,
-    VariableDescriptor | string | boolean | undefined
+    VariableDescriptor | string | boolean | unknown | undefined
```
jtlong3rd, line 11: If you agree with my suggestion to make `currentPlotFilters` a top-level property of the `VisualizationDescriptor`, you'll be able to do away with the `unknown`.

```suggestion
    VariableDescriptor | string | boolean | undefined
```

jtlong3rd (changes requested):
Overall, this is a clean approach! I just have some inline suggestions.

moontrip:
> Overall, this is a clean approach! I just have some inline suggestions.

@jtlong3rd as usual, thank you for your detailed comments! 👍 I will carefully check your suggestions and try to address them - will leave a reply for each item after done 😃

src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -130,21 +130,34 @@ function ConfiguredVisualizations(props: Props) {
...
+                <>
+                  <Link replace to={`${url}/${viz.visualizationId}`}>
+                    {viz.descriptor.thumbnail ? (
+                      <img
+                        alt={viz.displayName}
+                        src={viz.descriptor.thumbnail}
+                      />
+                    ) : (
+                      <div className={cx('-ConfiguredVisualizationNoPreview')}>
+                        Preview unavaiable
```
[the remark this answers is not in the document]
moontrip: 😄 fixed!

src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -130,21 +130,34 @@ function ConfiguredVisualizations(props: Props) {
...
+                    ) ? (
+                      <div className={cx('-ConfiguredVisualizationGrayOut')}>
+                        Open to sync with
+                        <br /> current subset
+                      </div>
+                    ) : (
+                      ''
+                    )}
+                  </Link>
+                </>
```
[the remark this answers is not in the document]
moontrip: @jtlong3rd I have changed custom functions to Lodash's isEqual. Yes I know that it would be expensive so your idea is good to go. I have complied with 👍

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
moontrip: @jtlong3rd Actually I was debating where to put the new props, `currentPlotFilters`, either configuration or descriptor as you said. But I like your idea to place the props at descriptor as it does not need to change all plot vizs. I have changed to use the useEffect at `FullscreenVisualization` component and it seems to work fine as well 🎉

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
moontrip: @jtlong3rd as I said above, I also defined the `currentPlotFilters` props in there: actually there is no generic `t` in there so just used `array(Filter)`

src/lib/core/hooks/findOutputEntity.ts
moontrip: @jtlong3rd by complying with your suggestions, yes I don't need to define unknown in there so removed it.

moontrip:
@jtlong3rd I have complied with your suggestions and made push with updates. Detailed replies are made at each of your request above 👍

src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -275,6 +273,20 @@ function FullScreenVisualization(props: Props & { id: string }) {
...
+      visualizations.map((v) =>
+        v.visualizationId !== id
+          ? v
+          : {
+              ...v,
+              descriptor: { ...v.descriptor, currentPlotFilters: filters },
+            }
+      )
+    );
+  }, [filters]);
```
jtlong3rd, line 288: This effect has a missing `updateVisualizations` dependency, which is a [potential source of bugs](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies). Since a stale value of `updateVisualizations` could cause a user's analyses to be mis-saved, we should address.

As you probably discovered on your own, we can't just add `updateVisualizations` to the effect's dependency list, because this causes an infinite loop! (The effect changes the `visualizations`, which causes `updateVisualizations` to change, which causes `visualizations` to change, which causes `updateVisualizations` to change...)

The React docs have a few [solutions](https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often) for this issue, but only one immediately works here: making a ref to the latest version of `updateVisualizations` and using that ref in the effect.

```suggestion
  // store a ref to the latest version of updateVisualizations
  const updateVisualizationsRef = useRef(updateVisualizations);
  
  // whenever updateVisualizations changes, update the updateVisualizationsRef
  useEffect(() => {
    updateVisualizationsRef.current = updateVisualizations;
  }, [updateVisualizations]);
  
  // update currentPlotFilters with the latest filters at fullscreen mode
  useEffect(() => {
    updateVisualizationsRef.current((visualizations) =>
      visualizations.map((v) =>
        v.visualizationId !== id
          ? v
          : {
              ...v,
              descriptor: { ...v.descriptor, currentPlotFilters: filters },
            }
      )
    );
  }, [filters]);
```

jtlong3rd (commented):
Your changes look great! I had one last suggestion, and then this should be good to merge.

src/lib/core/components/visualizations/VisualizationsContainer.tsx
moontrip: @jtlong3rd Thank you for your detailed explanation! 👍 Honestly I didn't know about the useRef solution to resolve this kind of dependency issue, so it is great to learn about that! I have committed with your solution.

jtlong3rd (approved):

src/lib/core/components/visualizations/VisualizationsContainer.tsx
bobular: Also thanks for the education.
Presumably `updateVisualizationsRef` doesn't need to be in the dependencies.  Like `setState` functions from `useState` - they are guaranteed not to change?

bobular (approved):
I did some thorough testing and reviewed the post-Jamie code that looks great!

Doing the deep `isEqual()` test is great because if you change a filter, then change it back, your old thumbnails come back :-)

moontrip:
@bobular thank you for your review and thorough tests 👍

src/lib/core/components/visualizations/VisualizationsContainer.tsx
jtlong3rd: Yup, it's OK to omit `updateVisualizationsRef` from the effect dependencies, because it's just a referentially stable container. (As with `useState` setters, our linter is smart enough to permit that omission with out complaining.)

Also of note - you should also not add `updateVisualizatuionsRef.current` to the dependencies. If you do that, our linter will issue a warning. The reason is that `updateVisualizatuionsRef.current` is a mutable value which isn't being tracked by React, and so is not a valid effect dependency.

(Great blog post on this topic [here](https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/).)

Issue #330, opened 2021-08-19
UI: Fix styling of control widgets, eg, "Show missing data" button


bobular:
- [x] #558
- [x] check correct sentence case of labels that were previously all-capsified
- [x] don't gray out the text of show missing data switch
- [x] don't have the text respond to hover/focus
- [x] fix widths of x and y axis widgets in the histogram filter

See https://github.com/VEuPathDB/web-eda/pull/486 for more details

Question: should the widget button itself be brought into focus on mouse-over, even if the text isn't?
Answer: Yes

Question: hack this now or await input from @eikonomega ?
Answer: Do it now

Source: [2021-08-18 dev mtg](https://docs.google.com/document/d/1W1iVvO0n29E2ZDN3M8VRoctg29voNOm-td8xLAxcGSA/edit?pli=1#heading=h.lk8sbqszdvzq)

steve-fischer-200:
i think this is needed in phase 1, so cant want for our overall styling upgrade

chowington:
@bobular Can I get clarification on the scope of this ticket? Is this just for the "show missing data" (not verbatim) button, or, for instance, for any place that still has all-caps styling?

Also, the current behavior regarding your first question is that the switch widget (separate from the text label) does focus on hover, which I'm fine with keeping.

bobular:
Sorry for the delay - didn't see this until now.

I'm pretty sure the ALL-CAPS should be gone from the entire app.

Yes keep the button focus on hover - I think it's useful feedback that distinguishes active buttons from disabled/greyed out ones.

danicahelb:
I'd prefer that the text changes from grey to black when there are points with no data for the strata vars. But it is not a high priority for phase 2

bobular:
Ah yes, sorry I didn't quite implement that to the specs.  The button grays out as desired, but not the text.

chowington:
I removed the #574 subtask now that it has its own ticket that @asizemore is working on. Please re-review or close as appropriate.

aaronwlsong:
On beta. Website, the text does not change its color when I active and inactive. Tested in several studies, got the same result. 

![image](https://user-images.githubusercontent.com/61077845/139490737-fd3ccdd4-f01e-4c79-8a5b-73bb28803d95.png)

![image](https://user-images.githubusercontent.com/61077845/139490818-28b3b59d-dada-4c23-8f95-493d2e3e5016.png)

PR #486, opened 2021-09-29
Decapitalize globally (mostly)

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/330

This changes the MUI default of making certain text all caps, though it may be overridden by a component's individual theme context (even implicitly).


chowington:
Blocked by https://github.com/VEuPathDB/web-components/pull/215

bobular (approved):
Looks great - I just tidied up a few things (reset buttons can be much less wide now) - removed some colons, histoviz control vertical layout, and removed the focus effect on range input labels.

Was easier to do than describe or add as suggestions!

chowington:
@bobular Are the x and y axis controls supposed to look so different? In width, defaults, etc.

image

bobular:
You're right and on my screen (at certain window sizes) it's even worse...
(India ICEMR Fever Surveillance)
Date variables in particular have the problem - maybe something to do with squeezing the extra units menu in?
image


Would you be able to fix it please?  If it behaves similarly to non-date numeric vars (e.g. below), that will be fine in my opinion 
image

chowington:
I pushed something but I don't think I'm quite satisfied with it. It gives the x-axis div more room to grow, but the widgets within don't seem to be interested in the extra space. It looks better when the screen is smaller, as the x control gets more space than it used to. But my change as a side effect also adds space between the x and y controls when the screen is large, which breaks precedent. I don't think it's worth it right now to find a better solution, but I'm also willing to revert back to the above existing styling. @bobular What do you think?

Smallish screen

image

Large screen

image

bobular:
I had a play and the problem seems to be with the date-based inputs - they are much narrower.
The following kind of works - though doesn't look great on very wide screens.

[diff.txt](https://github.com/VEuPathDB/web-eda/files/7315921/diff.txt)

Issue #487, opened 2021-09-29
Styling: Use new icons for My Analyses and download current subset


asizemore:
Follow up to Creating New Icons issue #415

asizemore:
Closed with above PRs. Note, only changes icons. General styling will be fixed when addressing #450

SheenaTomko:
Icons look great. Showing the new table icon for "My analyses" on the study card and within the EDA; for the Study summaries table in the drop down menu. Showing the new table icon + download icon for downloading the subset table

Issue #450, opened 2021-09-19
General: styling of all icon/label links (buttons?) should match


danicahelb:
Styling: 
- label should always be positioned to the right of the icon
- use same color for the icons and labels
- "view and download" link on the subsetting tab should have the same styling as the “download”, “+ new analysis”, "my analyses" on the study header (font, text size, color)
- maybe make them look like buttons?

moontrip:
Hi @danicahelb May I ask your opinion about the last item, `make them look like buttons` ? I have tried to change the top links (`Download`, `New analysis`, and `My analyses`) to be button style as follows. Certainly, the Download is the current style (without change), the New analysis is a button style with a color (color may be changed), and the My analyses is applied to have outline so that it looks like a button. Do you have any preference among the three ? It would be very helpful for me. 

Also I just want to make sure that the view and download link you mentioned points out the top right table-like icon at the second screenshot ?

- **button styles**
![material-button1](https://user-images.githubusercontent.com/12802305/135498489-c70b4fa1-1cdd-45b6-80dd-38d1bddff8ad.png)


- **view and download link ?**
![view-and-download](https://user-images.githubusercontent.com/12802305/135498513-5cd813d2-dded-430b-864e-fcc72778271f.png)

danicahelb:
thanks for the mockups @moontrip! I've checked in with the ClinEpi team, and we prefer option (1) `Download` (so no need to make them look like buttons.

Yes, the "View and download" link is the top right table icon. The styling for that link is discussed in this ticket: https://github.com/VEuPathDB/web-eda/issues/449. Could you implement these changes while you are working on this ticket? 

Use the table overlaid with the download icon, and label it: “View and download” 
include mouseover text with more information: “View and download current subset of Communities”

Ann created this icon for "View and download" (https://github.com/VEuPathDB/web-eda/issues/415)
![image](https://user-images.githubusercontent.com/25450900/135522879-4ad3d429-d3f3-4a70-83e5-bdcae9ebba9d.png)

danicahelb:
Oh! and we want to use a new icon Ann created for `My analyses` 

![image](https://user-images.githubusercontent.com/25450900/135523493-f99b57cb-e1c3-463d-80c7-de9d1c1963cb.png)

moontrip:
[...]

PR #488, opened 2021-09-29
Update table icons

asizemore:
In part addresses web-eda issue #487
Related to but does not depend on EbrcWebsiteCommon PR [#70](https://github.com/VEuPathDB/EbrcWebsiteCommon/pull/70)

This PR makes the following changes:
1. Replaces My Analyses table icon with custom ebrc icon
2. Replaces download current subset icon with custom ebrc icon.

See below.
Screen Shot 2021-09-29 at 2 14 20 PM

One problem I ran into was using these icons in the mui button. The table icon for My Analyses isn't centered vertically. This is more obvious when looking at where the icon should lie:
Screen Shot 2021-09-29 at 1 35 25 PM

I investigated the bug for a bit but could not find any solution that worked. Another clue then surfaced - none of the icons are actually centered vertically within the mui button. I think the table one just is the most obvious.
Screen Shot 2021-09-29 at 1 46 14 PM

Any ideas on this? All the solutions I found either didn't work or seemed far too complicated for what I imagine would be a simple fix?

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "@types/debounce-promise": "^3.1.3",
     "@veupathdb/components": "^0.7.18",
     "@veupathdb/wdk-client": "^0.4.4",
-    "@veupathdb/web-common": "^0.1.7",
+    "@veupathdb/web-common": "^0.1.9",
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
     "io-ts": "^2.2.13",
--- a/src/lib/workspace/EDAWorkspaceHeading.tsx
+++ b/src/lib/workspace/EDAWorkspaceHeading.tsx
@@ -4,15 +4,23 @@ import Path from 'path';
 import { cx } from './Utils';
 import { useStudyRecord } from '../core';
 import { safeHtml } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
-import { Button, Tooltip, Icon } from '@material-ui/core';
+import { Button, Tooltip, Icon, makeStyles } from '@material-ui/core';
 import { LinkAttributeValue } from '@veupathdb/wdk-client/lib/Utils/WdkModel';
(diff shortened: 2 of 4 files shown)


dmfalke:
I wonder if the icon should be centered vertically in the svg? Not sure if that's a possibility.

dmfalke:
I see now that you did that in https://github.com/VEuPathDB/EbrcWebsiteCommon/pull/70. I played with this briefly and haven't been able to find a solution. I'll try a little bit more.

dmfalke:
I played with the `--descent` argument to https://github.com/Workshape/icon-font-generator/, and the results look promising:

## default (current):
image
---

## --descent 8:
image
---

This will likely impact other icons, so it might be good to test with a microbiome dev site, as well.

asizemore:
Ahh interesting I will try it! Alternatively I can move the table up slightly in the svg and see how that works as well. I'm concerned about this latter strategy when the icon gets scaled but on the other hand it would only affect the table icon

asizemore:
Actually I wonder if it's not the icons/icon generation but has to do with placing that one specifically in the mui button start icon? Observe that on the cards where it's just an icon (not part of an mui component) the alignment with the text is great:

Screen Shot 2021-09-29 at 2 11 33 PM

asizemore:
So I might try something like [this](https://stackoverflow.com/questions/66146921/align-icon-inside-material-ui-button). More soon!

dmfalke:
:exploding_head: nice find!

asizemore:
@dmfalke what do you think of the new solution? Feels overkill for the simple task but works pretty well!
Screen Shot 2021-10-01 at 3 11 29 PM

dmfalke (approved):

dmfalke:
I'm happy with it!

asizemore:
Great! I'll go merge the related ebrcwebsitecommon PR and then will do the usual upgrading web-common here when ready?

asizemore:
I'll go ahead and merge this and then when addressing issue #450 will upgrade web-common if necessary.

PR #494, opened 2021-09-30
Scatter visualisation - implement limit on number of data points

### LINKED ISSUE #324: Visualizations: user friendly abort message when trying to plot too much data
As discussed in 8/16 EDA meeting:

Don’t implement new variable constraints

If there is too much data, show an error message indicating that the user is trying to plot too much data & that they should subset the data further or try a different plot type 

Data service API will need user friendly abort messaging (Backend currently returns 400 response (or 422 user input error) and the client is displaying this message)

We recognize that the backend will try to make the plot for ~30s before throwing up the error message, this delay is OK

bobular:
Resolves #324 

I have set the limit very low (5,000) for now - just so it's easier to test.  Should we increase this before or after testing from the QA crew? I think they may also have suggestions about the wording. 

(I wasn't sure how to make links to open a new histogram or boxplot.)

image

### DIFF
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -63,6 +63,10 @@ import { axisRangeMargin } from '../../../utils/axis-range-margin';
 import { VariablesByInputName } from '../../../utils/data-element-constraints';
 // util to find dependent axis range
 import { defaultDependentAxisRange } from '../../../utils/default-dependent-axis-range';
+import { useRouteMatch } from 'react-router';
+import { Link } from '@veupathdb/wdk-client/lib/Components';
+
+const MAXALLOWEDDATAPOINTS = 5000;
 
 const plotDimensions = {
   width: 750,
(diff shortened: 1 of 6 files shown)


dmfalke (approved):
This all looks good to me. I don't have any suggestions about what the limit should be. I'd have to play around with some configurations to get a sense for what's pushing the limits.

bobular:
See main issue for discussion on setting the limit.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -371,9 +377,24 @@ function ScatterplotViz(props: VisualizationProps) {
           }}
         >
           <i className="fa fa-warning" style={{ marginRight: '1ex' }}></i>{' '}
-          {data.error instanceof Error
-            ? data.error.message
-            : String(data.error)}
+          {data.error instanceof Error ? (
+            data.error.message.match(/400.+too large/is) ? (
```
bobular, line 381: Hi @dmfalke - just closing this tab. This was the line that is a bit dodgy.

PR #495, opened 2021-09-30
Fix small typo that was causing the configured "example analyses auth…

jtlong3rd:
…or" to be ignore in local development


jtlong3rd:
I'll be self-merging this because the fix is self-explanatory.

PR #500, opened 2021-10-01
Data shape fixes for non-continuous vars in histogram viz

### LINKED ISSUE #491: Visualization: No response for chosen variables in histogram
In some india studies, histogram does not respond chosen variables. No any warning messages, just looks like always loading data. 

eg.1. In _india_ICMER_SevereMalaria_ study, for histogram, when main="Calcium", overlay="" or any var
image

eg. 2 In _India ICEMR Fever Surveillance_ study,  When x=" Dengue OD cutoff value, Panbio Dengue IgM capture ELISA"," Dengue OD cutoff value, Panbio Dengue Early ELISA".
image

bobular:
Resolves #491

study: India ICEMR Severe P. vivax and falciparum Cohort
back end: dfalke-b

Calcium variable was never showing a histogram in the viz tab.  No request was made.  A type guard was being over-zealous.

Now 'number' variables like Calcium, Bicarbonate, Alkaline phosphatase all fail with 500 at the histogram endpoint.  The requests look well formed. 
Interestingly an 'integer' variable like "fasting blood sugar" with only 2 distinct values works fine.  

Another issue was that NumberVariable and DateVariable should have rangeMin/Max and regardless of data shape.

binWidth needs to be optional because some numeric/date variables have just one distinct value and have no binWidth - this didn't have any TypeScript issues but may have runtime consequences.  An example is "EUPATH_0021230" "Vitamin B12"

### DIFF
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -26,12 +26,15 @@ import {
 import { usePromise } from '../../../hooks/promise';
 import { useDataClient, useStudyMetadata } from '../../../hooks/workspace';
 import { Filter } from '../../../types/filter';
-import { StudyEntity } from '../../../types/study';
+import {
+  DateVariable,
+  NumberVariable,
+  StudyEntity,
+} from '../../../types/study';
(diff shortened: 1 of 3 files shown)


dmfalke (approved):
Looks great!

PR #502, opened 2021-10-03
Fix for API change to user service data table columns

bobular:
This is a basic fix. I was unable to make a backwards-compatible fix, so all old analyses will fail to load.

Do we want a backwards-compatible fix?


ryanrdoherty (approved):
Looks good.

jtlong3rd (approved):
Thanks, @bobular!

PR #504, opened 2021-10-04
allow entity diagram scrolling on overflow

### LINKED ISSUE #428: Add scroll bar to entity diagram section
The entity diagram can be too wide for the page. If users resize the page to be smaller, then you can't even see some of the boxes. Recommend adding a scroll bar as needed. 
Screen Shot 2021-09-08 at 8 50 03 AM

asizemore:
With web-components [PR #218](https://github.com/VEuPathDB/web-components/pull/218), resolves #428 

Goal: Add scroll bar to entity diagram when it is wider than the window.

A first attempt at simply adding `overflow-x: auto` failed, in that the left portion was unreachable with the scroll bar. Thanks to the good people at SO, I learned about and implemented [this approach](https://stackoverflow.com/questions/33454533/cant-scroll-to-top-of-flex-item-that-is-overflowing-container). Required giving a class name to the expanded diagram, hence the need for the `web-components` PR.

Screen Shot 2021-10-04 at 1 50 07 PM

### DIFF
--- a/src/lib/workspace/EDAWorkspace.scss
+++ b/src/lib/workspace/EDAWorkspace.scss
@@ -157,7 +157,11 @@
 
   .Entities {
     display: flex;
-    justify-content: center;
+    overflow-x: auto;
+  }
+
+  .expanded-diagram {
+    margin: auto;
   }
 
   &-Variables {


dmfalke (approved):

dmfalke:
Nice work, @asizemore. As a reminder, you will want to upgrade @veupathdb/components once https://github.com/VEuPathDB/web-components/pull/218 is merged:

```
yarn add @veupathdb/components@latest
```

src/lib/workspace/EDAWorkspace.scss
```diff
@@ -157,7 +157,12 @@
 
   .Entities {
     display: flex;
-    justify-content: center;
+    overflow-x: auto;
```
chowington, line 160: @asizemore About your comment on the `web-components` PR, you need to add the `max-width` to the outermost container that contains only the entity diagram, in this case the `Entities` class element (this kind of stuff is definitely hard to describe, lol). I also wouldn't do the full `100vw`, instead adding a bit of buffer to accommodate the margin on the left side of the page. I'd suggest:

```suggestion
    overflow-x: auto;
    max-width: 97vw;
```

src/lib/workspace/EDAWorkspace.scss
```diff
@@ -157,7 +157,12 @@
...
+  }
+
+  .expanded-diagram {
+    margin: auto;
+    max-width: 100vw;
```
chowington, line 165: And getting rid of the `max-width` here:

```suggestion
```

Issue #381, opened 2021-09-03
Change mouseover text for provider label


SheenaTomko:
Change to "The name of this variable in the data files that were integrated into ClinEpiDB"
<img width="614" alt="Screen Shot 2021-09-03 at 3 07 04 PM" src="https://user-images.githubusercontent.com/54597319/132060647-f65dfbfe-ecb5-41ad-976b-31f843f17135.png">

Issue #290, opened 2021-08-06
Subsetting: Update how original variable names appear


SheenaTomko:
The "Original variable name" under the label for each variable currently looks like:
Original variable name: ["F4A_FUEL_ELEC","F7_FUEL_ELEC"]

Need to update this to 
1) remove the brackets
2) remove the quotation marks
3) add a space after commas separating multiple variables

Should look like:
Original variable name: F4A_FUEL_ELEC, F7_FUEL_ELEC

dmfalke:
It looks like this is being formatted as a JSON string on the backend somehow. Is there a way for this to be formatted as Sheena suggested instead? It can be done on the frontend, but it adds unnecessary complexity.

moontrip:
This work extended to include the treatment of lengthy variable names for multifilter case. See the comment in the link and follow-up discussions after that.

https://github.com/VEuPathDB/web-eda/issues/380#issuecomment-931733308

moontrip:
@danicahelb @SheenaTomko 
I have tested the lengthy variable names for multifilter case as written in my previous comment above. Since there was no definite solution you wanted, a couple of solutions were tested as follows. Please see the screenshots below and let me know which one is better or other approach would be welcome as well.

(For @dmfalke only: I have also tested the component you did let me know, `OverflowingTextCell` at wdk-client, but it did not work for this case: just showed full list of variable names at single line without having Read more link. Didn't carefully checked the code but from my glance I guess that it is because the area to display variable names does not have specific width property so it did not have ellipsis effect)

Certainly, single variable (i.e., not multifilter variable) is shown as the single line without a link to show/hide, e.g.,
![single variable](https://user-images.githubusercontent.com/12802305/135903417-cf4aa9aa-cc40-4a2e-b327-34f852b36343.png)

# showing three variables at single line and then add show more/less link (the link is text-based one)
- before pressing the link
![multifilter-variable-line1](https://user-images.githubusercontent.com/12802305/135902554-0f1153ab-67a2-4a2d-8072-2396105e1706.png)

- after pressing the link
![multifilter-variable-line1-after](https://user-images.githubusercontent.com/12802305/135902573-519c824c-f4f6-4171-b1f3-5bbfe5dbac2e.png)

# show/hide all style: single line per variable and the link is button style (like `Download` or `+ New analysis` links on top of the page)
- before pressing the link
[...]

Issue #289, opened 2021-08-06
Subsetting: Add space between variable definition and histogram stats


SheenaTomko:
For histograms on the subsetting tab, the stats for the histogram appear directly under the original variable label or definition (if there is one). Need to add some vertical space here to differentiate between info about the variable and info about the data it contains

moontrip:
Hi @SheenaTomko I just need your quick review and confirmation whether the new space I have made looks okay for you. Two screenshots are shown in the following, before and after change in order. Thanks!

- Before (current display)
![before-marginTop](https://user-images.githubusercontent.com/12802305/135496382-e90ed99e-1d92-4fe3-971f-b7ef729a915b.png)

- After change
![after-marginTop](https://user-images.githubusercontent.com/12802305/135496483-a52c50ec-b702-4bf5-baae-95098dcaaeb8.png)

moontrip:
Just for a record, Sheena replied me via Slack, saying that it looks great after the change. So I will use that spacing

aaronwlsong:
It looks that this issue exists on qa.website, but not on live site.

For example, in _GEMS1 Case Control study,_ the vars of "weight" and "age".

![image](https://user-images.githubusercontent.com/61077845/136262078-b591d301-1979-4ed0-9186-f3e289ed5830.png)

![image](https://user-images.githubusercontent.com/61077845/136262137-c2a86600-a56a-485b-afd3-b2a779c81b13.png)

SheenaTomko:
Spacing looks great now
<img width="667" alt="Screen Shot 2021-10-07 at 8 37 40 AM" src="https://user-images.githubusercontent.com/54597319/136395425-5e911ad7-04fd-4981-a030-45f4391a6a60.png">

PR #505, opened 2021-10-04
various layout works at Subset tab

moontrip:
This PR addresses following issues:

- #381 
change-question-tooltip-text

- #289 
Sheena approved my mockup - discussed at slack

- #290 
a) change the text style
b) add show more/less link for multifilter case
Discussed with Sheena and she chose one of my mockup, which is included in this PR. See more details at the corresponding issue, #290 with screenshots

### DIFF
--- a/src/lib/workspace/Variable.tsx
+++ b/src/lib/workspace/Variable.tsx
@@ -55,14 +55,16 @@ export function VariableDetails(props: Props) {
             <div key={variable.id}>
               {variable.displayName}:{' '}
               {variable.providerLabel
-                .replace(/[\[\]"]/g, '')
+                .replace(/[[\]"]/g, '')
                 .replace(/[,]/g, ', ')}
               ;&nbsp;
             </div>
           );
         }
+        return null;
       })
-    : variable.providerLabel.replace(/[\[\]"]/g, '').replace(/[,]/g, ', ');
+    : // : variable.providerLabel.replace(/[\[\]"]/g, '').replace(/[,]/g, ', ');
+      variable.providerLabel.replace(/[[\]"]/g, '').replace(/[,]/g, ', ');
 
   // make variables for after the first three
   const providerLabelLeftover = MultiFilterVariable.is(variable)
@@ -75,12 +77,13 @@ export function VariableDetails(props: Props) {
             <div key={variable.id}>
               {variable.displayName}:{' '}
               {variable.providerLabel
-                .replace(/[\[\]"]/g, '')
+                .replace(/[[\]"]/g, '')
                 .replace(/[,]/g, ', ')}
               ;&nbsp;
             </div>
           );
         }
+        return null;
       })
     : '';
 
--- a/src/lib/workspace/Variable.tsx
+++ b/src/lib/workspace/Variable.tsx
@@ -1,3 +1,4 @@
+import React, { useState } from 'react';
 import { HelpIcon } from '@veupathdb/wdk-client/lib/Components';
 import { ErrorBoundary } from '@veupathdb/wdk-client/lib/Controllers';
 import {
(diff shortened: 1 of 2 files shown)


src/lib/workspace/Variable.tsx
```diff
@@ -31,32 +32,88 @@ export function VariableDetails(props: Props) {
...
+            <>
+              <a
+                className="variable-show-more-link"
+                style={{ cursor: 'pointer' }}
+                onClick={() => {
+                  setShowMore(!showMore);
+                }}
+              >
+                {showMoreLink}
+              </a>
```
jtlong3rd, line 110: Instead of rendering an `a`nchor with an inline `cursor: pointer` style, you can use a `button` with a `link` class:

```suggestion
              <button
                className="variable-show-more-link link"
                onClick={() => {
                  setShowMore(!showMore);
                }}
              >
                {showMoreLink}
              </button>
```

("link" is a utility class offered by [@veupathdb/web-common](https://github.com/VEuPathDB/EbrcWebsiteCommon/blob/6a4cd50198a0aaa17c78961716e9975ea6afbef9/Client/src/styles/AllSites.scss#L93) which can be used to style a button as though it were a link.)

src/lib/workspace/Variable.tsx
```diff
@@ -31,32 +32,88 @@ export function VariableDetails(props: Props) {
...
+          &nbsp; {threeProviderLabel} &nbsp;
           <HelpIcon>
-            The name for this variable as provided with the original study's
-            data set. The VEuPathDB team curates variable names and places
-            variables into an ontology framework.
+            The name of this variable in the data files that were integrated
+            into ClinEpiDB
           </HelpIcon>
+          &nbsp;&nbsp;
+          {MultiFilterVariable.is(variable) ? (
```
jtlong3rd, line 100: I'd add some logic to only display the "show more" if there are more than 3 original variable names. (I'm thinking about things like India ICEMR Cohort's "resident status.")

jtlong3rd (commented):
I have just two small inline suggestions, and then this should be good to merge!

src/lib/workspace/Variable.tsx
moontrip: @jtlong3rd thank you for your review and suggestions 👍 Actually I was also thinking of the use of button but just did it with <a> 😄 Will change it to <button> as suggested

src/lib/workspace/Variable.tsx
moontrip: yes I also thought of that edge case but didn't make it. I totally agree that it needs to be checked just in case!

moontrip:
@jtlong3rd I have addressed your feedbacks. For the second one, I made a variable to check the array length of the multifilter variable and use it as additional condition to display `Show more` link. An example screenshot for the case of `resident status` you suggested, which should not have the expansion link. 

multifilter-variable-two-variables

src/lib/workspace/Variable.tsx
```diff
@@ -31,32 +32,95 @@ export function VariableDetails(props: Props) {
...
+        groupBy(entity.variables, (variable) => variable.parentId)
+      ).length
+    : 1;
+
+  // show the first three if multifilter variable
+  const threeProviderLabel = MultiFilterVariable.is(variable)
+    ? findMultifilterVariableLeaves(
+        variable,
+        groupBy(entity.variables, (variable) => variable.parentId)
+      ).map((variable, i) => {
```
jtlong3rd, line 52: The linter points out: `Unnecessary escape character: \[`. Could you fix? (You can probably just replace `\]` with `]`.)

The linter also points out `Array.prototype.map() expects a value to be returned at the end of arrow function`. To fix that, I would return either `null` or  `<></>` when `i >= 3`.

src/lib/workspace/Variable.tsx
```diff
@@ -31,32 +32,95 @@ export function VariableDetails(props: Props) {
...
+        }
+      })
+    : variable.providerLabel.replace(/[\[\]"]/g, '').replace(/[,]/g, ', ');
+
+  // make variables for after the first three
+  const providerLabelLeftover = MultiFilterVariable.is(variable)
+    ? findMultifilterVariableLeaves(
+        variable,
+        groupBy(entity.variables, (variable) => variable.parentId)
+      ).map((variable, i) => {
```
jtlong3rd, line 72: The linter also throws similar warnings here.

jtlong3rd (commented):
Two more inline suggestions, and then I'll leave you alone - honest! 😁

src/lib/workspace/Variable.tsx
moontrip: @jtlong3rd you are always welcome to bother me 😄 
Anyway, actually I knew that the first warning but forgot to address it. Yes it works by simply removing the backslash. As for the second, I added `return null;` for those cases

jtlong3rd (approved):

src/lib/workspace/Variable.tsx
dmfalke: Wouldn't it make more sense to only `map` the first three items in the array? It's fewer iterations and clearer, imo.

```typescript
findMultifilterVariableLeaves(
  variable,
  groupBy(entity.variables, (variable) => variable.parentId)
)
.slice(3)
.map((variable, i) => {
```

dmfalke (approved):
I left a follow-up comment to @jtlong3rd's inline comment. Otherwise, looks great.

src/lib/workspace/Variable.tsx
jtlong3rd: I agree with @dmfalke here - you could `.slice(0, 3)` to map the first three elements.

src/lib/workspace/Variable.tsx
jtlong3rd: Similarly, you could `.slice(3)` to map all but the first three elements.

src/lib/workspace/Variable.tsx
dmfalke: Thank you for the clarification @jtlong3rd~

jtlong3rd (approved):

Issue #424, opened 2021-09-09
Issues in mouseover info for starred variables filter


nkittur-uga:
1. The mouseover text for the starred variable filter (grey below) often persists long after the cursor is moved away from the starred variable toggle switch.

2. Please remove the period at the end of this text. It should simply read, "To use this filter, star at least one variable below"

<img width="1164" alt="Screen Shot 2021-09-09 at 2 27 14 PM" src="https://user-images.githubusercontent.com/67710160/132746633-0e8f9f58-3cb0-4db9-af44-05b01ea902f0.png">

asizemore:
Closed with #520

PR #520, opened 2021-10-05
Fix star toggle hover

asizemore:
Resolves issue #424 

The overly-persistent tooltip was an issue only when the button is disabled. Found a discussion of this issue [here](https://github.com/bsidelinger912/react-tooltip-lite/issues/46).

Other solutions included modifying the div structure, but this was the most straightforward solution.

Additionally updated the toggle text as suggested in the issue's description. Removed all periods from this text for consistency.


jtlong3rd (approved):
Great catch! Based on a quick read of the [MUI docs](https://mui.com/components/tooltips/#disabled-elements), I guess the general solution is either to:

1. When using native `button` elements, add `pointer-events: none` styling, as you did here
2. Use MUI `Button`s instead of native `button` elements

asizemore:
Thanks @jtlong3rd! Great to know!

Issue #387, opened 2021-08-25
In numeric data, use commas only for values >9999.


nkittur-uga:
Year (eg. 1934) appears as 1,934 due to the placement of comma denoting the thousands. This could be prevented if commas are used only for numbers 10,000 and beyond. 

<img width="290" alt="Screen Shot 2021-08-25 at 2 42 25 PM" src="https://user-images.githubusercontent.com/67710160/130847516-1e2c9acb-c357-43c7-9696-21d795f358b9.png">

PR #521, opened 2021-10-05
number comma over 9999, resolve some lint warning

moontrip:
This PR addresses the link issues, having number comma over 9999, not typical thousand level: essentially avoid comma for year data. Also some lint warnings at `HistogramFilter` component are addressed.

Here are screenshots:
- birth year that was demoed at [the linked issue](https://github.com/VEuPathDB/web-eda/issues/387#issue-990107914)
birth-year

- a number more than thousand: max stats value is 50000, so it is displayed as 5,0000.
more-than-9999


src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -769,9 +770,15 @@ function formatStatValue(
...
-        maximumFractionDigits: 4,
-      });
+    : // need to check integer not to add additional decimal points
+    Number(value) % 1 === 0
+    ? changeNumberWithCommas(Number(value))
+    : changeNumberWithCommas(Number(value).toFixed(4));
+}
+
+// a function to add comma for a number > 9999
+function changeNumberWithCommas(x: number | string) {
```
bobular, line 780: If the true purpose is to avoid years being rendered as 2,021, it might be better to continue to use `toLocaleString()` with the `useGrouping` option set to false for numbers in the range (1900-2100) ?

moontrip: @bobular thank you for your suggestion 👍 Will think about this

jtlong3rd (approved):
TIL about `useGrouping`!

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -770,15 +770,15 @@ function formatStatValue(
...
-    : changeNumberWithCommas(Number(value).toFixed(4));
-}
-
-// a function to add comma for a number > 9999
-function changeNumberWithCommas(x: number | string) {
-  return x.toString().replace(/\B(?<!\.\d*)(?=(\d{4})+(?!\d))/g, ',');
+    : Number(value) >= 1900 && Number(value) <= 2100
+    ? Number(value).toLocaleString(undefined, {
+        maximumFractionDigits: 4,
+        useGrouping: false,
```
bobular, line 776: Would this be the same and more concise? (and still readable!)
```
useGrouping: !(Number(value) >= 1900 && Number(value) <= 2100),
```

PR #522, opened 2021-10-05
Add bin start/end inclusivity/exclusivity props to fix missing bin histogram problem

### LINKED ISSUE #515: Subsetting: WASHb Kenya histogram for "Time to water source" is not showing all of the data
In the WASHb Kenya study, the histogram on the subset tab for 'Time to water source' has an x-axis that extends from 600, but the data seems to stop ~200 (even after changing to log scale & zooming in). The range indicator states that the data ranges from 0-600. 

image

Using the visualization histogram, I can see the data at ~600.
image

Live site data also extends to ~600.
image

bobular:
Resolves #515

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -574,6 +574,8 @@ function HistogramPlotWithControls({
       <Histogram
         {...histogramProps}
         data={data}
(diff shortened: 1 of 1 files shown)


dmfalke:
I find this confusing. The range in the input (0-600) is different from the range in the plot (0-625):

image

PR #528, opened 2021-10-06
Analyses Table Updates

### LINKED ISSUE #448: General: Public analyses table format
On the EDA site, the Public Analyses link should take you to a stand-alone “Public analyses” table

This is in place of having the Public analyses table live within a tab on “My search strategies” (this is where the current Public strategies table lives in the WDK site)

Structure of the Public analyses table will look similar to the My analysis table currently on the EDA site, and will have the following columns in the following order:  
1. **Study:** Study label
2. **Analysis:** user-provided name given to the analysis
3. **Description:** user-provided description of the analysis
4. **Author:** who created the analysis
5. **Organization:** affiliated institution for the author of the analysis
6. **Created:** when the analysis was created
7. **Modified:** when the analysis was modified

### LINKED ISSUE #452: Updates to My analyses table
Change the table so the first column is the study and the second column is the analysis. A similar request is being made for the public analyses table in a separate ticket. 

Screen Shot 2021-09-17 at 2 17 40 PM

### LINKED ISSUE #513: General: append import time stamp to the analysis name when public analysis is opened by user
As discussed in 10/1 UX meeting:

When a user opens up a public analysis into their analyses workspace, it adds a copy of the public analysis into their My analyses table. We need to update the name of the copied public analysis 

For phase 1, append “imported from author name on date-time stamp”

The updated analysis name should appear on the top of the analysis page and also in the My analyses table

### LINKED ISSUE #527: Verbiage change: "Select a visualization" -> "New visualization"
image

jtlong3rd:
Resolves #448, resolves #452, resolves #513, resolves #527.

### DIFF
--- a/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -83,7 +83,7 @@ function ConfiguredVisualizations(props: Props) {
     <Grid>
       <Link replace to={`${url}/new`} className={cx('-NewVisualization')}>
         <i className="fa fa-plus"></i>
-        Select a visualization
+        New visualization
       </Link>
       {computation.visualizations
         .map((viz) => {
--- a/src/lib/core/hooks/publicAnalyses.ts
+++ b/src/lib/core/hooks/publicAnalyses.ts
@@ -1,8 +1,11 @@
-import { useCallback } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 
 import { PromiseHookState, usePromise } from './promise';
 
-import { AnalysisClient } from '../api/analysis-api';
+import {
+  AnalysisClient,
+  SingleAnalysisPatchRequest,
+} from '../api/analysis-api';
 import { PublicAnalysisSummary } from '../types/analysis';
 
 export function usePublicAnalysisList(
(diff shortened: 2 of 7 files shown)


dmfalke (approved):
I skimmed the code, and nothing jumped out.

jtlong3rd:
> Hi Jamie - I'm having trouble getting past first base in testing this If I click on on an analysis link in the public analyses table, it gets stuck on this: image
> 
> The URL bar has this at the time of being stuck: `https://localhost:3000/eda/DS_4902d9b7ec/699b5eafdbbc3e287eace90a4ef53fd7/import/1926010?description=First%20VEuPathDB%20public%20analysis%0A%0A(Imported%20from%20VEuPathDB%20BRC%20%5BVEuPathDB%5D%20on%2007%20Oct%202021%2012%3A37%3A03%20GMT.)`
> 
> I'm using the dfalke-b back end

I just added another Public Analysis - this one should be importable!

The issue you were experiencing arises because:

1. "copy analysis"  is 500ing for all analyses in the existing Public Analyses table. (I haven't investigated, but I suspect these 500s might be due to last week's API upgrade.)
2. The client should be catching this 500 and displaying the error associated message.

I'll address issue 2 in the PR.

(@dmfalke - issue 2 seems to be a generic issue with our [`Task` utility](https://github.com/VEuPathDB/WDKClient/blob/master/Client/src/Utils/Task.ts) - a quick read of the code suggests that if you don't pass an `onRejected` callback to the `run` method, then any errors the `Task` throws will be "swallowed." I'm guessing the solution is change the default `onRejected` callback from `noop` to a function which always throws the argument it receives.)

bobular (approved):
I had a good browse of the code and tested it both as a civilian and as the ExampleAnalysesAuthor - all seems to look and work fine :-)

Issue #490, opened 2021-09-29
When logging in, transfer analyses from the guest session


jtlong3rd:
(Currently, if you create some analyses as a guest user and then log in, your work will be lost.)

PR #535, opened 2021-10-07
Lazily transfer guest analyses

jtlong3rd:
Resolves https://github.com/VEuPathDB/web-eda/issues/490.


dmfalke (approved):
This looks great! I haven't tested it because I'm in the middle of some other changes, but will do so once this is merged.

Issue #449, opened 2021-09-19
Subsetting: results table link and data table popup styling


danicahelb:
**Link on subsetting tab**
- Use the table overlaid with the download icon, and label it: “View and download”
- include mouseover text with more information: “View and download current subset of Communities”

**Results table view**
- Even though we have decided to keep the text in the label simple, we want to make it VERY obvious that the link brings you to a SUBSET of data in a particular ENTITY SPACE
- Mike is putting the results table into a popup. would we prefer it to go to a separate page? solicit feedback after demo...

asizemore:
Closed with #537

SheenaTomko:
Qc'd and looks fixed, but now we will move to having a separate downloads tab so moot anyway

PR #537, opened 2021-10-08
General button styling updates

### LINKED ISSUE #450: General: styling of all icon/label links (buttons?) should match

asizemore:
Resolves #450 and one half of #449. 

This PR addresses the styling of the EDA header and subsetting buttons. Specifically, those buttons in the header are styled with their own class. An alternative route could be to use [styled](https://mui.com/customization/how-to-customize/#2-reusable-style-overrides) in the future if we find we are doing this more frequently. The main reason the styling of these three buttons differs is because some, but not all, are using a Link component.

Finally, the View and Download current subset button was updated with new language as described in #449.

Note, as mentioned in #450 there will be a separate issue for modifying the download icon, because it will be a more involved process and should be it's own decision, as it affects more sites than just clinepi.

Screen Shot 2021-10-08 at 11 05 54 AM

### DIFF
--- a/src/lib/workspace/EDAWorkspace.scss
+++ b/src/lib/workspace/EDAWorkspace.scss
@@ -88,9 +88,11 @@
     gap: 1em;
     margin-right: auto;
     margin-left: 2em;
-    a {
-      font-size: 0.75em;
-    }
+    align-items: center;
+  }
+
+  .Linkouts-buttons {
+    font-size: 0.75em;
   }
 
   &-Analysis {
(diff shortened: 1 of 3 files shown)


eikonomega (approved):
I think this is all good. Thank you @asizemore!
As an aside, there is now a very flexible `SwissArmyButton` component in the core-components library that we can use (hopefully - for all buttons!) for future updates.

Issue #185, opened 2021-06-22
allowing date string for scatter plot and its viz


moontrip:
currently, scatter plot viz does not work properly with date string, e.g., birth date, enrollment date, etc. This needs to be addressed, especially for the part to make plotly friendly data using backend response.

PR #548, opened 2021-10-11
addressing y-ticks for logscale at barplot

moontrip:
This will address [#185 at web-components](https://github.com/VEuPathDB/web-components/issues/185) and requires [the corresponding PR at web-components](https://github.com/VEuPathDB/web-components/pull/225). 

Here are some examples for both count and proportion cases: used GEMS1A study
case or control participant1

case or control participant2

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@mui/styles": "^5.0.1",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.8.1",
+    "@veupathdb/components": "^0.8.2",
     "@veupathdb/wdk-client": "^0.4.4",
     "@veupathdb/web-common": "^0.1.9",
     "debounce-promise": "^3.1.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3049,10 +3049,10 @@
     react-transition-group "^4.4.1"
     shape2geohash "^1.2.5"
 
-"@veupathdb/components@^0.8.1":
-  version "0.8.1"
-  resolved "https://registry.yarnpkg.com/@veupathdb/components/-/components-0.8.1.tgz#bd4b87072453ee8ecd922b7426b7b46eb9dcf898"
-  integrity sha512-8+Y7grI0hmTk8ejxRCXA6p0ZmSHeq5JYi1noIUEyVgq2CdkrhRaEKemoa4UFd3wIgPENzSoOUtDK6wsmbm8rZw==
+"@veupathdb/components@^0.8.2":
+  version "0.8.2"
+  resolved "https://registry.yarnpkg.com/@veupathdb/components/-/components-0.8.2.tgz#4df63c02938aecae0baa1450cbf944669f5ad597"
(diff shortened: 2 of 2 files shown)


src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -222,6 +222,31 @@ function BarplotViz(props: VisualizationProps) {
...
+      ? Math.max.apply(
+          Math,
+          data?.value?.series.map((o) => {
+            let localMax = Math.max(...o.value);
+            console.log('local max =', localMax);
+            return localMax;
+          })
+        )
+      : undefined;
+  }, [data, data.value, variable, overlayVariable]);
```
bobular, line 237: Don't need both `data` and `data.value` ?

bobular: Elsewhere we've used flatMap and lodash Max, I think?

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -222,6 +222,31 @@ function BarplotViz(props: VisualizationProps) {
...
+
+  // set min/max
+  const dependentAxisRange =
+    defaultDependentMaxValue != null
+      ? {
+          // set min as 0 (count) or 0.001 (proportion)
+          min: vizConfig.valueSpec === 'count' ? 0 : 0.001,
+          // add 5 % margin
+          max: defaultDependentMaxValue * 1.05,
```
bobular, line 246: nice

src/lib/core/utils/default-dependent-axis-range.ts
```diff
@@ -4,14 +4,14 @@ import { NumberOrDateRange } from '@veupathdb/components/lib/types/general';
...
   plotName: string,
-  // yMinMaxRange: NumberOrDateRange | undefined,
   yMinMaxRange:
     | { min: number | string | undefined; max: number | string | undefined }
     | undefined
 ): NumberOrDateRange | undefined {
   // make universal range variable
-  if (variable != null && variable.dataShape === 'continuous') {
-    if (variable.type === 'number') {
+  if (variable != null && plotName === 'scatterplot') {
```
bobular, line 12: Is this a fix for another ticket?

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
moontrip: @bobular thank you for your review! As for the scatterplot fix, honestly I don't know if it is necessary so your input would be helpful. Here is the situation I have faced for that. When I checked one of my favorite test variables (haha) for scatterplot viz, Sleeping rooms in dwelling (X) & Persons sleeping in dwelling count, I realized that the type of Persons sleeping in dwelling count is now classified as integer (previously number). default-dependent-axis-range.ts only checks number or date so if integer type comes in, no (dependent axis) range is set. I thought that such an integer type should also have the (fixed) range like number. The change is related to this. Do you think this is reasonable to change ?

As for data and data.value, yeah actually one of them works. Perhaps data will be better: will change it

PR #551, opened 2021-10-11
add a prefix at a table heading and its center alignment

moontrip:
This PR needs to address two things

- [x] add 'All' prefix to table heading and align center (`TableFilter` component)
- [x] put the pink legend for "Subset of Households" to the left of the grey legend for "All Households" (`HistogramFilter` component)


moontrip:
The first one is almost done: need a confirmation from Danica. It also relies on the [corresponding PR at WDKClient](https://github.com/VEuPathDB/WDKClient/pull/158)

moontrip:
The change at WDKClient is approved and its version is also updated to 0.4.5 by Dave. Accordingly, this PR also includes such changes: the first task is done.

package.json
```diff
@@ -7,7 +7,7 @@
     "@mui/styles": "^5.0.1",
     "@types/debounce-promise": "^3.1.3",
     "@veupathdb/components": "^0.8.3",
-    "@veupathdb/wdk-client": "^0.4.4",
+    "@veupathdb/wdk-client": "^0.4.5",
```
bobular, line 10: Since the wdk-client is already merged, I'll comment here.

The centered text is not quite right.

I think what's needed is in FilterParam.css, the `right` below needs to be `center`
 
```
.filter-param .membership-filter .MesaComponent th:nth-child(3) .HeadingCell-Content-Label, .filter-param .membership-filter .MesaComponent th:nth-child(4) .HeadingCell-Content-Label {
	text-align: right;
}
```

bobular (changes requested):
Looks good apart from the centering CSS issue.

I was worried that reversed trace order might put the pink bars behind the grey ones, but it does not. :-)

package.json
moontrip: @bobular Thank you for finding the exceptions! 👍 yes you are right it is the place to change. I made a [PR at WDKClient](https://github.com/VEuPathDB/WDKClient/pull/159) after changing it. Here is a screenshot after the change:

table-heading-centering

PR #553, opened 2021-10-12
Update useAnalysis to support both new and saved analyses

### LINKED ISSUE #497: Last viewed variable is lost when transitioning from new → saved analysis
To reproduce:
1. Create a new analysis
2. Select a variable other than the default
3. Go to **Visualize** and create a visualization. Any will do. This will cause the new analysis to be saved.
4. Go back to **Browse and subset** and see that the variable selected in (2) is no longer selected.

The bit of state keeping the last viewed variable is stored as component state (see src/lib/workspace/AnalysisPanel.tsx), which is lost when the component is unmounted. It seems that going from a new analysis to a saved analysis is causing the component to be unmounted. This state likely needs to be moved to a parent component to prevent this.

### LINKED ISSUE #550: Copy analysis button is creating the "Copy of My Analysis" but not switching to it

### LINKED ISSUE #552: Deleted analyses should be from removed from the analysisCache
Easiest way to reproduce:

1. Create and save an analysis
2. Click on "My Analyses"
3. In the "My Analyses" table, delete the analysis from step 1
4. Hit the browser's back button

jtlong3rd:
Resolves #497, fixes #550, and fixes #552.

At a high level, this PR updates the `useAnalysis` hook to manage the state of both new and saved analyses. This allows us to use a common `EDAWorkspace` component to render new and saved analyses, and thus prevent the contained `AnalysisPanel` component from being unmounted when transitioning from a new to a saved analysis.

Sometime soon, we should revisit the UX around the transition from new -> saved analysis. Currently, most `AnalsyisState` updates are blocked while the analysis is being created.* We might consider:

1. Displaying some sort of "Creating/saving analysis" overlay during the analysis creation
2. Allowing `AnalysisState` updates during the analysis creation, and then, once the analysis is created, `PATCH`ing the analysis with changes that were made mid-creation

(* @bobular - this is the source of that transient "visualization not found" message you've been observing. Creating a visualization in a "new" analysis doesn't actually change the `AnalysisState`, just triggers the creation of a saved analysis which contains that visualization.)

### DIFF
--- a/src/lib/workspace/EDAWorkspaceHeading.tsx
+++ b/src/lib/workspace/EDAWorkspaceHeading.tsx
@@ -79,11 +79,11 @@ export function EDAWorkspaceHeading({
                 color="primary"
                 startIcon={<Icon className="fa fa-plus fa-fw" />}
                 onClick={
-                  /** If (1) there is no analysis, (2) we're in an unsaved new
-                   * analysis (here `analysis` is still undefined in this case),
-                   * or (3) we're in a renamed analysis, just go straight to the
-                   * new analysis. Otherwise, show the renaming dialog. */
-                  analysis && analysis.displayName === DEFAULT_ANALYSIS_NAME
+                  /** If we're in an unnamed saved analysis, show the renaming dialog.
+                   *  Otherwise, just go straight to the new analysis.
+                   */
+                  isSavedAnalysis(analysis) &&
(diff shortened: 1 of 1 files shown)


bobular (approved):
I haven't gone through all the code, but I've played with this branch and everything still seems to work - and without the "Visualization not found" message too.  Seems snappier.

PR #568, opened 2021-10-13
temporarily remove bulk download button

asizemore:
As discussed on slack, temporarily removes the study download button.

Screen Shot 2021-10-13 at 2 17 45 PM


moontrip (approved):
Thank you for handing this as well ! 👍

PR #570, opened 2021-10-13
Histogram viz yaxis range

moontrip:
This PR resolves disappearance of y-axis tick labels for certain variables at histogram viz. I have made screenshots for before and after setting y-axis range: note that I captured screenshots before merging main branch so white gap between bar is still showing 😃 

### - before this work: both count and proportion with log scale (GEMS1 Case Control data)
histogramViz-noRange-count-log

histogramViz-noRange-proportion-log

### - after applying this work: both count and proportion with log scale (GEMS1 Case Control data
histogramViz-Range-count-log

histogramViz-Range-proportion-log

### DIFF
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -11,12 +11,15 @@ import {
   TimeDelta,
 } from '@veupathdb/components/lib/types/general';
 import { isTimeDelta } from '@veupathdb/components/lib/types/guards';
-import { HistogramData } from '@veupathdb/components/lib/types/plots';
+import {
+  HistogramData,
+  HistogramDataSeries,
+} from '@veupathdb/components/lib/types/plots';
 import { preorder } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
 import { getOrElse } from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/function';
 import * as t from 'io-ts';
-import { isEqual } from 'lodash';
+import { isEqual, sortBy, uniq, flatMap, max } from 'lodash';
 import React, { useCallback, useEffect, useMemo, useRef } from 'react';
 import {
   DataClient,
@@ -275,6 +278,24 @@ function HistogramViz(props: VisualizationProps) {
     [xAxisVariable]
   );
 
+  // find max of stacked array, especially with overlayVariable
+  const defaultDependentAxisMaxValue = useMemo(
+    () =>
+      data.value && data.value.series.length > 0
+        ? findMaxOfStackedArray(data.value.series)
+        : undefined,
+    [data, xAxisVariable, overlayVariable]
+  );
+
+  // set default dependent axis range for better displaying tick labels in log-scale
+  const defaultDependentAxisRange =
+    defaultDependentAxisMaxValue != null
+      ? {
+          min: vizConfig.valueSpec === 'count' ? 0 : 0.001,
+          max: defaultDependentAxisMaxValue * 1.05,
+        }
+      : undefined;
+
   return (
     <div style={{ display: 'flex', flexDirection: 'column' }}>
       <div style={{ display: 'flex', alignItems: 'center', zIndex: 1 }}>
@@ -337,6 +358,8 @@ function HistogramViz(props: VisualizationProps) {
         independentAxisLabel={axisLabelWithUnit(xAxisVariable) ?? 'Main'}
         // variable's metadata-based independent axis range
         independentAxisRange={defaultIndependentRange}
+        // add dependent axis range for better displaying tick labels in log-scale
+        dependentAxisRange={defaultDependentAxisRange}
         interactive
         showSpinner={data.pending}
         filters={filters}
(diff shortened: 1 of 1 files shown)


src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -275,6 +278,24 @@ function HistogramViz(props: VisualizationProps) {
     [xAxisVariable]
   );
 
+  // find max of stacked array, especially with overlayVariable
+  const defaultDependentAxisMaxValue = useMemo(
+    () =>
+      data.value && data.value.series.length > 0
+        ? findMaxOfStackedArray(data.value.series)
+        : undefined,
+    [data, xAxisVariable, overlayVariable]
```
bobular, line 287: Do we need `xAxisVariable, overlayVariable` in the dependencies?

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -639,3 +662,62 @@ function reorderData(
...
+      for (let j = 0; j < uniqueBinStart.length; j++) {
+        // check if binStart exists compared to unique binStart array
+        if (dataArray[i].binStart.includes(uniqueBinStart[j])) {
+          let indexCount = dataArray[i].binStart.indexOf(uniqueBinStart[j]);
+          // make tempArray for data.count
+          tempArray.push(dataArray[i].count[indexCount]);
+        } else {
+          // add filler count, 0 for non-existence case
+          tempArray.push(0);
+        }
```
jtlong3rd, line 716: Slight simplification using `findIndex`:

```suggestion
        // check if binStart exists compared to unique binStart array
        const indexCount = dataArray[i].binStart.findIndex(bStart => bStart ===  uniqueBinStart[j]);
        
        if (indexCount >= 0) {
          // make tempArray for data.count
          tempArray.push(dataArray[i].count[indexCount]);
        } else {
          // add filler count, 0 for non-existence case
          tempArray.push(0);
        }
```

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -639,3 +662,62 @@ function reorderData(
...
+    uniq(dataArray.flatMap((data) => data.binStart))
+  );
+
+  // make array of count array (2D)
+  let totalCountArray = [];
+
+  // construct 2D totalCountArray
+  // [# of stratification] rows and [# of unique binStart] at each low
+  if (uniqueBinStart != null) {
+    for (let i = 0; i < dataArray.length; i++) {
```
jtlong3rd, line 703: A [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...ofl) loop might improve readability.

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -639,3 +662,62 @@ function reorderData(
...
+      // temporary array to store each row
+      let tempArray = [];
+      // looping through each data array
+      for (let j = 0; j < uniqueBinStart.length; j++) {
```
jtlong3rd, line 707: A [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...ofl) loop might improve readability.

moontrip:
@bobular Thank you for your reivew! 👍 I have addressed two issues you found. Removed unnecessary dependencies (for both histogram and barplot viz) and set a condition to check vizConfig.dependentAxisLogScale to set range min of 0 (non-log) or 0.001 (log) for proportion

jtlong3rd:
> @jtlong3rd Thank you for your review and suggestions! 👍 May I ask one question regarding `for-of` for clarification ? so for the example with the first for loop, i.e., `for (let i = 0; i < dataArray.length; i++)`
> 
> Did you have in mind using like this ? `for (const i of dataArray.keys())`

Almost! You can actually iterate over the elements of the dataArray like this: `for (const dataArrayElement of dataArray) {`. You can even do this: `for (const { binStart, count } of dataArray) {`.

bobular:
@moontrip  I'm not usually a fan of reduce (right, Jamie?!), but I think it's what's needed for the max-finder function

This returns the exact same result as your function, on one plot I tested with.
```
function findMaxOfStackedArray(
  data: HistogramDataSeries[]
) {
  // calculate the sum of all the counts from bins with the same label
  const sumsByLabel = data
    .flatMap( // make an array of [ [ label, count ], [ label, count ], ... ] from all series
      (series) => zip(series.bins.map((bin) => bin.binLabel),
		      series.bins.map((bin) => bin.count)))
      // then do a sum of counts per label
    .reduce<Record<string, number>>(
      (map, [ label, count ]) => {
	if (label != null && count != null) { // zip can potentially return nulls
	  if (map[label] == null)
	    map[label] = 0;
	  map[label] = map[label] + count;
	}
	return map;
      },
      // empty map for reduce to start with
      {}
    );

  return max(Object.values(sumsByLabel));
}
```

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -11,12 +11,15 @@ import {
...
+import {
+  HistogramData,
+  HistogramDataSeries,
+} from '@veupathdb/components/lib/types/plots';
 import { preorder } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
 import { getOrElse } from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/function';
 import * as t from 'io-ts';
-import { isEqual } from 'lodash';
+import { isEqual, sortBy, uniq, flatMap, max } from 'lodash';
```
bobular, line 22: You'd need `zip` here, but no need for `flatMap` it's standard JS the way we use it.

jtlong3rd:
> @moontrip I'm not usually a fan of reduce (right, Jamie?!), but I think it's what's needed for the max-finder function
> 
> This returns the exact same result as your function, on one plot I tested with.
> 
> ```
> function findMaxOfStackedArray(
>   data: HistogramDataSeries[]
> ) {
>   // calculate the sum of all the counts from bins with the same label
>   const sumsByLabel = data
>     .flatMap( // make an array of [ [ label, count ], [ label, count ], ... ] from all series
>       (series) => zip(series.bins.map((bin) => bin.binLabel),
> 		      series.bins.map((bin) => bin.count)))
>       // then do a sum of counts per label
>     .reduce<Record<string, number>>(
>       (map, [ label, count ]) => {
> 	if (label != null && count != null) { // zip can potentially return nulls
> 	  if (map[label] == null)
> 	    map[label] = 0;
> 	  map[label] = map[label] + count;
> 	}
> 	return map;
>       },
>       // empty map for reduce to start with
>       {}
>     );
> 
>   return max(Object.values(sumsByLabel));
> }
> ```

Slick! How about `(series) => series.bins.map(bin => [bin.binLabel, bin.count])` instead of `(series) => zip(series.bins.map((bin) => bin.binLabel), series.bins.map((bin) => bin.count)))`?

jtlong3rd (approved):
🎉

bobular:
Cool - I think we're over the line.  :tada:  :beers:  :fireworks:

Issue #576, opened 2021-10-14
Circular dependency with the @veupathdb/web-common npm package


dmfalke:
There is a circular dependency between `web-common` and `web-eda`. `web-eda` depends on four things from `web-common`:

1. website bootstrap code used for dev
2. Icons
    1. move to its own package
3. data restriction
    1. move to its own package
4. data-api
    1. move to its own package

If we move 2-4 out of web-common, we can make web-common a devDependency, which will resolve the issues with circular dependencies.

### Tasks

- [ ] Icons
- [ ] Data restriction
- [ ] Data API

dmfalke:
This is related to #484

PR #603, opened 2021-10-18
Use @veupathdb/http-utils

dmfalke:
The new package `@veupathdb/http-utils` provides the base class `FetchClient` used by the client api modules in this package. The base class used to live in `@veupathdb/web-common`, but it was contributing to a circular dependency between that package and this package.

Partially addresses #576


dmfalke:
I'm anxious to merge this, so here goes!

Issue #571, opened 2021-10-08
Original variable names styling with multifilter variables


danicahelb:
In phase 1 we decided to implement Show more/less for original variable names for multifilter variables.  we have decided to show the first three variables if the number of variables are larger than 3, and then the show more buttons

![image](https://user-images.githubusercontent.com/25450900/136610729-2d34ba69-da48-4b96-96e3-c2efb3f8a286.png)

We'd like these styling changes:

- [ ] Once all original variable names are expanded, the "Show less" button appear in the middle of the list. Please move it to the end of the list (so in the example, after F4A_FUEL_OTHER)

- [ ] a semi-colon is showing up after the last item in the list, please remove this

sweverschulman:
The variable expansion button is at the end of the list with commas betwee
<img width="585" alt="Screen Shot 2021-10-29 at 2 41 49 PM" src="https://user-images.githubusercontent.com/62298921/139486219-f9b90fce-be0e-4570-823a-78a2286efcfa.png">
n
<img width="587" alt="Screen Shot 2021-10-29 at 2 41 55 PM" src="https://user-images.githubusercontent.com/62298921/139486227-c1016a4b-cbcb-4b0b-b19c-bd4b1b386b97.png">
.

PR #604, opened 2021-10-18
change styling of variable names at Subset

moontrip:
This will address two issues at #571. Screenshots are attached as follows.

show-more1

show-less1


jtlong3rd (approved):

moontrip:
@jtlong3rd thank you for your review! 👍

PR #628, opened 2021-10-21
Include Auth-Key header in all EDA related service requests.

### LINKED ISSUE #619: Add auth key to all eda service requests
This is needed to support backend data restrictions.

dmfalke:
- Introduce FetchClientWithCredentials base class
- Refactor AnalysisClient to leverage base class
- Add new hooks for all EDA client classes

fixes #619


jtlong3rd (approved):
This is great!

I'm wondering if `FetchClientWithCredentials` (maybe sans the `getClient` static method) should live in https://github.com/VEuPathDB/web-http-utils? MultiBLAST and the study access dashboard would both put it to good use.

Issue #482, opened 2021-09-21
Visualizations: 2x2 RxC mosaic tabs that switch between plot and table have styling issues


bobular:
On some browsers (to be further diagnosed) the selected tab is "whited out" like below, after changing tabs.  I found that it is possible to avoid this if mousing out of the tab very quickly after clicking (near the edge helps). YMMV.

Below is a screenshot from Danica's screen, but I get it too on FF Linux.

![image](https://user-images.githubusercontent.com/308639/134196224-cae84303-3724-48fd-b31e-5616d0921120.png)

danicahelb:
I was using Chrome, if that helps

danicahelb:
Hey @chowington there is still something strange happening. If I click back and forth between mosaic and table tabs, sometimes it looks like image 1 and sometimes it looks like image 2. This is still a huge improvement over how it previously looked, but it would be good to have consistency:

![image](https://user-images.githubusercontent.com/25450900/137350297-75c6074d-4bc9-4107-aa39-6fb917bda150.png)

![image](https://user-images.githubusercontent.com/25450900/137350340-d5d9cf17-66ee-4fce-b98a-25c76b69cfe0.png)

danicahelb:
@chowington this can be fixed for b55

danicahelb:
looks perfect, thanks!

PR #629, opened 2021-10-22
Change tabs styling

chowington:
(Really just using new tabs component altogether)

Fixes https://github.com/VEuPathDB/web-eda/issues/482


bobular:
> Note that the grey background-color here is no longer focus highlighting but active-tab highlighting. There is no focus highlighting in these tabs currently. Is this a problem for accessibility? I don't know what our stance is on that as a site.

Good question.  I've added this as an item to discuss on Wednesday.

Issue #172, opened 2021-06-10
Visualizations: Some Plotly state is not being persisted (such as hiding traces via the legend)


d-callan:
when ive minimized a visualization (i was looking at scatter) and gone back to where i can add another, its reasonable a user might be looking to have the configurations they set when expanded/full screen to still apply.

bobular:
So if a user chooses log scale y axis in histogram, the next freshly made histogram should have log scale on by default?
Some settings (like log y scale) could be sticky, but others, like bin width, should not.  

Currently, a user can copy/clone a visualisation if they wanted to preserve some settings.  I tried this with histogram.  You can set a custom log-scale and bin width, then clone the viz, then **change the variable** in the new viz, the log scale persists, but the bin width goes back to default, which is IMO the desired behaviour.

d-callan:
I mean when I configure a viz and then minimize it, the thumbnail version should probably keep the configuration. I didn't mean to apply the configuration to a new viz. Sorry I wasn't clear.

bobular:
That's OK.

The behaviour you describe should be what happens!  There may be a bug.
Could you give more details please?

bobular:
Did a quick test. The scatter plot config (the plot type radio buttons) is
honoured in the thumbnail versions, so I don't see the problem you see -
yet!

d-callan:
right. it may be that what im asking for isnt very easy. but the client side controls dont seem to persist. like if i click something off in the legend for ex. but im not sure a user has any reason to know why some things persist and others dont. itll probably be particularly noticeable for something like switching axes. because some types of plots i think client side switching makes sense, and others a new request should be made to the data service.

d-callan:
or maybe its even specific to the plotly legend, i havent played around w it too much yet tbh.

dmfalke:
We are not persisting the result of legend interactions. I'm sure this is something we can address.

bobular:
Oh I see.  Won't it be nasty persisting internal plotly things?  That's why we've disabled virtually all of the interactivity features.

bobular:
implementation hint: plotly has a callback to capture entire state of a plot (ask @dmfalke for more)

bobular:
post phase 1 - roll our own legend will make this work redundant (will need to redo it)

nkittur-uga:
[...]

PR #637, opened 2021-10-27
Custom legend

moontrip:
@bobular @dmfalke  I have made PR for initial custom legend works presented at the dev call. This relies on [the PR at web-components](https://github.com/VEuPathDB/web-components/pull/237) as well.


src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -543,10 +651,18 @@ function ScatterplotWithControls({
...
+  //DKDK add async/await for correct thumbnail capture
   useEffect(() => {
-    plotRef.current
-      ?.toImage({ format: 'svg', ...plotDimensions })
-      .then(updateThumbnailRef.current);
+    (async () => {
+      if (data != null && data.series.length > 1) {
+        // const nameArray = data.value?.dataSetProcess.series.map((data: any) => data.name);
+        // use this to set all checked
+        await setCheckedLegendItems(legendItems.map((item) => item.label));
```
bobular, line 660: Also here

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -334,6 +342,82 @@ function ScatterplotViz(props: VisualizationProps) {
...
+            // scatter plot has defined mode - still need to find the way to distinguish CI, No data, etc.
+            marker: data.mode != null ? data.mode : '',
+            //DKDK think markerColor needs to be added here
+            markerColor: 'markerColor',
+            // set hasData based on filters: do we need to check other type like dateSet, numberSet etc?
+            hasData:
+              filters != null && overlayVariableFilter
+                ? filters
+                    .map((filter) => {
+                      if (filter.type == 'stringSet' && data.name != null)
```
bobular, line 388: Same here :-)

bobular (changes requested):
Looks good apart from the thing we already discussed about checking for hasData.

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -451,11 +533,25 @@ function HistogramPlotWithControls({
...
   useEffect(() => {
-    plotRef.current
-      ?.toImage({ format: 'svg', ...plotDimensions })
-      .then(updateThumbnailRef.current);
-  }, [data, histogramProps.dependentAxisLogScale]);
+    (async () => {
+      if (data != null && data.series.length > 1) {
+        // const nameArray = data.value?.dataSetProcess.series.map((data: any) => data.name);
+        // use this to set all checked
+        await setCheckedLegendItems(legendItems.map((item) => item.label));
```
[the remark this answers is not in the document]
moontrip: I also noticed it but without that await, somehow it did not work correctly so I left it as is

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -315,6 +327,61 @@ function HistogramViz(props: VisualizationProps) {
...
+  //DKDK set useState to track checkbox status
+  const [checkedLegendItems, setCheckedLegendItems] = useState<string[]>([]);
+
+  //DKDK set all checkbox checked initially
+  useEffect(() => {
+    if (data.value != null && data.value.series.length > 0) {
+      // extract legendItems.label for passing to both PlotlyPlot and PlotLegend
+      setCheckedLegendItems(legendItems.map((item) => item.label));
+    }
+  }, [data]);
```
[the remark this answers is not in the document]
moontrip: yep will add!

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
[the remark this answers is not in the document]
dmfalke: Can you explain what doesn't work, when you remove `await`?

dmfalke (commented):
This all looks pretty good. i'm curious about the `await` issue (see inline comment).

moontrip:
@dmfalke so, I noticed an thumbnail capture issue at Scatter plot case with the custom legend: from my limited test it does not occur at other viz though. It happens when using plot option other than 'raw', i.e., smoothed mean or best fit, and thumbnail is not made correctly in those cases - showing previous plot. I think that it is because the custom legend's status is asynchronously happened as compared to thumbnail capture you made. So, I made async/await at the same useEffect so that custom legend is firstly updated and then thumbnail capture is done after that.

moontrip:
I think now both histogram and barplot vizs are working correctly with the custom legend. Surprisingly the checkbox behavior (turning on/off plot trace) worked just fine for the faceted plot without much changes (at least for histogram and barplot vizs) 🎉 Since they both used the same marker style (solid square without border), the current custom legend (PlotLegend component) are working for them even if currently it only has that square marker case. Will need to check marker style for other plot vizs and add markers accordingly.

dmfalke:
Thanks for the update @moontrip. One suggestion is to store the checkedLegendItems as a part of the visualization config object. If you do this, be sure to make it optional.

moontrip:
> Thanks for the update @moontrip. One suggestion is to store the checkedLegendItems as a part of the visualization config object. If you do this, be sure to make it optional.

@dmfalke Actually I thought of the case as well because, if I remember correctly, there was a talk to keep the status of legend selection. Perhaps I will think of that case next week as I am off from VEuPathDB today.

bobular:
> 
> 
> @dmfalke so, I noticed an thumbnail capture issue at Scatter plot case with the custom legend: from my limited test it does not occur at other viz though. It happens when using plot option other than 'raw', i.e., smoothed mean or best fit, and thumbnail is not made correctly in those cases - showing previous plot. I think that it is because the custom legend's status is asynchronously happened as compared to thumbnail capture you made. So, I made async/await at the same useEffect so that custom legend is firstly updated and then thumbnail capture is done after that.

The `await` does somehow magically make the thumbnail sort-of work.  However, it doesn't work as desired: the thumbnail isn't made from the last-viewed scatterplot, which may have series deselected via the new legend.

I think the problem can be solved by moving the "set all checked" useEffect to the main `ScatterplotViz` function, where I think it belongs more naturally (this is where `legendItems` is made).

Then the toImage useEffect just needs `checkedLegendItems` as a dependency.

Here's the diff from what I was playing around with

```diff
diff --git a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
index fa86efd4..30577b88 100755
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -365,6 +365,7 @@ function ScatterplotViz(props: VisualizationProps) {
       : [];
   }, [data, filters]);

+
   //DKDK
   console.log(
     'data.value?.dataSetProcess.series =',
@@ -377,6 +378,11 @@ function ScatterplotViz(props: VisualizationProps) {

   console.log('checkedLegendItems =', checkedLegendItems);

+  useEffect(() => {
+    // use this to set all checked
+    setCheckedLegendItems(legendItems.map((item) => item.label));
+  }, [legendItems]);
+
   return (
     <div style={{ display: 'flex', flexDirection: 'column' }}>
       <div style={{ display: 'flex', alignItems: 'center', zIndex: 1 }}>
@@ -499,7 +505,6 @@ function ScatterplotViz(props: VisualizationProps) {
           //DKDK pass checked state of legend checkbox to PlotlyPlot
           checkedLegendItems={checkedLegendItems}
           setCheckedLegendItems={setCheckedLegendItems}
-          legendItems={legendItems}
         />

         {/* DKDK custom legend */}
@@ -573,8 +578,8 @@ type ScatterplotWithControlsProps = XYPlotProps & {
   // add disabledList
   disabledList: string[];
   //DKDK legend checkbox
+  checkedLegendItems: string[];
   setCheckedLegendItems: (checkedItems: string[]) => void;
-  legendItems: LegendItemsProps[];
 };

 function ScatterplotWithControls({
@@ -589,8 +594,8 @@ function ScatterplotWithControls({
   disabledList,
   updateThumbnail,
   //DKDK legend checkbox
+  checkedLegendItems,
   setCheckedLegendItems,
-  legendItems,
   ...scatterplotProps
 }: ScatterplotWithControlsProps) {
   // TODO Use UIState
@@ -613,20 +618,17 @@ function ScatterplotWithControls({
   //DKDK add async/await for correct thumbnail capture
   useEffect(() => {
     (async () => {
-      if (data != null) {
-        // use this to set all checked
-        await setCheckedLegendItems(legendItems.map((item) => item.label));
-      }
       await plotRef.current
         ?.toImage({ format: 'svg', ...plotDimensions })
         .then(updateThumbnailRef.current);
     })();
-  }, [data, legendItems]);
+  }, [data, checkedLegendItems]);

   return (
     <div style={{ display: 'flex', flexDirection: 'column' }}>
       <XYPlot
         {...scatterplotProps}
+        checkedLegendItems={checkedLegendItems}
         ref={plotRef}
         data={data}
         // add controls
```

bobular:
However, a bigger structural change is that we need the checked items (if there are any) to be saved in `vizConfig`.  So I guess we should use `vizConfig.checkedLegendItems` (and `updateVizConfig` etc) for the storage instead of `useState()`

moontrip:
> > @dmfalke so, I noticed an thumbnail capture issue at Scatter plot case with the custom legend: from my limited test it does not occur at other viz though. It happens when using plot option other than 'raw', i.e., smoothed mean or best fit, and thumbnail is not made correctly in those cases - showing previous plot. I think that it is because the custom legend's status is asynchronously happened as compared to thumbnail capture you made. So, I made async/await at the same useEffect so that custom legend is firstly updated and then thumbnail capture is done after that.
> 
> The `await` does somehow magically make the thumbnail sort-of work. However, it doesn't work as desired: the thumbnail isn't made from the last-viewed scatterplot, which may have series deselected via the new legend.
> 
> I think the problem can be solved by moving the "set all checked" useEffect to the main `ScatterplotViz` function, where I think it belongs more naturally (this is where `legendItems` is made).
> 
> Then the toImage useEffect just needs `checkedLegendItems` as a dependency.
> 
> Here's the diff from what I was playing around with
> 
> ```diff
> diff --git a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
> index fa86efd4..30577b88 100755
> --- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
> +++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
> @@ -365,6 +365,7 @@ function ScatterplotViz(props: VisualizationProps) {
>        : [];
>    }, [data, filters]);
> 
> +
>    //DKDK
>    console.log(
>      'data.value?.dataSetProcess.series =',
> @@ -377,6 +378,11 @@ function ScatterplotViz(props: VisualizationProps) {
> 
>    console.log('checkedLegendItems =', checkedLegendItems);
> 
> +  useEffect(() => {
> +    // use this to set all checked
> +    setCheckedLegendItems(legendItems.map((item) => item.label));
> +  }, [legendItems]);
> +
>    return (
>      <div style={{ display: 'flex', flexDirection: 'column' }}>
>        <div style={{ display: 'flex', alignItems: 'center', zIndex: 1 }}>
> @@ -499,7 +505,6 @@ function ScatterplotViz(props: VisualizationProps) {
>            //DKDK pass checked state of legend checkbox to PlotlyPlot
>            checkedLegendItems={checkedLegendItems}
>            setCheckedLegendItems={setCheckedLegendItems}
> -          legendItems={legendItems}
>          />
> 
>          {/* DKDK custom legend */}
> @@ -573,8 +578,8 @@ type ScatterplotWithControlsProps = XYPlotProps & {
>    // add disabledList
>    disabledList: string[];
>    //DKDK legend checkbox
> +  checkedLegendItems: string[];
>    setCheckedLegendItems: (checkedItems: string[]) => void;
> -  legendItems: LegendItemsProps[];
>  };
> 
>  function ScatterplotWithControls({
> @@ -589,8 +594,8 @@ function ScatterplotWithControls({
>    disabledList,
>    updateThumbnail,
>    //DKDK legend checkbox
> +  checkedLegendItems,
>    setCheckedLegendItems,
> -  legendItems,
>    ...scatterplotProps
>  }: ScatterplotWithControlsProps) {
>    // TODO Use UIState
> @@ -613,20 +618,17 @@ function ScatterplotWithControls({
>    //DKDK add async/await for correct thumbnail capture
>    useEffect(() => {
>      (async () => {
> -      if (data != null) {
> -        // use this to set all checked
> -        await setCheckedLegendItems(legendItems.map((item) => item.label));
> -      }
>        await plotRef.current
>          ?.toImage({ format: 'svg', ...plotDimensions })
>          .then(updateThumbnailRef.current);
>      })();
> -  }, [data, legendItems]);
> +  }, [data, checkedLegendItems]);
> 
>    return (
>      <div style={{ display: 'flex', flexDirection: 'column' }}>
>        <XYPlot
>          {...scatterplotProps}
> +        checkedLegendItems={checkedLegendItems}
>          ref={plotRef}
>          data={data}
>          // add controls
> ```

@bobular thank you for your suggestion. Actually the approach I have used async/await for both legendItems and thumbnail at useEffect was designed for scatter plot and it worked just fine. I don't know why you did not have the screenshot correctly. Anyway, the approach you suggested was my initial approach to use useEffect for legendItems (set function) at ScatterplotViz but it didn't work properly for me, especially with plot options with smoothed mean and best fit line: that was the reason why I switched to async/await. However, what I missed is the dependency at useEffect of thumbnail so it may make difference. I will test it too!

And yes I started to set vizConfig.checkedLegendItems for histogram viz and then barplot viz. If it would be done, I think that we can merge this custom legend stuff so that I can add this custom legend to other plot vizs that are being grounded by you 😄

moontrip:
@bobular after testing, the missing dependency at thumbnail's useEffect() at my implementation was the culprit to prevent from correct capture 👍

dmfalke:
Great! Good catch @bobular, and nice work @moontrip!

@moontrip, there is a [long-standing issue](https://github.com/VEuPathDB/web-eda/issues/172) to retain the state of what legend items are visible.

moontrip:
> 
> @moontrip, there is a [long-standing issue](https://github.com/VEuPathDB/web-eda/issues/172) to retain the state of what legend items are visible.

@dmfalke I will think of that specific case later: need to make custom legend work for all vizs, including faceted plot, at first :)

moontrip:
@bobular and @dmfalke histogram and barplot vizs with faceted plot are now changed to use vizConfig. 

@bobular Actually I also changed scatterplot viz in this branch (i.e., without faceted plot) to use vizConfig, but I am hesitating to commit because you already started working previous scatter plot to have faceted plot: in that case I can change yours later. If you didn't start the scatter plot yet, then I will also commit it here. Just let me know 👍

bobular:
> Actually I also changed scatterplot viz in this branch (i.e., without faceted plot) to use vizConfig, but I am hesitating to commit because you already started working previous scatter plot to have faceted plot: in that case I can change yours later. If you didn't start the scatter plot yet, then I will also commit it here. Just let me know

Hi @moontrip  - I haven't started on scatterplot faceting - so please commit!

moontrip:
> > Actually I also changed scatterplot viz in this branch (i.e., without faceted plot) to use vizConfig, but I am hesitating to commit because you already started working previous scatter plot to have faceted plot: in that case I can change yours later. If you didn't start the scatter plot yet, then I will also commit it here. Just let me know
> 
> Hi @moontrip - I haven't started on scatterplot faceting - so please commit!

Thank you for your info, @bobular 👍 I just committed scatter plot viz as well.

moontrip:
now make this as a normal PR so that it can be used for other Vizs with faceted plot, which I will handle the addition of this custom legend to those vizs - done for histogram and barplot vizs, and scatter plot viz in part.

dmfalke (changes requested):
I think this all looks pretty good. I added an inline comment about the state of the checked items not being retained when I reopen an existing visualization. Once that has been addressed, I think this will be good to go.

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -366,6 +379,54 @@ function HistogramViz(props: VisualizationProps) {
...
+      : [];
+  }, [data]);
+
+  // use this to set all checked
+  useEffect(() => {
+    if (data != null) {
+      // use this to set all checked
+      onCheckedLegendItemsChange(legendItems.map((item) => item.label));
+    }
+  }, [data, legendItems]);
```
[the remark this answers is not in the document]
moontrip: @dmfalke thank you for your suggestion. Actually I am putting that work in the future (possibly next release as I have several things to do) as it would be quite tricky one to make appropriate conditions that cover up all use cases. To accomplish it, it should be like: 

```
useEffect(() => {
...
(if condition A) - generate checkedLegendItems
onCheckedLegendItemsChange(legendItems.map((item) => item.label));
(else/else if condition B) - this will keep previous checkedLegendItems for a use case of leaving the viz and coming back to the one
onCheckedLegendItemsChange(legendItems.map(vizConfig.checkedLegendItems);
...
```

Although there are a couple of more use cases that your suggestion is not working, One simple use case is when a user **changes** variable (X, Y, Main, or Overlay) from existing one - since vizConfig.checkedLegendItems was previously defined in this case, checkedLegendItems does not change because onCheckedLegendItemsChange is not called.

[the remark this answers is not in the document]
dmfalke: That makes sense. I'm fine with putting this off, then.

dmfalke (approved):

bobular:
Reviewing as fast as I can now.

From testing, I see there's a general logic/wiring issue that I'll make another ticket for and fix at a later date. Here's the ticket: https://github.com/VEuPathDB/web-eda/issues/676

We shouldn't show the gray No data option for overlay variable sex in the example below.  The problem is that `showMissing` or `showMissingOverlay` booleans don't take into account the `completeCases` data, which they should.
image

moontrip:
Thanks @bobular ! 👍 I also saw your report regarding the bug(?). Initially I thought I made a mistake to incorrectly show the legend item, No data, but after reading the description carefully, I realized that it was related to the switch behavior. Glad that my custom legend helps to identify the issue (and also relieved myself a bit that the legend logic worked fine :) )

src/lib/core/hooks/findOutputEntity.ts
```diff
@@ -8,7 +8,7 @@ export function useFindOutputEntity(
   // need to add string at Record's Type due to valueSpecConfig
   dataElementVariables: Record<
     string,
-    VariableDescriptor | string | boolean | undefined
+    VariableDescriptor | string | boolean | string[] | undefined
```
bobular, line 11: Oh, I see why... because we pass the whole of `vizConfig` to this function.  Looks like some technical debt here!

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -366,6 +379,54 @@ function HistogramViz(props: VisualizationProps) {
...
+              rank: 1,
+            };
+            // histogram viz uses stack option so reverse the legend order!
+          })
+          .reverse()
+      : [];
+  }, [data]);
+
+  // use this to set all checked
+  useEffect(() => {
```
bobular, line 423: I thought I had a simple solution to the problem of this overwriting the saved configuration, but it wasn't so simple.

So are we going to solve this separately?  Don't forget to make a ticket if so.

bobular (approved):
Looks good - note that I've done more testing than code review.

Don't forget to add tickets for the remaining tasks, which I think are

- legend markers for scatterplot
- saving/restoring checked legend items correctly

src/lib/core/hooks/findOutputEntity.ts
moontrip: yes indeed

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: will make a ticket to handle saving/restoring checked legend items. Actually legend markers are for both boxplot and scatter plot as boxplot utilized different marker style unlike histogram etc.: square but light color and border for boxplot ones.

moontrip:
thank you for your review and comments @bobular ! 👍 Will work for merging

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: As pointed out by Dave, we already have saving/restoring ticket made long time ago, #172, so I don't need to make it: . just assigned it to me so that I won't forget 😃 Considering other tasks I have to do for b55, I will defer it, though.

PR #640, opened 2021-10-29
Faceting: Wiring up Barplot and Histogram visualizations for basic FacetedPlot component.

bobular:
Depends on https://github.com/VEuPathDB/web-components/pull/162

I think the big question is...

Do we want to enforce the facet layer(s) in the plot components' data types?
A non-faceted plot would have a dummy/null facet:

```
facets: [
  {
    label: undefined,
    data: [ regular plot component data ],
  }
]
```

We'd have to think how this would look if there were two facet "dimensions"...  keep one `label` but add `x` and `y` coordinates for layout?


bobular:
OK, so I am wondering whether we should merge this and I work on the remaining "wiring" of boxplot, scatter and mosaic in separate PRs.

This would give everyone the bare bones to work with, without having to branch from a branch.

Thoughts @dmfalke @jtlong3rd ?

bobular:
Don't review or approve yet!  Still some wiring issues to sort out with Barplot (and presumably Histogram)

bobular:
> Don't review or approve yet! Still some wiring issues to sort out with Barplot (and presumably Histogram)

OK, all good again now!

PR #641, opened 2021-10-29
DataGrid View for Subsetting

eikonomega:


dmfalke (commented):
I've reviewed all of the changes, and they all look good! Nothing stands out as problematic. I am now going to do a sanity check on my local dev site, and then address some follow-up issues we discussed in the past:

1. Add client-side data restriction
2. Address styling issues w/ the CheckboxTree component (if any)
3. Hide the button that invokes the modal

dmfalke (approved):

Issue #572, opened 2021-09-18
General: add "Notes" tab (for Phase 2)


danicahelb:
Dave had a great idea to add a “Notes” tab next to “Visualize” where the user can write notes about that particular saved analysis (ie, purpose, steps taken, etc)

Users could fill out the study description (needed for making an analysis public) within the notes tab; UI for making public analyses could guide them to filling out a designated section within the notes tab

steve-fischer-200:
we need to confirm that the Eda User Service supports both Brief Description and Details.

(We need the brief description for display in the Analyses Tables)

eikonomega:
This ticket has a dependency on https://github.com/VEuPathDB/EdaUserService/issues/10

nkittur-uga:
Tested Notes tab on qa.restricted and it is working well! 
There is a small issue with the text box overlapping the footer so I am not moving to "Complete" yet.

<img width="1084" alt="Screen Shot 2021-11-10 at 2 06 04 PM" src="https://user-images.githubusercontent.com/67710160/141177371-338057c5-ef47-46cb-b3b1-f9b4046c0ac9.png">

nkittur-uga:
Text box is no longer overlapping footer. Looks good!

PR #642, opened 2021-10-29
NotesTab

eikonomega:
Fulfills #572


dmfalke:
Doing some functional testing. Looks great! One thing I noticed is the character count in the description collides with the text.
image

package.json
```diff
@@ -2,29 +2,31 @@
...
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@mui/styles": "^5.0.1",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.8.8",
     "@veupathdb/core-components": "^0.2.43",
+    "@veupathdb/web-common": "^0.1.10",
```
dmfalke, line 13: This should not be a dependency. It is only needed as a devDependency, and will create a circular dependency if left here.
 
```suggestion
```

package.json
```diff
@@ -2,29 +2,31 @@
...
     "fp-ts": "^2.9.3",
     "io-ts": "^2.2.13",
-    "lodash": "^4.17.20",
+    "lodash": "^4.17.21",
     "monocle-ts": "^2.3.11",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-draggable": "^4.4.3",
     "react-resizable": "^3.0.4",
+    "react-router": "^5.2.1",
```
dmfalke, line 28: I think only `react-router-dom` is needed.

```suggestion
```

public/index.html
```diff
@@ -9,6 +9,12 @@
       name="description"
       content="Web site created using create-react-app"
     />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Inter:wght@100;200;300;400;500;600;700;800;900&family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
```
dmfalke, line 17: Where are these fonts being used?

dmfalke (changes requested):
Once the dependencies are cleaned up, and the glitch I mentioned above is addressed, I think this will be good to go!

package.json
eikonomega: I see a number of places using imports from react-router.

For ex. see `MapVeuContainer.tsx` or `AnalysisPanel.tsx`

dmfalke (approved):

PR #644, opened 2021-10-30
add temporal var constraint

### LINKED ISSUE #643: add isTemporal variable constraint
this will be needed for the x axis of the time series plot, when we get to it, where we only want terms annotated as isTemporal = true to be available.

d-callan:
should resolve #643

### DIFF
--- a/src/lib/core/utils/data-element-constraints.ts
+++ b/src/lib/core/utils/data-element-constraints.ts
@@ -22,6 +22,7 @@ export function filterVariablesByConstraint(
     (constraint.allowedShapes == null &&
       constraint.allowedTypes == null &&
       constraint.maxNumValues == null &&
+      constraint.isTemporal == null &&
       constraint.allowMultiValued)
   )
     return rootEntity;
@@ -73,6 +74,8 @@ function variableConstraintPredicate(
         constraint.allowedTypes.includes(variable.type)) &&
       (constraint.maxNumValues == null ||
         constraint.maxNumValues >= variable.distinctValuesCount) &&
+      (constraint.isTemporal == null ||
+        constraint.isTemporal == variable.isTemporal) &&
       (constraint.allowMultiValued || !variable.isMultiValued))
   );
 }
@@ -117,6 +120,7 @@ export function flattenConstraints(
       if (
         isEmpty(constraint.allowedShapes) &&
         isEmpty(constraint.allowedTypes) &&
+        isEmpty(constraint.isTemporal) &&
         constraint.maxNumValues === undefined &&
         constraint.allowMultiValued
       )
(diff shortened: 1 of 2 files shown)


src/lib/core/utils/data-element-constraints.ts
```diff
@@ -139,12 +143,16 @@ export function flattenConstraints(
       const passesMaxValuesConstraint =
         constraint.maxNumValues === undefined ||
         constraint.maxNumValues >= variable.distinctValuesCount;
+      const passesTemporalConstraint =
+        isEmpty(constraint.isTemporal) ||
+        constraint.isTemporal == variable.isTemporal;
```
dmfalke, line 148: This means that, if `isTemporal` is set for the variable constraint, then it **must** also be set on the variable for it to pass. If the constraint contains `isTemporal = false`, would a variable without `isTemporal` set be considered a viable candidate?

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -170,6 +178,12 @@ export function mergeConstraints(
     keys.map((key): [string, DataElementConstraint] => {
       const constraintA = constraintMapA[key];
       const constraintB = constraintMapB[key];
+      const mergedIsTemporal =
+        constraintA == null || constraintB == null
+          ? undefined
+          : constraintA.isTemporal && constraintB.isTemporal
+          ? true
+          : undefined;
```
dmfalke, line 186: This doesn't seem quite right, to me. If `isTemporal` is `false` for both constraints, or if only one has it set, then the merged value is `undefined`. I think if only one is set, then that one should be the merged value. If both are set, then the `&&` evaluation should be the merged value:

```typescript

      const mergedIsTemporal =
        constraintA == null || constraintB == null
          ? undefined
          : constraintA.isTemporal == null ? constraintB.isTemporal
          : constraintB.isTemporal == null ? constraintA.isTemporal
          : constraintA.isTemporal && constraintB.isTemporal;
```

I suppose this also depends on how we treat the absence of `isTemporal`...

dmfalke (commented):
I've made some inline comments. I think the semantics of `isTemporal == null` will determine if any changes are required.

src/lib/core/utils/data-element-constraints.ts
d-callan: hmm. i had assumed, perhaps erroneously, that isTemporal would always be available for a variable. i figured if it was annotated as being temporal that value would be true, false otherwise. is that not the case?

src/lib/core/utils/data-element-constraints.ts
d-callan: i think youre right about the case where both are false, the merged result should also be false. but if either is undefined, i take that to mean its not restricted on temporal terms either way, and the merged result shouldnt be either.

d-callan:
went for lineplot rather than time series in the end

Issue #636, opened 2021-10-27
Scatterplot/XYPlot has a lot of client-side logic and Plotly specifics


bobular:
Before tackling this, we need to think about how we will handle line plots, density plots etc.

There should be full, Plotly-independent data APIs defined for each of these, and probably separate web-components: ScatterPlot, DensityPlot, LinePlot, etc which call XYPlot.

PR #645, opened 2021-10-31
Lineplot viz

### LINKED ISSUE #825: Visualizations frontend: Line plot viz tool (continuous y-vars only)
We need a working line plot viz tool for b56.

Mockups can be found here: https://docs.google.com/presentation/d/16xCzGUs744KD3IQ91ITMt30pbAF9SJ7yvOkWZoE-AeE/edit#slide=id.p

- _Note that the deliverable for b56 only includes "General Line Plots" in the mockups (ignore "Frequency line plots", "Cumulative frequency plots", and "Kaplan Meier plots" for now)_
- _Note also that categorical y axis variables are NOT included in this current ticket_ - see #956 for that

NOTES FOR TESTING/DEMO-ING

* Do not use medians and error bars at the same time, or things will look a bit funny. Known back end issue.
* Mouse-overs are all on one line. Splitting into a nicer layout may/may not be possible. Future ticket?
* When switching off binning for date x-variables, actually a bin width of "1 day" is sent to the back end. This avoids a 500 error. Was this what we intended anyway, @d-callan?
* Y-axis starting at zero or not?  Not sure yet. Wasn't in the mock-ups. Comments please!

d-callan:
Resolves #825 
Depends on https://github.com/VEuPathDB/web-components/pull/289


bobular:
Yeah, there is definitely refactoring on the table... https://github.com/VEuPathDB/web-eda/issues/636

Cool to see you on the client side!  Will review after I do some Redmine pennance.

src/lib/core/components/visualizations/PlaceholderIcon.tsx
```diff
@@ -3,7 +3,6 @@ import rxc from './implementations/selectorIcons/RxC.svg';
 import heatmap from './implementations/selectorIcons/heatmap.svg';
 import density from './implementations/selectorIcons/density.svg';
 import map from './implementations/selectorIcons/map.svg';
-import line from './implementations/selectorIcons/line.svg';
```
bobular, line 6: Confused why you removed the icon...?

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -170,6 +178,12 @@ export function mergeConstraints(
     keys.map((key): [string, DataElementConstraint] => {
       const constraintA = constraintMapA[key];
       const constraintB = constraintMapB[key];
+      const mergedIsTemporal =
```
bobular, line 181: This looks like a no-op.  It's not used below.

Do you mean to return
```
  isTemporal: mergedIsTemporal
```
in the return object?

src/lib/core/components/visualizations/PlaceholderIcon.tsx
bobular: Also the icon is greyed out and not clickable, so I can't do any testing of the viz. From the code I can't immediately see why (looks like you've implemented everything and wired it up), but can look deeper later. Or maybe it's something at my end.
image

bobular (commented):
Look forward to seeing this (see comments) !!

src/lib/core/utils/data-element-constraints.ts
d-callan: yes. thanks.

src/lib/core/components/visualizations/PlaceholderIcon.tsx
d-callan: just didnt seem like we needed it there any longer, as theres actually a viz for it to go in now. do we prefer to leave them?

bobular: Oh, duh, yeah - the clue's in the name!

bobular:
Hi @d-callan 

The add new viz button isn't working because of a mismatch between the name of the plugin in the `/apps` endpoint response (where it is called `lineplot` with displayname `Time series`)

And in `PassThroughComputation.tsx` where is is `timeseries` but I think will work when it is `lineplot`.  Just testing this...  Yes that's all that's needed to fix it.  Will add a comment/suggestion as a review.

src/lib/core/components/computations/PassThroughComputation.tsx
```diff
@@ -32,7 +33,7 @@ const visualizationTypes: Record<string, VisualizationType> = {
   twobytwo: twoByTwoVisualization,
   conttable: contTableVisualization,
   scatterplot: scatterplotVisualization,
-  // lineplot: scatterplotVisualization,
+  timeseries: lineplotVisualization,
```
bobular, line 36:
```suggestion
  lineplot: lineplotVisualization, // Time series
```

d-callan: thanks @bobular . that should resolve w the new backend changes. also explains why you were seeing something different to me. im running the data service branch locally as well.

d-callan: i suppose now ive tested it a bit i could merge the data service work.. but was kind of hoping to merge these together given the api changes. if what you suggest above is enough to get things testable for you, maybe thats fine for now?

bobular: Actually I didn't manage to get any plots working with my minor change - but I haven't looked into why - as I wasn't sure it was ready for that kind of testing (I realise now it is) - I'll have another look tomorrow.

dmfalke (commented):
I haven't really dug into this, nor have I attempted to test it. I can do so soon, running the backend branch locally.

I agree with your approach in splitting the scatter and line plot visualizations. I think this will make future refactoring more straightforward.

Thanks for working on this!

d-callan:
i did merge the data service work yesterday, fyi

d-callan:
i did try to incorporate recent work around things like facets and custom legends. youll have to let me know how successful i was.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -0,0 +1,1088 @@
...
+      else if (vizConfig.yAxisVariable == null || yAxisVariable == null)
+        return undefined;
+
+      const vars = [xAxisVariable, yAxisVariable, overlayVariable];
+      const unique = vars.filter((item, i, ar) => ar.indexOf(item) === i);
+      if (vars.length !== unique.length)
+        throw new Error(
+          'Variables must be unique. Please choose different variables.'
+        );
+
```
dmfalke, line 331: This is handled at lines 303-311
```suggestion
```

dmfalke (approved):
Overall, this looks good. I don't see anything egregious to be concerned about. I did suggest removing some duplicate code, but that's it.

In terms of the shifting axes when toggling overlay variables, you will want to look at the `independentAxisRange` and `dependentAxisRange` props passed to `LineplotWithControls`. I haven't looked at what is really going on, but it seems like it is currently only looking at variable metadata.

Nice work!

moontrip:
> In terms of the shifting axes when toggling overlay variables, you will want to look at the `independentAxisRange` and `dependentAxisRange` props passed to `LineplotWithControls`. I haven't looked at what is really going on, but it seems like it is currently only looking at variable metadata.
 
Just FYI, we use independentAxisRange based on metadata, while dependentAxisRange is computed based on data as the backend does not return min/max stats. defaultDependentAxisRange() and defaultIndependentAxisRange() utils that are used at most Vizs would be helpful - please refer to other viz like histogram Viz (scatterplot Viz use them too but slightly different because we added some empty margins not to cut markers at edges): though, they will be potentially changed by me due to other tasks.

bobular (commented):
Looks great from a functional point of view, though from a dataviz point of view I'm not sure the plots I made are very informative:
image

Seems that binning and error bars are essential for this kind of fine-grained temporal data

Looks much nicer for already pre-binned age-in-months:
image

overlay is 
image
btw

I noticed that `getBounds()` is unused.

In general we have an issue with the EDA 'clients' of XYPlot - they have to do a lot of processing into Plotly-like data - and this is something we should try to sort out. But I'm not sure when...

I recommend looking into the suggestions from @moontrip about the utility functions handling axis ranges.

bobular:
> i did try to incorporate recent work around things like facets and custom legends. youll have to let me know how successful i was.

ShowMissing etc work properly as far as I can tell :-)

image


I noticed the --x-- markers on the line there.

They seem to appear when there aren't many points to join a line between

image

Is this intentional or some default Plotly behaviour showing through?

d-callan:
im still adding features. binning, error bars and the ability to support the categorical y vars are all on the list. @danicahelb requested binning be the last on that list, but i agree the more i play w this the more binning seems to be a priority in my mind. 

i also noticed the markers down show consistently. i havent looked at that yet though.

d-callan:
im going to leave this a draft while i continue adding features etc.

d-callan:
also, i need to consider yet that x axis binning is intended to be optional

jtlong3rd (approved):
The functionality works well!

PR #649, opened 2021-11-02
@veupathdb/components and plotly.js version up

moontrip:
simply version up of @veupathdb/components did not resolve plotly.js version, so I have added it at dependencies. But also realized that in the meantime, plotly.js was upgraded to v2.6.0. For matching the version, I also made a [PR at web-components](https://github.com/VEuPathDB/web-components/pull/242)


moontrip:
I have also looked at plotly.js's release note but no specifically potential issue was noticed for our cases: https://github.com/plotly/plotly.js/releases/

bobular (approved):

package.json
```diff
@@ -21,6 +21,7 @@
     "io-ts": "^2.2.13",
     "lodash": "^4.17.20",
     "monocle-ts": "^2.3.11",
+    "plotly.js": "^2.5.1",
```
bobular, line 24: Are we 100% sure we need this?  I'm going to try without...

bobular (changes requested):
I approved, but then I saw the direct dependency for plotly.   Not sure!

package.json
bobular: I'm stuck too!

If I do `yarn remove plotly.js` then yarn.lock seems to have entries for v2.6.0 but the all important `node_modules/plotly.js/package.json` still says `"version": "1.58.5"`

I tried deleting the node_modules directory and rebuilding but it's the same.

Very strange - need help please @dmfalke @jtlong3rd

moontrip: @bobular First of all, hank you for your approval and merge at web-components PR - I just checked emails. As for the plotly.js entry, yes it probably needs to have it at dependencies as it did not automatically upgrade with the web-components. That was the reason why I added it in there. This issue arose sometime last week when Connor noticed this. Such an approach was my solution to resolve.

dmfalke: I believe the issue last week was to do with using `yarn link`. That will not necessarily trigger a transitive dependency to be upgraded.

Running `yarn upgrade <packageName>` should cause transitive dependencies to be updated. I will investigate further soon.

dmfalke: I was able to resolve this by deleting `yarn.lock` and running `yarn install` again. I just pushed this change.

dmfalke (approved):

Issue #287, opened 2021-08-06
General: collapse nodes in variables tree if only one child


SheenaTomko:
Create the ability to collapse nodes in the variable tree. With the WDK site, this is controlled in the presenters. Need to decide if that's appropriate here too or not

chowington:
@danicahelb Can I get a link to an example for the WDK site?

And I'm not sure how I could let outreach control this...maybe @dmfalke has an idea?

chowington:
And for clarity, we're talking about doing something like this, right? Here temp1/2/3 are nested folders with only one item each. We want to collapse the variables like these folders are collapsed?

![image](https://user-images.githubusercontent.com/12377649/139947375-2e9a440b-02e2-4a11-b5c4-d0dbf4af1e95.png)

SheenaTomko:
Here's an example from India ICEMR Cohort. The first screenshot of the owl file shows that the variable "Electricity" is under the Category "Assets". However, since there is only one variable under that category, on the WDK site, Electricity appears directly under "Socioeconomic factors" (second screenshot). Note that we want the variable to inherit the displayOrder of it's parent that is being hidden

<img width="295" alt="Screen Shot 2021-11-03 at 8 44 58 AM" src="https://user-images.githubusercontent.com/54597319/140071532-109f3df8-c2e0-4da3-a588-4e70d666084b.png">

<img width="226" alt="Screen Shot 2021-11-03 at 8 45 04 AM" src="https://user-images.githubusercontent.com/54597319/140071671-1233b05f-c24b-4cb6-965a-6dda57b9210a.png">

dmfalke:
@chowington I indicated above (https://github.com/VEuPathDB/web-eda/issues/287#issuecomment-936524595) how to implement this. Regarding making this configurable, we're going to have to think about how to do this. On the live site, we expose a property on the WDK Question object, but we don't have that concept in EDA. We can probably add a property or attribute to the WDK dataset record, but I think @steve-fischer-200 and I (at a minimum) will need to discuss options.

chowington:
@dmfalke Thanks for the tip and the response! I saw the comment when I started working on the ticket. I almost have the collapsing done, and I can get a PR up for it soon, but should I wait on the configurability solution for this ticket?

dmfalke:
[...]

PR #664, opened 2021-11-04
Collapse variable tree nodes if they have only one child

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/287


dmfalke (commented):
This looks great. I think it would be good to add a prop to `VariableList` to control this behavior, which could default to false (don't apply the behavior). I'm struggling to find a clear, short name for this prop. Maybe you have some ideas?

This will set us up to allow a property of the study record, which the clinepi team can set, to control this behavior.

dmfalke (approved):
Look good!

Issue #607, opened 2021-10-18
Visualizations: confusing behavior when copying an open visualization


danicahelb:
This isn't really a bug but I am marking it as such because it is very confusing.

I don't like the current behavior for clicking on "copy" from a visualization that is open. Note, the copy button from the visualization thumbnail works as I'd expect.

When I click the copy icon from within an open visualization, there is no indication that the visualization has been copied. This is because I am still working on the original version of the visualization, and a copied version has been silently created. I can see the copied visualization when I go back to the thumbnail view.

Instead, I would prefer that pressing the "copy" button brings me into an open copy of the visualization (titled "Copy of...) and I can modify this copy as I'd like. The original version should still be accessible in the thumbnail view and remain unchanged after I press copy.

chowington:
In addition to this change, I would suggest that we consider in the future adding an unintrusive popup notification for actions like this that can easily go missed, in the style of https://mui.com/components/snackbars/#customized-snackbars.

Also fun fact, the browser not going into the new copy is in fact a bug---after looking at the code, it's obvious to me that that was actually meant to happen :)

SAK22:
Says "Copy of" but not @chowington great suggestion, not sure if that should have been implemented yet as well? I think that would be really useful as I didn't notice it changed to "Copy of" very obviously, so if not should there be a new ticket or keep this one open?

PR #666, opened 2021-11-04
Fix bug that prevents automatic opening of new viz copy

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/607


jtlong3rd (approved):
Good catch!

PR #669, opened 2021-11-09
Rudimentary handling of empty facets for HistogramVisualization only

bobular:
This is one way to do it, but it's not ideal.
GEMS1 filter country=India

image

An improvement would be for FacetedPlot to render the facet title outside of Plotly (this would give us more control over styling and things like overflow/ellipsis handling).  FacetedPlot could also render a large "No data for this facet" for individual facets where `facet.data == null`.  Thoughts?

Note that the back end doesn't return empty data for empty facets, but it's easy for the client to put in an `undefined` where needed (I'll add an annotation to the PR).

### DIFF
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -802,21 +802,32 @@ function reorderData(
   facetVocabulary: string[] = []
 ): HistogramDataWithCoverageStatistics | HistogramData {
   if (isFaceted(data)) {
+    // for each value in the facet vocabulary's correct order
+    // find the index in the series where series.name equals that value
+    const facetValues = data.facets.map((facet) => facet.label);
+    const facetIndices = facetVocabulary.map((name) =>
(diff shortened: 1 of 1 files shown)


src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -802,21 +802,32 @@ function reorderData(
...
+          facetVocabulary[j] +
+          (data.facets[i] ? '' : ' (no plottable data for this facet)'),
+        data: data.facets[i]
+          ? (reorderData(
+              data.facets[i].data,
+              overlayVocabulary,
+              facetVocabulary
+            ) as HistogramData)
+          : // dummy data for empty facet
+            { series: [] },
```
bobular, line 825: This is where we could send `facet.data: undefined` to FacetedPlot

dmfalke (approved):
I agree with everything you said. Will you make a new issue? Thanks!

bobular:
Thanks - yes I have just made it: https://github.com/VEuPathDB/web-eda/issues/681

PR #675, opened 2021-11-09
Facet wiring mosaic 2x2 and RxC visualizations

bobular:
Depends on https://github.com/VEuPathDB/web-components/pull/249

Not expecting four reviews - but you're all welcome to have a look.

Note the faceted table version:
image


moontrip:
I have just tested this and it seems to work fine. But this raises a question to me concerning its combination with custom legend, which I will do. So, perhaps this is a good place to ask, especially @chowington and @bobular. 

The legend in the Mosaic plot Vizs (both 2x2 and RxC) now has left hand Plotly's legend, and it is not clickable. While faceted plot does not have such a legend. I wonder what kind of format the Mosaic custom legend should have ? Shouldn't it be clickable ? Or not even have to have checkbox in the custom legend ? In other words, is a static legend sufficient for Mosaic Vizs ?

moontrip:
In conjunction with my above question, perhaps we can introduce a hybrid approach if we need to keep the left-sided legend for X-/Y-axes variables (i.e., without facet variable). Like we only introduce custom legend for faceted plot, while we keep Connor's legend work for X/Y case (no faceted plot).

bobular:
> The legend in the Mosaic plot Vizs (both 2x2 and RxC) now has left hand Plotly's legend, and it is not clickable. On the other hand, faceted plot does not have such a legend. I wonder what kind of format the Mosaic custom legend should be ? Shouldn't it be clickable ? Or not even have to have checkbox in the custom legend ? In other words, is a static legend sufficient for Mosaic Vizs ?

Hi @moontrip - actually for Phase 2 build 55 I think the individual faceted mosaic plots **will** each have left-hand legends. FacetedPlot may need an extra prop to tell it not to turn off the per-plot legends...

@dmfalke is this what you thought the plan was?

moontrip:
Thank you for your info, @bobular Sounds like there is no need to have custom legend for Mosaic ones. I could see that FacetedPlot component has set to `displayLegend={false}` so yes exposing `displayLegend` as a prop would be able to control it.

moontrip:
@bobular a bug I found when enabling the toggle of no data, which causes the following error at statsTable. Will test more on the Mosaic

mosaicViz-bug1

bobular:
> @bobular a bug I found when enabling the toggle of no data, which causes the following error at statsTable. Will test more on the Mosaic

Thanks. I think this is what I reported in https://github.com/VEuPathDB/plot.data/issues/109

bobular:
> Thank you for your info, @bobular Sounds like there is no need to have custom legend for Mosaic ones. I could see that FacetedPlot component has set to `displayLegend={false}` so yes exposing `displayLegend` as a prop would be able to control it.

Or the Mosaic component can "extract" the displayLegend prop and not pass it on to PlotlyPlot?

dmfalke:
> > The legend in the Mosaic plot Vizs (both 2x2 and RxC) now has left hand Plotly's legend, and it is not clickable. On the other hand, faceted plot does not have such a legend. I wonder what kind of format the Mosaic custom legend should be ? Shouldn't it be clickable ? Or not even have to have checkbox in the custom legend ? In other words, is a static legend sufficient for Mosaic Vizs ?
> 
> Hi @moontrip - actually for Phase 2 build 55 I think the individual faceted mosaic plots **will** each have left-hand legends. FacetedPlot may need an extra prop to tell it not to turn off the per-plot legends...
> 
> @dmfalke is this what you thought the plan was?

I don't remember what the final decision was, regarding faceted mosaic. I can ask.

dmfalke:
> > > The legend in the Mosaic plot Vizs (both 2x2 and RxC) now has left hand Plotly's legend, and it is not clickable. On the other hand, faceted plot does not have such a legend. I wonder what kind of format the Mosaic custom legend should be ? Shouldn't it be clickable ? Or not even have to have checkbox in the custom legend ? In other words, is a static legend sufficient for Mosaic Vizs ?
> 
> > 
> 
> > Hi @moontrip - actually for Phase 2 build 55 I think the individual faceted mosaic plots **will** each have left-hand legends. FacetedPlot may need an extra prop to tell it not to turn off the per-plot legends...
> 
> > 
> 
> > @dmfalke is this what you thought the plan was?
> 
> 
> 
> I don't remember what the final decision was, regarding faceted mosaic. I can ask.

Confirmed this is the decision

dmfalke (approved):
I'm not going to have time to review this in detail, but nothing stands out as problematic, so I will go ahead and approve. Thanks!

moontrip (approved):

PR #677, opened 2021-11-09
Faceted Plot Thumbnails

jtlong3rd:
(Relates to https://github.com/VEuPathDB/web-components/issues/241; depends on the companion work in [this web-components PR](https://github.com/VEuPathDB/web-components/pull/250))

This PR introduces a utility for creating thumbnails for `FacetedPlots`, and abstracts our per-viz thumbnail-saving logic to a reusable `useUpdateThumbnailEffect` hook. As a proof of concept, the histogram viz is refactored to use `useUpdateThumbnailEffect`; refactoring the other visualizations should be just as straightforward, and could be done in a followup PR, once their associated wiring / custom legend work is merged.

**Note**

`useUpdateThumbnailEffect` is a stopgap; @dmfalke recently suggested that instead of triggering thumbnail saves via an effect, that we expose the `PlotRef` / `FacetedPlotRef` instance in the various callback props we pass to our visualizations.


src/lib/core/utils/thumbnails.tsx
```diff
@@ -0,0 +1,104 @@
...
+    >
+      {plotImageNodes}
+      <text
+        style={{
+          fontFamily: 'sans-serif',
+        }}
+        x={thumbnailDimensions.width * 0.95}
+        y={thumbnailDimensions.height * 0.1}
+        fontSize={thumbnailDimensions.height * 0.1}
+        textAnchor="end"
```
bobular, line 94: The number clashes with the mouse-over icons (trash can and duplicate), so I played around with a larger central number, but I'm sure we'll get more opinions from outreach!
```
        x={thumbnailDimensions.width * 0.5}
        y={thumbnailDimensions.height * 0.55}
        fontSize={thumbnailDimensions.height * 0.2}
        textAnchor="middle"
```
image
image

bobular (approved):
Looks good.

I personally don't quite understand the concept of a "ref callback" - and I had a similar question in the web-components PR - so maybe this is related.

But this is my problem, not yours!  I've googled "ref callback" and am now a bit the wiser!

src/lib/core/utils/thumbnails.tsx
dmfalke: What about the same size font that is current in use, but on the bottom right?

dmfalke (approved):
This is great. I agree that the number should be moved (see my reply to Bob's comment). Thanks!

src/lib/core/utils/thumbnails.tsx
jtlong3rd: In the short term, I went with a variation of @bobular's suggestion (centered, with a slightly smaller font size than proposed).

This implementation of @dmfalke's suggestion holds promise, but sometimes "clashes" with the lower-right-hand-plot:

```
        x={thumbnailDimensions.width * 0.925}
        y={thumbnailDimensions.height * 0.925}
        fontSize={thumbnailDimensions.height * 0.1}
        textAnchor="end"
```

image

Issue #635, opened 2021-10-27
Visualizations: Enable first iterate of faceting


dmfalke:
The following features comprise the first iterate of faceting:

1. One facet var - always column facet
3. Max var values allowed is 8
4. Thumbnail image shows 2 panels
5. No data always plotted in the last column
6. Fullscreen/enlarging one panel (extra - could be dropped if time does not allow)

There is a **web-components** [component](https://github.com/VEuPathDB/web-components/pull/162) to this feature, which is already under development.

Subtasks

- [x] merge https://github.com/VEuPathDB/web-components/pull/162
- [x] finish wiring/sorting (solves 4.) for all visualizations https://github.com/VEuPathDB/web-eda/pull/640 , #658, #659, #660
- [x] fully implement FacetedPlot https://github.com/VEuPathDB/web-components/issues/240
- [x] thumbnail image capture https://github.com/VEuPathDB/web-components/issues/241 
- [x] custom legend https://github.com/VEuPathDB/web-components/pull/237
- [x] placement/layout of legend/side plots/controls for faceted and unfaceted plots in EDA https://github.com/VEuPathDB/web-eda/issues/648
- [x] y-axis scaling for Boxplot https://github.com/VEuPathDB/web-eda/issues/650
- [x] handle empty facets appropriately - wiring done, just finalising some styling here: https://github.com/VEuPathDB/web-components/pull/262 
- [x] finalise mosaic layout #725 

ETA: here is a mock-up
![image](https://user-images.githubusercontent.com/365139/138996609-b2f34520-0504-4d12-971a-987d8dda9aed.png)

dmfalke:
@jtlong3rd has agreed to work on the thumbnail and layout issues.

bobular:
Closing manually.  All subtasks done :tada:

SheenaTomko:
Moving to complete. Steph is still qa'ing https://github.com/VEuPathDB/web-eda/issues/648, but that can be reopened or a new ticket made if needed

PR #687, opened 2021-11-11
Faceted plot styling

dmfalke:
refs #635 and https://github.com/VEuPathDB/web-components/issues/240
depends on https://github.com/VEuPathDB/web-components/pull/252

This PR is primarily focused on the styling of plots when they are used with `FacetedPlot`. Eventually, some of this work may be moved to web-components as variants of `FacetedPlot` (e.g., `FacetedHistogram`, etc).

Not, there are potentially some issues with ticks (definitely with mosaic). Any tips for how to address this would be great, but I can try to figure it out myself.

ETA I also threw in an ordering on the studies list.


moontrip:
@dmfalke Thank you for your info! 👍 I have briefly tested some cases but it seems to me that smaller size of faceted plots (half of normal plot) causes some issues for categorical data having lengthy texts, e.g., Floor material, etc. Certainly I believe you would address it but I doubt shorter ellipsis may not be sufficient for such cases. Here is an example screenshot. (GEMS1 Case Control)

barplot-floor-material

dmfalke:
Thanks for the feedback, @moontrip. I'm not really sure how best to address this. I will try by using a shorter length. I'm not really sure what other options we have.

dmfalke:
Another option is to make the plots taller. I will do that first, since it's easier

dmfalke:
I'm having a lot of trouble with mosaic
image

jtlong3rd (approved):
The code looks solid! 🤞 that the long categorical data label issue isn't too tricky to resolve.

bobular:
> I'm having a lot of trouble with mosaic

What's the problem?  The overlapping tick labels?

dmfalke:
I just pushed a commit that makes the subplots taller, which handles the long tick labels.

@bobular yes, I'm referring to the overlapping tick labels. However, I see this is not unique to faceting, so I guess we can kick the can down the road a bit longer.

jtlong3rd (approved):

Issue #650, opened 2021-11-02
Faceting: calculate dependentAxisRange from plugin response for Boxplot and maybe others


bobular:
See BarplotVisualization in `facet-approach1` branch or this [PR](https://github.com/VEuPathDB/web-eda/pull/640) for what was done there (`defaultDependentMaxValue` and `dependentAxisRange`)

moontrip:
made a PR for the Boxplot Viz, #688

SheenaTomko:
looks good

PR #688, opened 2021-11-11
set dependent axis range for Boxplot Viz with facet

moontrip:
This will address the ticket, [Faceting: calculate dependentAxisRange from plugin response for Boxplot and maybe others #650](https://github.com/VEuPathDB/web-eda/issues/650)

A part of an example screenshot is attached (GEMS1 Case Control; X: Floor material; Y: Sleeping rooms in dwelling count; Overlay: Drinking water container; Facet: Country). They showed the same dependent axis range

boxplot-dependent-axis-range

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
@@ -45,6 +45,8 @@ import {
   head,
   map,
   values,
+  max,
+  min,
 } from 'lodash';
 // import axis label unit util
 import { axisLabelWithUnit } from '../../../utils/axis-label-unit';
@@ -65,6 +67,7 @@ import PlotLegend, {
   LegendItemsProps,
 } from '@veupathdb/components/lib/components/plotControls/PlotLegend';
 import { ColorPaletteDefault } from '@veupathdb/components/lib/types/plots/addOns';
+import { NumberOrDateRange } from '@veupathdb/components/lib/types/general';
 
 type BoxplotData = { series: BoxplotSeries };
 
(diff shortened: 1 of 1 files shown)


src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -312,6 +315,73 @@ function BoxplotViz(props: VisualizationProps) {
...
+        (yAxisVariable?.type === 'number' || yAxisVariable?.type === 'integer')
+        ? {
+            min:
+              (min([
+                0,
+                min(
+                  data.value.facets.flatMap((facet) =>
+                    facet.data.series
+                      .flatMap((o) => o.outliers as number[][])
+                      .flatMap((o) => o)
```
jtlong3rd, line 331: You can use `.flat()` instead of `.flatMap((o) => o)`.

(Why is the method called `flat` and not `flatten`? [That's a long story...](https://developers.google.com/web/updates/2018/03/smooshgate))

jtlong3rd (approved):
I had one small style suggestion, but otherwise, looks good!

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
moontrip: @jtlong3rd Ah, you are absolutely right! 👍 I thought I used flat() but did not change it. Updated 😃

bobular (approved):
This is great.

The `as number` seem to be necessary because the Boxplot component takes NumberOrDateRange for the dependentAxisRange

But I wonder if it could just be NumberRange on the EDA side?

Leave it for now though!

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -312,6 +315,73 @@ function BoxplotViz(props: VisualizationProps) {
...
+                      .flatMap((o) => o.outliers as number[][])
+                      .flat()
+                  )
+                ),
+                min(
+                  data.value.facets.flatMap((facet) =>
+                    facet.data.series.flatMap((o) => o.lowerfence as number[])
+                  )
+                ),
+              ]) as number) * 1.05,
```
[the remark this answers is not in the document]
moontrip: I applied 5% increase for both min and max, just in case, i.e., multiplied by 1.05 regardless of +/-. So for example, if min and max is assumed to be [-1, 1], then actual plot range will be [-1.05, 1.05]: -1.05 is smaller than -1 😃

PR #693, opened 2021-11-11
Facet wiring for scatterplot

### LINKED ISSUE #659: Faceting: wire up Scatterplot visualization with basic faceting
Follow-on from #640

### LINKED ISSUE #690: Faceting Custom Legend - Scatterplot best fit R-squared in overlay variable legend needs to be removed
Don't want to forget this!

We need to replace the "R2 = 0.1234" with "best fit".   I'm not sure what the long-term fate of the R2 numbers will be.

image

bobular:
Resolves #659 
Resolves #690 

This was quite tricky but I refactored `processInputData` to only use the `response.scatterplot.data` data, so that I could use this function to process the data after the groupBy facetVariableDetails.value.  It needed some global response info, so this is now provided as the `hasMissingData` prop.

The rest is fairly straightforward.  I made some new tickets (in Next cycle) for some things I found along the way.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -384,7 +384,8 @@ function HistogramViz(props: VisualizationProps) {
   const legendItems: LegendItemsProps[] = useMemo(() => {
     const legendData = !isFaceted(data.value)
       ? data.value?.series
-      : data.value?.facets[0].data.series;
+      : data.value?.facets.find(({ data }) => data.series.length > 0)?.data
+          .series;
 
     return legendData != null
       ? legendData
@@ -408,9 +409,7 @@ function HistogramViz(props: VisualizationProps) {
                         el.data.series[index].bins != null &&
                         el.data.series[index].bins.length > 0
                     )
-                    .includes(true)
-                ? true
-                : false,
+                    .includes(true),
               group: 1,
               rank: 1,
             };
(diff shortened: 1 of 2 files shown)


moontrip:
@bobular thank you for your wiring works on the Scatter plot! 👍 I have briefly tested, and I noticed that faceted plot do not show all smoothed mean/confidence interval or best fit in the custom legend as compared to the one with the overlay variable only. Initially I thought there might be a bug but given that it is faceted, perhaps it is correct: when I deselected all smoothed mean/confidence interval or best fit related legend items at the faceted legend, all faceted plots did not show those graphs visually so it seems to be correct.

After this PR is merged, I will take care of legend marker stuffs

dmfalke:
I think the issue that @moontrip described is because the legend label for the best fit line is different for each subplot - it includes the R^2 value. I haven't confirmed, but my guess is that the first non-empty faceted data series is being used to determine the legend labels. I would further guess that a solution could be to remove `, R^2 = ...` from the plot label and series name.

There was discussion about this specific issue recently, and I believe the decision was to have a statistics table that includes the specific R^2 value for each facet, and to omit it from the legend.

dmfalke:
I was able to confirm the above with this change:

```diff
@@ -1202,10 +1202,8 @@ function processInputData<T extends number | string>(
           ? fixLabelForNumberVariables(
               el.overlayVariableDetails.value,
               overlayVariable
-            ) +
-            ', R² = ' +
-            el.r2
-          : 'Best fit, R² = ' + el.r2,
+            )
+          : 'Best fit',
         mode: 'lines', // no data point is displayed: only line
         line: {
           // use darker color for best fit line
```

moontrip:
@dmfalke Bob made a ticket regarding that, #690. As Bob mentioned in there, I think that it is better to indicate 'best fit' even if R-square part is removed.

bobular:
For the record, the smoothed mean lines have always been working for me.  
image

bobular:
Thanks for the reviews and testing.  Closing and merging.  Added ticket #694 for R^2 table

PR #695, opened 2021-11-12
Fix Histogram faceting bug #682

### LINKED ISSUE #682: Faceting / Visualizations: histogram faceting broken for some variables
After merging #669 it seems that something is very wrong

image

Histogram is fine with no facet variable.
Even nastier error is thrown when choosing "Floor material" as facet var.

Back end seems to be returning the proper data.  Front end must be broken :-(

bobular:
Resolves #682


bobular:
Merged as this is unlikely to break anything.

Issue #679, opened 2021-11-09
Visualization: Wiring custom legend to faceted plots


moontrip:
Bob is converting Vizs to support faceted plots, thus I will add the custom legend to those plots. Currently, histogram and barplot Vizs are done for custom legend at #637 but will need to work for other Vizs such as boxplot, mosaic, and scatter plot Vizs. This work also involves a task to set legendItems object appropriately, which is different from each plot.

bobular:
Remember to switch off legend in main plot

moontrip:
Boxplot Viz is done via #683. Mosaic (2x2 & RxC) would not have custom legend for now, so only Scatterplot Viz is left

nkittur-uga:
Custom legend looks good for histogram, bar plot, boxplot, scatter. But both mosaic plots don't have it yet so I am not moving to complete yet.

moontrip:
Hi @nkittur-uga thank you for your reviews and tests! 👍 Please note that we will not implement the custom legend to the two Mosaic plots (at least for the time being) because they utilizes Plotly's legend in the left side as Y-axis labels. For this reason, the custom legend is not necessary for them for now.

Issue #678, opened 2021-11-09
Visualization: Custom legend markers and colors


moontrip:
while a prototype of custom legend works are done, it is still required to add markers at legend labels like plotly's legend. A basic square markers are already made at #637 which would work for histogram, barplot, mosaic (2x2 and RxC) Vizs but still need to consider two more cases, scatter plot and boxplot Vizs. This ticket will handle the task.

moontrip:
Boxplot Viz is done via [#251](https://github.com/VEuPathDB/web-components/pull/251)

sweverschulman:
@moontrip 
Before I approve this ticket could you confirm the following things.
1. That the open dots are the correct icon in the scatter plot legend
2. That the clickable legend thing is supposed to be there 

I am attaching photos 
<img width="700" alt="Screen Shot 2021-11-15 at 11 24 45 AM" src="https://user-images.githubusercontent.com/62298921/141818202-114c985f-a40e-4eb5-a849-99abf524a4f3.png">

moontrip:
@sweverschulman Thank you for your review and tests. Sorry that it is not clear of your question 1: The attached one is boxplot and you are saying scatter plot legend. FYI, it seems that my custom legend works for the scatter plot (and possibly the latest works from developers) are not updated correctly at qa/qa.restricted: asking Dave about this. So, if you are asking about the style of scatter plot then yes they are almost identical to what the Plotly's default legend shows. You can see [my screenshot in this link](https://github.com/VEuPathDB/web-eda/pull/698#issue-1052360283), which shows what the new markers/colors of custom legend for the scatter plot look like.

As for the number 2, unlike Plotly's default legend that offers clicking texts, we are not supporting it. This is the reason why we have introduced checkbox instead, which controls on-off plot trace(s).

Hope this would be helpful for you.

sweverschulman:
thanks @moontrip. So to clarify the custom legend this. 
<img width="210" alt="Screen Shot 2021-11-15 at 11 40 33 AM" src="https://user-images.githubusercontent.com/62298921/141819984-ebf27e61-79ab-4c5f-95c8-015b80bf9fad.png">

moontrip:
@sweverschulman yes correct. FYI, Jamie just committed his work for overall layout of plot, custom legend, bird eye plot, and table: certainly not reflected at as site yet. Also, we will disable the Plotly's default legend sooner or later and stick to the new custom legend instead for both regular plots and faceted plot.

[...]

PR #698, opened 2021-11-12
legend markers and colors for scatter plot

moontrip:
Depends on the PR at web-components, [#253](https://github.com/VEuPathDB/web-components/pull/253)

This PR addresses #678  and #679, which are related to custom legend for scatter plot viz. 

A screenshot is attached:

scatterplot-legend-marker-example1


src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -89,6 +89,7 @@ import PlotLegend, {
 } from '@veupathdb/components/lib/components/plotControls/PlotLegend';
 import { isFaceted } from '@veupathdb/components/lib/types/guards';
 import FacetedPlot from '@veupathdb/components/lib/plots/FacetedPlot';
+import { config } from 'process';
```
jtlong3rd, line 92: Oops! I'm guessing you don't need this import?

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -388,18 +389,68 @@ function ScatterplotViz(props: VisualizationProps) {
...
+      ? // the name 'dataItem' is used inside the map() to distinguish from the global 'data' variable
+        legendData.map((dataItem: XYPlotDataSeries, index: number) => {
           return {
-            label: data.name != null ? data.name : '',
-            // TO-do: need a way to appropriately make marker info
-            // scatter plot has defined mode - still need to distinguish CI, No data, etc.
-            marker: data.mode != null ? data.mode : '',
-            // set marker colors
-            markerColor: 'markerColor',
+            label: dataItem.name != null ? dataItem.name : '',
```
jtlong3rd, line 416:
```suggestion
            label: dataItem.name ?? '',
```

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -388,18 +389,68 @@ function ScatterplotViz(props: VisualizationProps) {
...
+            marker:
+              dataItem.mode != null
+                ? dataItem.name === 'No data'
+                  ? 'x'
+                  : dataItem.mode === 'markers'
+                  ? 'circle'
+                  : dataItem.mode === 'lines'
+                  ? 'line'
+                  : ''
+                : dataItem.fill != null && dataItem.fill === 'tozeroy'
```
jtlong3rd, line 427: Maybe something like

```suggestion
                : dataItem?.fill === 'tozeroy'
```

jtlong3rd (approved):
Save for some small inline comments, this should be good to merge!

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: @jtlong3rd Thank you for your review and suggestions! 👍 IDK why it was entered but perhaps due to a sort of VS Code's function which is often not that smart 😄

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: Yep will do

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: Yep!

bobular (changes requested):
Looks good.  Nice colormap solution to the tricky problem of all the series being interleaved.

Might be nice to put `', Smoothed mean'`, `', 95% Confidence interval'` etc into "global" constants for this file, so that future edits will only need to be done in one place.

There is potentially a problem with some real-life variables having actual "No data" values from the database/back end.  I don't know one off the top of my head, but will ask Sheena/Danica.

moontrip:
> Looks good. Nice colormap solution to the tricky problem of all the series being interleaved.
> 
> Might be nice to put `', Smoothed mean'`, `', 95% Confidence interval'` etc into "global" constants for this file, so that future edits will only need to be done in one place.
> 
> There is potentially a problem with some real-life variables having actual "No data" values from the database/back end. I don't know one off the top of my head, but will ask Sheena/Danica.

@bobular thank you for your review and suggestions! 👍 As for global constants, do you mean that the computation of the legendLabel is separated as a function/util ? It would be great if you can elaborate it or show an example for me.

I could see the potential issue if actual data having "No data" text in the variable name, but the only problematic case is it has exact "No data" as the string.includes() is case-sensitive - not quite sure if there still exists the exception, though

bobular:
Hi DK - I pushed a minor commit with the string literals as constants - no biggie.
One small step towards internationalisation (which we don't have to do, haha).

I have found an example of 'No data' as actual data and I'm going to spend half an hour to see if I can mitigate the problem now.  Enjoy your weekend!

bobular (approved):
I will make another "'No data' from database" ticket because it requires discussion with the data curation team too.  All looks good here. Thanks.
Here's the new ticket: https://github.com/VEuPathDB/EdaNewIssues/issues/83

moontrip:
@bobular Aha I got your point now with your changes - I was not quite sure of the exact meaning you had in mind. Thank you for changing it 👍 

I also see your suggestions in the other ticket as well but if I understand correctly your #2 solution may not fix it all cases as "No data" data can happen at OverlayVariable as well. I will comment it too, just in case

Issue #699, opened 2021-11-13
Visualization: bug on custom legend marker color for Scatter plot


moontrip:
For the case that no overlay variable is selected, multiple legend items can still exist due to the plot options. For example, if choosing Smoothed mean with raw option is chosen, then the legend items are 'Data', 'Smoothed mean', and '95% Confidence interval'. In this case, the marker colors at the custom legend should have the same color, but it is currently shown as different color like the screenshot below

![scatter-viz-bug](https://user-images.githubusercontent.com/12802305/141661570-10c6c032-a61d-4ac8-aa1c-ceb936af1c1a.png)

aaronwlsong:
Seems the issue has been solved by **removing** markers ahead of label in customized legend box. Closing ticket. 

![image](https://user-images.githubusercontent.com/61077845/141812315-1f47d085-6961-4726-be50-f12369144c27.png)

aaronwlsong:
Thanks for your follow up comment @moontrip :)  yeah, I tried my qa. website in different ways (reload/ refresh, clear cache, etc.), still cannot get the markers as they are  in your screen shot. Thanks for checking it with Dave.

PR #700, opened 2021-11-14
fix a bug of legend marker color for Scatter plot Viz

moontrip:
This address a bug of custom legend marker color for Scatter plot Viz, #699 

Screenshot after fix is attached here, which is the same example shown at #699 

scatter-viz-bug-fix


bobular (approved):
Good catch!

jtlong3rd (approved):

moontrip:
Thank you for taking time to review/test this PR in the weekend, @bobular and @jtlong3rd 👍

PR #712, opened 2021-11-15
Temporary fix to single facet detection

bobular:
DK spotted that the `processedData` Dictionary was using `'undefined'` the string as the key, not undefined the value.

We should probably think of a better solution than this.

We could get real life data with 'undefined' values.


moontrip:
@bobular this solution works for the cases I reported at Slack 👍

moontrip (approved):
I didn't know you did set undefined as a key in the code 😄 Yes it would work as well 👍

jtlong3rd (approved):

bobular:
> I didn't know you did set undefined as a key in the code smile Yes it would work as well +1

I'm surprised (and annoyed) it didn't complain and just used the string version...

Issue #426, opened 2021-09-08
Add units to column header of data table in subsetting tab


SheenaTomko:
In the subsetting tab, for variables with units that appear as categorical data tables, we want the units to appear next to the variable name in the column header. Screenshot shows the units in with the main variable name, but missing from the column header in the table below

<img width="516" alt="Screen Shot 2021-09-08 at 10 42 09 AM" src="https://user-images.githubusercontent.com/54597319/132560563-b6774a9b-c2f5-4b6d-a9af-c473e6e92510.png">

sweverschulman:
units are now displayed in the column header of data table in subsetting tab and passes qa. 
<img width="284" alt="Screen Shot 2021-11-17 at 4 11 07 PM" src="https://user-images.githubusercontent.com/62298921/142283083-d2d8cb12-d0f7-406c-b3d3-901da91479fe.png">
b

PR #714, opened 2021-11-15
add unit for table filter's header

moontrip:
This will address #426 : thinking of a bit harder solution while Dave did let me know much simpler solution to directly change variable from parent 😄 Used pre-existing util to have units. Here is an example at GEMS1 Study control

table-filter-header-units


dmfalke (approved):
Nice!

jtlong3rd (approved):

PR #717, opened 2021-11-15
disable Plotly's default legend

moontrip:
With the new custom legend, we do not need to have the Plotly's default legend anymore. Thus it is now disabled except Mosaic plot Viz where it still relies on the Plotly's legend.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
@@ -447,9 +447,6 @@ function BoxplotViz(props: VisualizationProps) {
     }
   }, [data, legendItems]);
 
-  console.log('data at boxplot viz =', data);
-  console.log('legendItems = ', legendItems);
-
   const plotNode = (
     <BoxplotWithControls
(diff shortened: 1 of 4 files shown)


src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -463,11 +463,7 @@ function BoxplotViz(props: VisualizationProps) {
       }
       orientation={'vertical'}
       // add condition to show legend when overlayVariable is used
```
jtlong3rd, line 465: I guess we can remove this comment now!

jtlong3rd (approved):

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
moontrip: yep I removed it 👍 Also removed unnecessary console.logs that were found at the same component 😄

bobular (approved):
Looks great.  Nice to remove some of that ugly displayLegend logic.

Issue #671, opened 2021-11-03
Scatterplot visualization - dependent axis range resets temporarily between overlay variable changes


bobular:
Here is a GEMS1 visualization, which (for better or worse) uses the global variable-annotation's axis ranges 

![image](https://user-images.githubusercontent.com/308639/140192135-e1baca46-5fe5-441e-beec-e99ffcb021a8.png)

If you change the overlay variable, or toggle the "show missing" switch, the axes temporarily reset to plotly default autoscaling.  When the new data comes from the back end, the axes go back to global scaling.

Need to stop this happening somehow.  Should be low cost.

sweverschulman:
tested.

PR #720, opened 2021-11-16
Centralize loading of entity counts

### LINKED ISSUE #713: Visualizations: Bird's eye loads slow
This is likely due to entity count requests. This should be done at a central, ancestor component and made available to children.

dmfalke:
Uses the tried and test approach of "props passing" (and lots of it). I considered using `React.Context` to avoid passing props all over the place, but this approach is simple and works.

closes #713

### DIFF
--- a/src/lib/core/components/BirdsEyeView.tsx
+++ b/src/lib/core/components/BirdsEyeView.tsx
@@ -1,4 +1,3 @@
-import { Filter } from '../types/filter';
 import { EntityCounts } from '../hooks/entityCounts';
 import { CoverageStatistics } from '../types/visualization';
 import BirdsEyePlot from '@veupathdb/components/lib/plots/BirdsEyePlot';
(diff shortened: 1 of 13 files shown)


src/lib/core/components/BirdsEyeView.tsx
```diff
@@ -25,19 +27,17 @@ export function BirdsEyeView(props: Props) {
     completeCasesAxesVars,
     stratificationIsActive,
     enableSpinner = false,
+    totalCounts,
```
bobular, line 30: I think we can lose the `filters` prop now.  Will review and test properly later.  Just had a quick look!

bobular (approved):
This is much faster now. Nice :+1: 

We may want a root-and-branch review of plot and table update behaviours: e.g. spinner/no-spinner, blank out (but keep screen space) when not current data (see also #671), or just silently update.

Currently there's a silent update of the birds eye plot if you cancel the filter (using global filters modal) while on a viz page.  The entity diagram pink bars have always silently updated (with no spinner or other indication) so this is probably fine, but a review would be good for later in Phase 2 maybe?

dmfalke:
Thanks, @bobular. I agree a review would be good. I will remove some of the unused props and then merge. Thanks!

Issue #681, opened 2021-11-10
Faceting: Plot components should render title/facet value outside of Plotly and render "no data" message for empty data


bobular:
Two types of empty plot:

1. Skeleton: what we currently see when starting a new viz - rendered by Plotly.
     - when PlotProps.data is "empty"
2. No data: should be a large grey "No data" in middle of where the plot would normally be (not rendered by Plotly)
     - when PlotProps.data is undefined

Start with barplot in storybook and discuss.

Old description:

Facet title should be done in regular HTML/React, instead of using the Plotly title.

Elements of FacetedData.facets that have `data: undefined` would render the "no data" message (maybe "No data for this facet") instead of an empty PlotlyPlot

```
facets:
  [
   {
      label: 'India',
      data: undefined,
    }
  ] 
```

I also wonder if we need a new root level `facetVariableDisplayName` or similar prop on FacetedData.  This would be useful for FacetedPlot to label the "facet axis".

chowington:
@SheenaTomko See https://github.com/VEuPathDB/web-components/pull/262 for more details for this ticket's requirements

SheenaTomko:
Looked at faceting in all plot types (scatter, box, bar, histogram, 2x2, rxc), choosing variables to stratify (overlay and facet) by with some missing data, both due to subsetting and due to the data naturally being incomplete. Toggled on and off the no data toggle. Saw generally that if you subsetted based on a variable then used it as a facet, you would get an empty plot for the value(s) you removed with a light grey overlay that said faintly "No data", even without turning on the toggle. If you faceted by a variable which naturally had missing data and you turned on the toggle, then you would get a "No data" facet plot. Looks good!

PR #721, opened 2021-11-16
Handle empty facets as `facet.data == undefined`

### LINKED ISSUE #706: Visualizations / Faceting: ensure all vizs provide empty plots for empty facets
I noticed that faceting Mosaic plots on Country (where country is filtered in subsetting to just one country) looks like a no-op (there's only one facet, so the facet layout handling is not triggered).  There should be mostly empty plots.

Should coordinate with @chowington who is working on https://github.com/VEuPathDB/web-eda/issues/681 - we discussed how to deal with it [here](https://docs.google.com/document/d/1W1iVvO0n29E2ZDN3M8VRoctg29voNOm-td8xLAxcGSA/edit?pli=1#bookmark=id.8jbeg43mqgqg) but I think we concluded the opposite to what we wrote down :grimacing: 

The plot components already test for "empty data" (usually `data.series.length === 0`) and render a skeleton.  However, the individual components all use 'empty data' as a fallback value for the `data` prop.  [Example here](https://github.com/VEuPathDB/web-components/blob/main/src/plots/Histogram.tsx#L92).

We were going to make the plot component (probably at PlotlyPlot.tsx) render its own HTML title and render a no-data message instead of the skeleton plot. We may need to change the fallback handling of `data` as described above.  I think these changes would be too much for a last minute b55 push.

I'll keep thinking about how best to do this in discussion with @chowington

bobular:
Resolves #706 
Depended on some trivial work in https://github.com/VEuPathDB/web-components/pull/260 that may impact @chowington's work on https://github.com/VEuPathDB/web-eda/issues/681

Doesn't need four reviews, but I thought you all might be interested.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
@@ -54,17 +54,7 @@ import {
 } from '../../../utils/visualization';
 import { VariablesByInputName } from '../../../utils/data-element-constraints';
 // use lodash instead of Math.min/max
-import {
-  max,
-  sortBy,
-  groupBy,
-  mapValues,
-  size,
-  map,
-  head,
-  values,
-  keys,
-} from 'lodash';
+import { max, groupBy, mapValues, size, map, head, values, keys } from 'lodash';
 import { isFaceted } from '@veupathdb/components/lib/types/guards';
 // for custom legend
 import PlotLegend, {
@@ -668,18 +658,24 @@ function reorderData(
 ): BarplotDataWithStatistics | BarplotData {
   // If faceted, reorder the facets and within the facets
   if (isFaceted(data)) {
+    // for each value in the facet vocabulary's correct order
+    // find the index in the series where series.name equals that value
+    const facetValues = data.facets.map((facet) => facet.label);
+    const facetIndices = facetVocabulary.map((name) =>
+      facetValues.indexOf(name)
+    );
+
     return {
       ...data,
-      facets: sortBy(data.facets, ({ label }) =>
-        facetVocabulary.indexOf(label)
-      ).map(({ label, data }) => ({
-        label,
+      facets: facetIndices.map((i, j) => ({
+        label: facetVocabulary[j],
         data:
-          data != null
+          data.facets[i]?.data != null
             ? (reorderData(
-                data,
+                data.facets[i].data!, // not sure why ! is needed
                 labelVocabulary,
-                overlayVocabulary
+                overlayVocabulary,
+                facetVocabulary
               ) as BarplotData)
             : undefined,
       })),
--- a/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
@@ -309,9 +309,9 @@ function BarplotViz(props: VisualizationProps) {
     if (isFaceted(data?.value)) {
       return data?.value?.facets != null
         ? max(
-            data.value.facets.flatMap((facet) =>
-              facet.data.series.flatMap((o) => o.value)
-            )
+            data.value.facets
+              .filter((facet) => facet.data != null)
+              .flatMap((facet) => facet.data?.series.flatMap((o) => o.value))
           )
         : undefined;
     } else {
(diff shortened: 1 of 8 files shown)


src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -667,17 +658,26 @@ function reorderData(
...
-          data,
-          labelVocabulary,
-          overlayVocabulary
-        ) as BarplotData,
+      facets: facetIndices.map((i, j) => ({
+        label: facetVocabulary[j],
+        data:
+          data.facets[i]?.data != null
+            ? (reorderData(
+                data.facets[i].data!, // not sure why ! is needed
```
bobular, line 675: I'd appreciate any wisdom on this!

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -465,6 +477,7 @@ function BoxplotViz(props: VisualizationProps) {
       // show/hide independent/dependent axis tick label
       showIndependentAxisTickLabel={true}
       showDependentAxisTickLabel={true}
+      dependentAxisRange={dependentAxisRange}
```
bobular, line 480: Here is the fix for the lost prop!

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -492,11 +494,12 @@ function ScatterplotViz(props: VisualizationProps) {
                 ? true
                 : false
               : allData.facets
+                  .filter((facet) => facet.data != null)
                   .map(
                     ({ data }) =>
-                      data.series[index]?.y != null &&
-                      data.series[index].y.length > 0 &&
-                      data.series[index].y[0] !== null
+                      data?.series[index]?.y != null &&
```
bobular, line 500: I already filtered for facet.data not null.  Not sure why the `?` is still needed.

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
dmfalke: I don't think typescript can guarantee that `i` hasn't been mutated since the not-null assertion on the line above, so it can't assume the expression still evaluates to not-null.

You can avoid the `!` assertion by storing `data.facets[i].data` in a variable.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
dmfalke: There is a long backstory to this, but you can you a type guard annotation in the `filter` callback to prevent this. It would basically look like this:

```typescript
.map(facet => facet.data)
.filter((data): data is D => data != null)
```

where `D` is the non-null type of `facet.data`.

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
bobular: Thanks! That works great, and makes it more readable.

bobular:
OK, thanks for the feedback @dmfalke
This is good for a final once-over or testing please (anyone)

jtlong3rd (approved):

moontrip:
@bobular I am testing some cases and find one issue for boxplot so far. It seems like the dependent axis range you have changed does not work correctly: more exactly, setting props dependentAxisRange at BoxplotWithControls seems to resolve the original issue at main branch, however, your changes on dependentAxisRange computation on this PR do not work, perhaps due to the .filter() (?), so please check it again. Here are two screenshots.

# at main branch (applying props dependentAxisRange at BoxplotWithControls temporarily)

boxplot-main

# at this branch
boxplot-error-case0

moontrip:
@bobular IDK why but I cannot reproduce the issue I reported above anymore. It would be great if you can test it at your side and see if it happens. If not, please ignore my comment 😄

dmfalke (approved):
Just skimmed the code, but don't see anything abhorrent, so all good by me :slightly_smiling_face:

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
bobular: Hi @moontrip  - the prop is definitely here.  Must have been some cache issue on your side?

moontrip:
@bobular yeah I suspect a sort of cache issue from my side when I tested it. Sorry to bother you for incorrect test result :(

bobular:
Yeah I can't replicate
BMI is a good y-axis var because of the outlier in Mozambique
image

bobular:
Thanks for the testing and reviews, folks.

PR #732, opened 2021-11-17
fix dependent axis range reset for scatter plot

moontrip:
This will address [Scatterplot visualization - dependent axis range resets temporarily between overlay variable changes #671](#671 ).

I thought the source of the problem might be complicated and was investigating this too deep. However, the actual reason turned out to be pretty simple: the dependentAxisRange had a condition to check data.pending 😐 Anyway, now it seems to work as expected without changing the range.


jtlong3rd (approved):

bobular (approved):
Yeah that's great.  I had another easy fix like this a while back (I think it was the bin slider clunkiness).

It turns out that `usePromise` keeps the previous value and makes things like this nice and smooth.

PR #733, opened 2021-11-17
Thumbnail improvements for B55

### LINKED ISSUE #716: Visualizations - faceted thumbnails should have "X Facets" overlaid in a box with a background
Suggestions during the meeting were to have the box at the bottom of the thumbnail - looking a bit like a tab label?

I guess a semi-opaque white background would look good.

### LINKED ISSUE #718: Faceting thumbnails go mostly blank when visiting a visualization and making no changes
So make a visualisation with faceting (I used histogram, but I don't think it matters).

Then go back to the visualizations 'home' with the thumbnails.  It looks nice.

Then go back to the same visualization, and admire it some more.  **But don't make any modifications.**

Then go back to the thumbnails. It's just a shadow of its former self! :cry: 

image

jtlong3rd:
Resolves #716 and resolves #718.

More specifically, this PR improves the thumbnails for faceted plots in the following ways:

* The facet count has been styled as per the spec in #716.
* Introduces a short-term fix for thumbnails going blank under certain sporadic circumstances.

Some followup work for B56+:

* The content / verbiage of faceted thumbnails will be revisited in a future UX meeting. (One proposal is to only display a single plot.)
* #718 is a symptom of a bigger issue with how we're generating thumbnails: there is a race condition between the effect which triggers the thumbnail capture (managed by React) and the drawing of the plot (managed by Plotly "outside" of the React lifecycle). In this PR, I have mitigated this race condition by "rigging the race" in favor of Plotly (by making the thumbnail capture a [layout effect](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) and introducing a `setTimeout`). In the long term, we'll want to eliminate the race condition by moving the thumbnail capturing from a React-managed effect to a Plotly-managed `onUpdate` callback. (I opened a tech debt issue [here](https://github.com/VEuPathDB/web-eda/issues/739).)

### DIFF
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -30,6 +30,7 @@
 
     img {
       width: 100%;
+      height: 100%;
       text-align: center;
     }
 
--- a/src/lib/core/hooks/thumbnails.ts
+++ b/src/lib/core/hooks/thumbnails.ts
@@ -1,4 +1,10 @@
-import { DependencyList, useCallback, useEffect, useRef } from 'react';
+import {
+  DependencyList,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+} from 'react';
 
 import { FacetedPlotRef, PlotRef } from '@veupathdb/components/lib/types/plots';
 import { Task } from '@veupathdb/wdk-client/lib/Utils/Task';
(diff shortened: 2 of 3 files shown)


bobular (approved):
This looks really smart now!

I did see an issue with the Histogram (the gap/size issue) - my guess is that the aspect ratio of the individual histograms is not the same as for the other vizs.

image


PS. boxplot wasn't in my tests - it is fine :-) 

image

jtlong3rd:
> This looks really smart now!
> 
> I did see an issue with the Histogram (the gap/size issue) - my guess is that the aspect ratio of the individual histograms is not the same as for the other vizs.
> 
> image
> 
> PS. boxplot wasn't in my tests - it is fine :-)
> 
> image

Thanks for reviewing, @bobular! I'll merge now, and we can address the histogram issue as followup work. (As you said, a different aspect ratios is likely in play - my various iterations of "Faceted" / "X Facets" labels all looked different for histogram.)

PR #734, opened 2021-11-17
Fix broken Birds Eye Plot tooltip with a unified new one

bobular:
Requires https://github.com/VEuPathDB/web-components/pull/267
Issue is in web-components: https://github.com/VEuPathDB/web-components/issues/263

> Reviewer(s) - please self-assign and approve trivial components PR when you review. Thanks! 

It was impossible to fix the built-in mouseover of the Plotly-based Birds Eye Plot, so I implemented it on the EDA side, but putting the numbers into the WDK tooltip, and making the whole birds eye plot trigger it (quicker than the default time too).

I hope Outreach like the solution!
image

I also fixed the top-alignment that irked Steve.

Note that the new layout requires a minimum of around 1430 horizontal pixels to keep everything in one row.

When the legend was inside the main plot, we managed to fit in into 1280px (target minimum width)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@mui/styles": "^5.0.1",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.10.2",
+    "@veupathdb/components": "^0.10.5",
     "@veupathdb/core-components": "^0.2.45",
     "@veupathdb/http-utils": "^1.0.1",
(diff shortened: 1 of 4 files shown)


jtlong3rd (approved):
Makes sense to me!

bobular:
Thanks Jamie. I'll merge tomorrow.
If anyone needs to merge before then, be my guest.

bobular:
Could possibly remove the word 'that' from the first three definitions.
Please :+1: if you agree.

dmfalke (approved):

PR #738, opened 2021-11-18
Faceted Layout for Mosaic Visualizations

### LINKED ISSUE #725: Finalise faceted layout of Mosaic visualizations
There were various discussion about vertical/horizontal layout of plots and tables.  I think we should be pragmatic and not do too much, given that we will be faceting in two dimensions in the not too far future.

We should at least get rid of the yellow placeholder (below) and @jtlong3rd mentioned having to do some refactoring to work around the oddities of this visualization, so I will assign to him, **provisionally**.

image

jtlong3rd:
(Probably don't need everybody to review the code, but testing / UX feedback would be welcome!)

Depends on https://github.com/VEuPathDB/web-components/pull/268

Resolves #725 

This PR uses the new `PlotLayout` component to handle the unfaceted and faceted layout for the Mosaic visualization. For the faceted layout, the tabs take up the full width of the visualization container, and in the `Table` and `Statistics` tables, the table are laid out horizontally, just as the corresponding mosaic plots are.

**Future work:**

In this visualization, our "data" tables are being styled by a mix of CSS and inline styling. We should probably unify these two batches of styles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@mui/styles": "^5.0.1",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.10.6",
+    "@veupathdb/components": "^0.10.7",
     "@veupathdb/core-components": "^0.2.45",
     "@veupathdb/http-utils": "^1.0.1",
     "@veupathdb/study-data-access": "^0.0.3",
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -195,15 +195,6 @@ $data-table-inner-border: 1px solid #888;
   }
 }
 
-.viz-plot-info {
-  display: grid;
-  grid-auto-flow: row;
-  gap: 0.75em;
-  margin-left: 3em;
-  margin-top: 1.5em;
-  margin-left: 3em;
-}
-
 .VariableCoverageTable {
   @include data-table;
   width: 400px;
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -217,6 +217,11 @@ $data-table-inner-border: 1px solid #888;
   @include data-table;
   overflow: auto;
 
+  th {
+    max-width: 150px;
+    white-space: normal;
+  }
+
   .contingency-table_column-label {
     border-bottom: $data-table-inner-border !important;
   }
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -152,9 +152,6 @@
       font-style: italic;
       margin-bottom: 1em;
     }
-    .MosaicVisualization-Plot .web-components-plot {
-      margin-left: 0;
-    }
   }
   &-FullScreenActions {
     display: flex;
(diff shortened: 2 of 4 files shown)


bobular (approved):
This looks good.

I think we may need to wrap long variable values in the table headers so they don't take up a silly amount of horizontal space
image

So with adding `maxWidth: '150px'` to `.contingency-table th` (in dev tools but I think it's line 186 of Visualizations.scss) it looks a bit better IMO:
image

I'm not personally keen on the mismatch between the plot titles just having the value ("India") and the table titles being "Country: India", but maybe we were asked to do this!

Do we want "No data for this facet" when there is no table? I see that annoyingly this is half in web-components (ContingencyTable, where there are complications with the use of a Spinner when there is no data) and half in web-eda (ContTableStats and TwoByTwoStats). So let's wait until we get some clear feedback and specifications.

Is there a trivially easy way for the horizontal scrolling position to be "remembered" between changes of the "Mosaic/Table/Statistics" tabs?

jtlong3rd:
> I think we may need to wrap long variable values in the table headers so they don't take up a silly amount of horizontal space image
> 
> So with adding `maxWidth: '150px'` to `.contingency-table th` (in dev tools but I think it's line 186 of Visualizations.scss) it looks a bit better IMO: image
> 

I agree - constraining the table headers to a max width of 150px looks much better.

jtlong3rd:
> Do we want "No data for this facet" when there is no table? I see that annoyingly this is half in web-components (ContingencyTable, where there are complications with the use of a Spinner when there is no data) and half in web-eda (ContTableStats and TwoByTwoStats). So let's wait until we get some clear feedback and specifications.

Yes, we definitely want some sort of explicit "no data for this facet." Let's wait for feedback + clear requirements.

jtlong3rd:
> I'm not personally keen on the mismatch between the plot titles just having the value ("India") and the table titles being "Country: India", but maybe we were asked to do this!

I'm also not keen on the mismatch. I'm pretty sure that during the dev meeting this Wednesday, I heard Ann making a comment like "Weren't we going to give the plots titles like `Country: India`?", but don't think we ever circled back to it. Am I remembering that correctly, @asizemore, or am I completely making that up?

asizemore:
No, @jtlong3rd , you're totally right! That format my preference and i agree i don't think we ever 100% decided. 

While we could get away with the title being "India" for this phase, my guess is that we'll move to something like "Country: India" in the near future. Having a "Country: India" style (something that says the variable name then variable value) will be important once we add the ability to have a second faceting variable and it's essential to adding the wrapping facets feature.

asizemore:
Regardless of my thoughts above, I agree with you both that keeping it consistent is most important. I'm fine with having them all say "India" for this release. I don't think it would be too jarring to have plot titles of "India" now and then for the next release when we add a second facet they'll say "Country: India \n Shoe size: 8", or something along those lines :)

jtlong3rd:
> Is there a trivially easy way for the horizontal scrolling position to be "remembered" between changes of the "Mosaic/Table/Statistics" tabs?

Only "non-trivial, low-medium cost" solutions are coming to mind, so let's make an issue and revisit soon. (One possibility would involve retaining the scroll positions in a `MosaicVisualization`-maintained ref, and introducing a layout effect which, on tab change, scrolls to the previous position.)

jtlong3rd:
Thanks for reviewing, @bobular! I pushed "give contingency table headers a max width of 150px." I'll merge now, and then we can make issues for the followup work you and @asizemore identified:

1. Make the facet titles consistent between tabs (I share @bobular and @asizemore's preference for `Country: India` over `India`.)
2. Remember the tab scrolling position 
3. Add "no data for this facet" for facets where no table is available

Issue #741, opened 2021-11-19
Visualization: a custom legend bug for Scatter faceted plot


moontrip:
While thinking and checking of a potential scatter plot R-square table with a couple of example cases, I found that there was a bug for for some cases of faceted plots. 

For an example like below with the plot options of Best fit with raw (or Smoothed mean with raw), the custom legend shows 5 Raw and 3 Best fit, however, it should show 5 Best fit legend items. 

![scatter-facet-bug](https://user-images.githubusercontent.com/12802305/142668926-079faa21-bab6-4876-8641-365d3186b186.png)

bobular:
Hi @moontrip - I was going to say are you sure this is a bug?  Maybe the other best fit lines aren't shown because there isn't enough data to calculate them, but there is, and you mentioned that also in the PR (the back end response contains all five best fit line reponses).

Good catch!

moontrip:
@bobular yes initially I was suspected the case you mentioned, but for the example case, I could see that the backend response contained all five best fit data, distributed in countries though :)

sweverschulman:
The bug has been corrected and only showing either best fit or smooth mea
<img width="1096" alt="Screen Shot 2021-11-22 at 1 25 36 PM" src="https://user-images.githubusercontent.com/62298921/142916559-9e8cafdd-e9ee-4b15-a7c8-ed6764c57a9b.png">
n
<img width="1043" alt="Screen Shot 2021-11-22 at 1 25 56 PM" src="https://user-images.githubusercontent.com/62298921/142916566-95e9b33c-c798-47d5-b33c-b8ee70240cea.png">

PR #742, opened 2021-11-19
fix a legend bug for scatter faceted plot

moontrip:
While thinking and checking of a potential scatter plot R-square table with a couple of example cases, I found that there was a bug for some specific data for the case of faceted plots. 

Just made a new ticket so this PR will address #741 

For an example like below with the plot options of Best fit with raw (or Smoothed mean with raw) using GEMS1 Study control, the custom legend showed 5 Raw and 3 Best fit. It is because the condition to find the legend list is based on the first non-undefined data. However, when I checked whole faceted data, it is found that actually all 5 Best fits exist throughout countries, thus it is necessary to find whole 5 Best fits by going through whole data, like below screenshot. While fixing the issue, it was inevitable to use complex conditions to properly order the legend items. Perhaps that part may be revisited for better logic/scheme in the future, especially considering the B55 release date, but at least this PR would work.

# before fix
scatter-facet-bug

# after fix
scatter-bug-fix

### DIFF
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -53,6 +53,7 @@ import {
   mapValues,
   map,
   keys,
+  uniqBy,
 } from 'lodash';
 // directly use RadioButtonGroup instead of XYPlotControls
 import RadioButtonGroup from '@veupathdb/components/lib/components/widgets/RadioButtonGroup';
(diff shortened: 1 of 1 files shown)


moontrip:
Since the freezing will happen today, I listed @dmfalke as well in case @bobular is not available. Certainly @jtlong3rd is also a reviewer too 😄

bobular (approved):
Hi DK - great catch and good solution.  Though I will admit I didn't spend the time to fully understand the new sorting logic.

What I did instead, which may reduce technical debt in the future, is get rid of the `as any[]` and used a trick that Dave showed me (adding the type guard to the `!= null` filter).  Given that `uniqBy` only cares about the `name` prop, I think it's fine to flatten all the facet series "whole" pulling out just the `{ name, mode, fill }` props. That avoids the TS issues.

moontrip:
@bobular yes I just saw your commits. I was also thinking of the case to use full data, but actually for legend, the three object keys are required so I just remade object containing the three keys/values only to save some memory. But it may not be that optimal, and your change looks great! 👍 

Thank you for your reviews and comments!

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -423,23 +424,40 @@ function ScatterplotViz(props: VisualizationProps) {
...
+
     const legendData = !isFaceted(allData)
       ? allData?.series
-      : allData.facets.find(
-          ({ data }) => data != null && data.series.length > 0
-        )?.data?.series;
+      : uniqBy(
+          allData.facets
+            .map((facet) => facet?.data?.series)
+            .flat()
```
jtlong3rd, line 442:
```suggestion
            .flatMap((facet) => facet?.data?.series)
```

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -449,9 +467,118 @@ function ScatterplotViz(props: VisualizationProps) {
...
+    let count = 0;
+
+    // new array of objects based on legendData array (a kind of partial sort)
+    // raw data is in the correct order, but not always for smoothed mean, CI, and Best fit
+    // e.g., showing 'data1' (raw), 'data2' (raw), 'data2, Best fit', 'data1, Best fit' in order
+    // needed to have complex conditions for treating Smoothed mean, CI, Best fit, No data, etc.
+    const sortedLegendData =
+      isFaceted(allData) && vizConfig.valueSpecConfig !== 'Raw'
+        ? legendData
+            ?.map((dataItem, index) => {
```
jtlong3rd, line 485:
```suggestion
            ?.flatMap((dataItem, index) => {
```

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -449,9 +467,118 @@ function ScatterplotViz(props: VisualizationProps) {
...
+                        return element.name?.includes('No data' + CI95SUFFIX);
+                      });
+                    }
+                  }
+                }
+              }
+              return [];
+              // observed No data is often undefined during data loading by toggling showMissingness for large scatter data/graphics
+            })
+            .flat()
```
jtlong3rd, line 575:
```suggestion
```

jtlong3rd (approved):
Just sneaking in the usual small suggestions before we merge. 😄

moontrip:
@jtlong3rd Thank you for your reviews! 👍 Yes, map( ) + flat() = flatmap() 😄 Applied your changes

@dmfalke since Bob and Jamie reviewed this PR, I will merge this.

PR #746, opened 2021-11-22
Multiple mui instances

### LINKED ISSUE #745: Faceted plot panel doesn't occupy 100% width and have browser's horizontal scroll bar when deployed on QA
Works fine in my dev instance, but not on QA.

I don't think screenshots are required ;-)

dmfalke:
fixes #745

This PR removes an import of a different version of `makeStyles`, which was causing classname collisions.

I will merge this before end of day (us/eastern) so that the qa site can be have the fix tomorrow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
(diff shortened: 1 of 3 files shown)


bobular (approved):
Looks good to me.  I can't test it on a dev instance, right?

dmfalke:
Not trivially, unfortunately.

Issue #747, opened 2021-11-22
Bug in scatterplot faceting with data changing based on application of best fit or smoothed means


SheenaTomko:
I was trying to qc https://github.com/VEuPathDB/web-components/pull/262 and based my selections off one of Bob's examples.
In GEMS, I set study site to be Kenya, Mali, Mozambique, and the Gambia. In Scatterplot, I chose the following parameters:
<img width="1178" alt="Screen Shot 2021-11-22 at 1 11 44 PM" src="https://user-images.githubusercontent.com/54597319/142921492-33e76f28-7611-4fe7-a7a9-515ad028e24a.png">

Data would appear/disappear from facets depending on what plot mode I chose. 

Raw
<img width="1220" alt="Screen Shot 2021-11-22 at 1 11 21 PM" src="https://user-images.githubusercontent.com/54597319/142921624-e3ba3cac-08b3-4526-901c-970b2e4da683.png">

Smoothed mean
<img width="1229" alt="Screen Shot 2021-11-22 at 1 11 31 PM" src="https://user-images.githubusercontent.com/54597319/142921642-7d845798-1cbf-4f52-9670-9e03d4df2254.png">

Best fit
<img width="1228" alt="Screen Shot 2021-11-22 at 1 11 07 PM" src="https://user-images.githubusercontent.com/54597319/142921669-fbe73e11-bd7c-4b49-9bb1-4b72f96ca109.png">

dmfalke:
This issue is caused by a limit that Chrome imposes on the number of webgl instances a browser instance can have. This is across all tabs and windows of the browser instance.

We are using the webgl implementation of scatter plot since it can handle a larger number of data points. We currently limit the number of data points to 100,000.

@moontrip has proposed using the non-webgl implementation when faceting.

moontrip:
Tested the qa.restricted just now, and it works fine for me. Thanks @dmfalke for updating the site fast! 👍

SheenaTomko:
Looks good

PR #748, opened 2021-11-22
use scatter type for scatter with facet

moontrip:
This will possibly address #747 

A solution is to use type: scatter for faceted plot, but use type: scattergl for non-faceted plot. Also needed to consider opacity of confidence intervals as two different ways.

Here is a screenshot with the same example reported in #747 : also tested with other Plot options like Smoothed mean with raw and Best fit with raw and they worked fine too

- after change
scatter-faceted-fix1


jtlong3rd (approved):

dmfalke (approved):
I think this looks great. I'll go ahead and merge so it can be available on the qa site tomorrow.

moontrip:
@jtlong3rd and @dmfalke Thank you for your reviews! 👍 And thank you for handling merge on my behalf, Dave!

Issue #694, opened 2021-11-12
Visualizations: Scatterplot needs table for best fit R^2 values, especially when faceting


bobular:
Related to #690 

When faceting, the series names must be identical for each facet so that the custom legend can switch them on/off correctly.

So we removed the R^2 from the legend, but it will need to go somewhere.

Theoretically, without faceting the R^2 *can* be displayed in the legend, but it may be more consistent to have it in the same place regardless of faceting.

PR #749, opened 2021-11-23
R-square table for Scatter plot with Best fit option

moontrip:
This will possibly address #694. Also, this depends on the [PR #274 at web-components](https://github.com/VEuPathDB/web-components/pull/274) to avoid a type error. Certainly this needs to be refined with layout etc. so this will be a draft PR for the time being. 

This PR adds a R-square table for Scatter plot with the plot options of 'Best fit line with raw'. For now, the new table is added as a last item of `tableGroupNode` as shown in the screenshot. Since faceted plot with overlay can possibly be very lengthy, maximum height is set and scroll bar is used.

@jtlong3rd I will leave this to you for layout etc. as I am not quite sure how we will show this table 😄 

- No overlay and no facet
scatter-r2-noOverlay-noFacet

- With overlay but no facet
scatter-r2-overlay-noFacet

- No overlay but with facet
scatter-r2-noOverlay-facet

- Both overlay and facet
scatter-r2-overlay-facet

### DIFF
--- a/src/lib/core/components/ScatterplotRsquareTable.tsx
+++ b/src/lib/core/components/ScatterplotRsquareTable.tsx
@@ -1,0 +1,134 @@
+import {
+  XYPlotDataSeries,
+  XYPlotData,
+} from '@veupathdb/components/lib/types/plots';
+import { VariableTreeNode } from '../types/study';
+
+interface facetDataProps {
+  label: string;
+  data?: XYPlotData | undefined;
+}
+
+export interface Props {
+  data: XYPlotDataSeries[] | facetDataProps[] | undefined;
+  isFaceted: boolean;
+  overlayVariable?: VariableTreeNode;
+  facetVariable?: VariableTreeNode;
+}
+
+export function ScatterplotRsquareTable({
+  data,
+  isFaceted,
+  overlayVariable,
+  facetVariable,
+}: Props) {
+  // non-facet or facet data
+  const filteredData = !isFaceted
+    ? overlayVariable != null
+      ? (data as XYPlotDataSeries[])?.filter((data) =>
+          data?.name?.includes(', Best fit')
+        )
+      : (data as XYPlotDataSeries[])?.filter((data) =>
+          data?.name?.includes('Best fit')
+        )
+    : (data as facetDataProps[]).filter((element) => element.data != null);
+
+  if (!isFaceted) {
+    return (
+      <div className={'ScatterRsquareTable'}>
+        <table>
+          <tbody>
+            <tr>
+              <th>
+                {overlayVariable != null ? overlayVariable.displayName : 'Name'}
+              </th>
+              <th className="numeric">
+                R<sup>2</sup> (Best fit)
+              </th>
+            </tr>
+            {filteredData != null
+              ? (filteredData as XYPlotDataSeries[]).map((data) => (
+                  <tr>
+                    <td>{data?.name?.split(', Best fit')[0]}</td>
+                    <td>{data.r2 ?? 'N/A'}</td>
+                  </tr>
+                ))
+              : ''}
+          </tbody>
+        </table>
+      </div>
+    );
+  } else {
+    return (
+      <div
+        className={'ScatterRsquareTable'}
+        style={{ maxHeight: 250, overflowX: 'hidden', overflowY: 'auto' }}
+      >
+        <table>
+          <tbody>
+            <tr>
+              <th>{facetVariable?.displayName}</th>
+              {overlayVariable != null && (
+                <th>{overlayVariable.displayName}</th>
+              )}
+              <th className="numeric">
+                R<sup>2</sup> (Best fit)
+              </th>
+            </tr>
+            {filteredData != null
+              ? (filteredData as facetDataProps[]).map((data) => (
+                  <>
+                    {data?.data?.series != null
+                      ? data.data.series
+                          .filter((series) =>
+                            overlayVariable != null
+                              ? series?.name?.includes(', Best fit')
+                              : series?.name?.includes('Best fit')
+                          )
+                          .map((series, index, array) => {
+                            if (index === 0) {
+                              return (
+                                <tr>
+                                  {/* each vocabulary/name have different number of available data ,so need to check rowSpan per data */}
+                                  <td
+                                    rowSpan={
+                                      overlayVariable != null
+                                        ? array.filter((arr) =>
+                                            arr?.name?.includes(', Best fit')
+                                          ).length
+                                        : 1
+                                    }
+                                  >
+                                    {data.label}
+                                  </td>
+                                  {overlayVariable != null && (
+                                    <td>
+                                      {series?.name?.split(', Best fit')[0]}
+                                    </td>
+                                  )}
+                                  <td>{series.r2 ?? 'N/A'}</td>
+                                </tr>
+                              );
+                            } else {
+                              return (
+                                <tr>
+                                  {overlayVariable != null && (
+                                    <td>
+                                      {series?.name?.split(', Best fit')[0]}
+                                    </td>
+                                  )}
+                                  <td>{series.r2 ?? 'N/A'}</td>
+                                </tr>
+                              );
+                            }
+                          })
+                      : ''}
+                  </>
+                ))
+              : ''}
+          </tbody>
+        </table>
+      </div>
+    );
+  }
+}
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -274,3 +274,20 @@ $data-table-inner-border: 1px solid #888;
     visibility: hidden;
   }
 }
+
+// R-square table for Scatter plot's Best fit option
+.ScatterRsquareTable {
+  @include data-table;
+  width: 400px;
+
+  .percentage {
+    color: #333;
+    font-size: 0.85em;
+  }
+  th {
+    vertical-align: baseline;
+  }
+  th:first-of-type {
+    border-bottom: $data-table-inner-border;
+  }
+}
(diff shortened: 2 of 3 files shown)


moontrip:
Update: the [PR #274 at web-components](https://github.com/VEuPathDB/web-components/pull/274) is now merged so its new version is updated in this PR: no need to use `yarn link` for test this PR.

src/lib/core/components/ScatterplotRsquareTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+              </th>
+              <th className="numeric">
+                R<sup>2</sup> (Best fit)
+              </th>
+            </tr>
+            {filteredData != null
+              ? (filteredData as XYPlotDataSeries[]).map((data) => (
+                  <tr>
+                    <td>{data?.name?.split(', Best fit')[0]}</td>
+                    <td>{data.r2 ?? 'N/A'}</td>
```
jtlong3rd, line 53: This line is giving me a TypeScript error - `Property 'r2' does not exists on type 'XYPlotDataSeries'`.

I'm guessing you'll need to introduce some new types in `ScatterplotVisualization.tsx`, perhaps something like this:

```
interface XYPlotDataSeriesWithRSquare extends XYPlotDataSeries {
  r2?: number;
}

interface XYPlotDataWithRSquare {
  series: XYPlotDataSeriesWithRSquare[];
}
```

You'd then have to update `XYPlotDataWithCoverage` to something like `XYPlotDataWithRSquareAndCoverage`:

```
// define XYPlotDataWithRSquareAndCoverage
interface XYPlotDataWithRSquareAndCoverage extends CoverageStatistics {
  dataSetProcess: XYPlotDataWithRSquare | FacetedData<XYPlotDataWithRSquare>;
  // change these types to be compatible with new axis range
  yMin: number | string | undefined;
  yMax: number | string | undefined;
}
```

Let me know if you have another idea!

src/lib/core/components/ScatterplotRsquareTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+export function ScatterplotRsquareTable({
+  data,
+  isFaceted,
+  overlayVariable,
+  facetVariable,
+}: Props) {
+  // non-facet or facet data
+  const filteredData = !isFaceted
+    ? overlayVariable != null
+      ? (data as XYPlotDataSeries[])?.filter((data) =>
```
jtlong3rd, line 28: I have a few ideas for getting rid of these type assertions. Because my ideas partially depend on how you decide to resolve the TypeScript error I pointed out, I'll wait until you've addressed the TypeScript error, and then explain more.

jtlong3rd (commented):
Off to a great start!

I think the layout is fine for a first iterate - adding the new table to the end of the `tableGroupNode` is exactly what I would have done.

src/lib/core/components/ScatterplotRsquareTable.tsx
moontrip: @jtlong3rd Thank you for your comment and suggestions! 👍 I am on vacation today but I found your nice explanation. Actually it is quite strange as I have defined it at XYPlotDataSeries as you approved the PR, [#274 at web-components](https://github.com/VEuPathDB/web-components/pull/274),  and updated the web-components version at this PR. Can you please try `git pull origin scatter-r2-table` and then `yarn install` ? I don't have type error.

I may have more time to check this tomorrow if you would still have the issue. Thanks in advance!

jtlong3rd: @moontrip Crisis averted! I was `yarn link`ed to an old version of web-components that didn't have `r2` in `XYPlotData`. All I had to do was `yarn unlink`.

jtlong3rd (commented):
Hi @moontrip, as promised, here's a possible approach for excising the type assertions. Let me know if you have any questions!

moontrip:
Thank you for handling the type assertion! @jtlong3rd 👍 👍 I will take care of them tomorrow and it would certainly be easy to handle with your great explanations.

moontrip:
@jtlong3rd I have applied your suggestions 👍 It took more time than I expected due to a minor bug I found and missed the hidden conversation in this PR. Anyway it works now 😄

jtlong3rd (approved):
Glad the rehaul worked out!

One more thing: could you fix the `Warning: Each child in a list should have a unique "key" prop.` console [warnings](https://reactjs.org/docs/lists-and-keys.html#keys) that `ScatterplotRsquareTable.tsx` is triggering?

moontrip:
> Glad the rehaul worked out!
> 
> One more thing: could you fix the `Warning: Each child in a list should have a unique "key" prop.` console [warnings](https://reactjs.org/docs/lists-and-keys.html#keys) that `ScatterplotRsquareTable.tsx` is triggering?

@jtlong3rd Oh, I didn't notice that error. Thanks, I will try to fix it.

moontrip:
@jtlong3rd I am puzzled why the unique key does not work. So, for the case of no faceted data (X/Y/Overlay variables only), I could set a unique key at <tr>, however, the same approach (setting keys at two <tr>) does not work for faceted data case (with Facet variable). Just in case, tried to set all <tr> and <td> but it fails. 

Can you please look at this and advise me what I have missed ? Setting keys at <tr> are committed [here](https://github.com/VEuPathDB/web-eda/pull/749/commits/fd17dbf3bea26778096b7a5177de02ed7171e74a).

jtlong3rd:
> @jtlong3rd I am puzzled why the unique key does not work. So, for the case of no faceted data (X/Y/Overlay variables only), I could set a unique key at , however, the same approach (setting keys at two ) does not work for faceted data case (with Facet variable). Just in case, tried to set all and but it fails.
> 
> Can you please look at this and advise me what I have missed ? Setting keys at are committed [here](https://github.com/VEuPathDB/web-eda/pull/749/commits/fd17dbf3bea26778096b7a5177de02ed7171e74a).

The keys you've committed look good! There's one other place that needs a key - I'll add suggestions to the code.

src/lib/core/components/ScatterplotRsquareTable.tsx
```diff
@@ -0,0 +1,140 @@
...
+              <th>{facetVariable?.displayName}</th>
+              {overlayVariable != null && (
+                <th>{overlayVariable.displayName}</th>
+              )}
+              <th className="numeric">
+                R<sup>2</sup> (Best fit)
+              </th>
+            </tr>
+            {typedFilteredData.data?.map((data) => (
+              <>
```
jtlong3rd, line 91: You'll want to give a key to each of these [React fragments](https://reactjs.org/docs/fragments.html). To do this, you'll need to switch from using the ["short" <> syntax](https://reactjs.org/docs/fragments.html) (which can't receive a `key`) to the `<React.Fragment>` syntax, which [*can* receive a `key`](https://reactjs.org/docs/fragments.html#keyed-fragments).

```suggestion
              <React.Fragment key={data.label}>
```

src/lib/core/components/ScatterplotRsquareTable.tsx
```diff
@@ -0,0 +1,140 @@
...
+                              {overlayVariable != null && (
+                                <td>{series?.name?.split(', Best fit')[0]}</td>
+                              )}
+                              <td>{series.r2 ?? 'N/A'}</td>
+                            </tr>
+                          );
+                        }
+                      })
+                  : ''}
+              </>
```
jtlong3rd, line 133:
```suggestion
              </React.Fragment>
```

src/lib/core/components/ScatterplotRsquareTable.tsx
jtlong3rd: In general, whenever you're `.map`ping to an array of React elements (even `<>`), those elements should be given unique keys.

moontrip:
@jtlong3rd Oh my... I missed that I used the React.Fragment in there! Perhaps it was too small to find, especially after long Thanksgiving break 😄 I thought the top level one inside map() was just <tr>s focusing on `return`

Yes it works now. Thank you for your suggestion! 👍

jtlong3rd (approved):
🎉

bobular (commented):
Functional review only!

With no overlay variable, the twice-appearance of "Best fit" looks a bit odd in the table:
image

Looks great with an overlay variable.

Then things looked good with a facet variable, but then I got this

image

Happened again after a page reload - so doesn't seem to be an intermittent back end thing.
Maybe faceting isn't ready for testing yet?  I can give more diagnostics if needed though.

bobular:
The error didn't happen with a difference facet (and y) variable... mysterious
image

moontrip:
@bobular thank you for your tests! 👍 As for the single table, the reason why I used Best fit in there is to match the name at legend label. Certainly it can be determined: this is the main reason why this PR is set to be draft PR 😃 

I could get the same error message like you, even for my test variables that have been working fine this afternoon - suddenly it started throw those error message somehow.

bobular (commented):
I just tested/reviewed again and I'm still getting those 500s :-( 
```
Unhandled Rejection (Error): 500 Internal Server Error
Error occured while trying to proxy: localhost:3000/users/578013513/analyses/PuxDtIq
```
I don't know what's going on here.

The PATCH request body is at least 1,048,576 characters - my browser tools say "Request has been truncated"

Maybe something in this request is reaching a limit?  But why only in this branch? Aha no it's also happening in the main branch.  Is this related to thumbnails and disabling GL?  I'll make a ticket if there isn't one already.  Problem seems to go away when subsetting down to one smaller country...

Here's the new ticket https://github.com/VEuPathDB/EdaNewIssues/issues/102 
Seems to behave differently depending on deployment locally or on beta

moontrip:
@bobular I have updated this PR as you suggested. Now the R-square value is shown in the legend for no overlay/facet variable, instead of a table. Here is a screenshot for an example. Will change this to be normal PR.

scatter-r2

jtlong3rd (approved):

bobular (approved):
Great - I will approve based on the screenshot because I'm having problems getting my dev site working.

Edit: and the code - it looks fine!

Issue #676, opened 2021-11-09
Visualizations: no data detection needs to be more sophisticated now we have overlay and facet vars


bobular:
This is what it should NOT look like:

(GEMS1 beta site, 8/Dec/2021)

A. Empty "No data" facet for a variable (Country) with no missing data.  (The overlay variable which does have missing data, is triggering this.) 
![image](https://user-images.githubusercontent.com/308639/145263576-043640bd-bcff-48da-be62-5760e26271e3.png)

B. Empty "No data" item in overlay variable legend (facet variable is triggering this)
![image](https://user-images.githubusercontent.com/308639/145264019-1ec0cc79-9972-46f5-a995-4c24e97f70c6.png)
 
Please also test a "Main variable" that has missing data. E.g., "Given stored water for drinking"

bobular:
Hi @danicahelb - this has both b55 and b56 labels!  DSR said it was not a showstopper that unintended "No data" overlays and facets might appear in certain edge-cases.  I will work on it next, but am concerned that a substantial change in the logical behaviour of visualisations might introduce further bugs this close to release.  Your thoughts?

bobular:
Just an update to say I am still working on this!  It's complicated but I am hopefully arriving at a parsimonious solution soon.

aaronwlsong:
Looks good now, closing ticket.
_____________________________________________________________________________________________
The case 1 :  "No data" facet
![image](https://user-images.githubusercontent.com/61077845/146431992-3a501891-2979-4fd4-85ef-efe500518db4.png)
The case 2. "No data" item in overlay variable
![image](https://user-images.githubusercontent.com/61077845/146432336-d1bfe12e-0643-424c-b5e7-4e3129290d3e.png)
The case 3. "No data" in Main variable
![image](https://user-images.githubusercontent.com/61077845/146440682-e88e063c-26d0-47a7-a8e4-5c908140f8e1.png)

PR #753, opened 2021-12-07
keep the state of checked legend items

moontrip:
This will address #172 . The purpose of this is to keep the state of last checked legend items after coming out and in again.
Tried several approaches and logics but settled down to the current approach using useState() and two useEffect() hooks, in conjunction with if-else conditions and dependencies at useEffect hook.

There exist four user behaviors to be considered:
1) making new plot
  a) need to update checked legend items with a whole new legend list
2) after 1), changing variable (e.g., overlay variable) within the new plot
  should be the same to 1a)
3) out of the plot and then re-entering (from thumbnail, Subset tab, etc.)
  b) need to use previous/last checked legend items
4) after 3), change variable (e.g., overlay variable) within the plot
  should be the same to 1a)

Note that currently, this only applied to barplot Viz. Will try to apply the same logic to other Vizs (except Mosaic as the custom legend is not used in there)


moontrip:
well I don't know why CI/build failed as this PR does not have any dependency. Hmm...

moontrip:
@dmfalke do you happen to know why this PR keeps failing when CI/build? This PR does not have any dependency (like web-components) and no specific error was found in my vs code. Perhaps a GitHub glitch? I will re-run the build tomorrow and see if it was a temporary GitHub issue

moontrip:
@dmfalke Just FYI, one of Build error is as follows. I could find a relevant [thread](https://github.com/nodejs/node/issues/38367) regarding this but I wonder why it happens now at our CI/build: it seems not to be a recent issue based on the thread (reported in April).

```
/home/runner/.node-gyp/16.13.0/include/node/v8-internal.h:492:38: error: ‘remove_cv_t’ is not a member of ‘std’; did you mean ‘remove_cv’?
  492 |             !std::is_same<Data, std::remove_cv_t<T>>::value>::Perform(data);
      |                                      ^~~~~~~~~~~
      |                                      remove_cv
```

dmfalke:
I just started seeing this, too. I will need to investigate. Seems like something changed with the CI build environment.

jtlong3rd (commented):
Hi @moontrip!

I wasn't able to test the PR today (due to the Penn maintenance), but here's some initial suggestions:

1. The two effects you're using can likely be moved into some sort of reusable [custom hook](https://reactjs.org/docs/hooks-custom.html); that way, you won't have to copy-and-paste into the other visualizations
2. We'll want to think about how to avoid introducing missing `useEffect` dependencies

Happy to go into more detail tomorrow. 🙂

moontrip:
Hi @jtlong3rd and @bobular The service seems to be back and had a chance to test. Below includes Bob's questions in Slack (private talk) and my answers, which is also related to Jamie's question.

BM: `onCheckedLegendItemsChange(vizConfig.checkedLegendItems)` at the first useEffect() is no-op?
DK: yes you are correct. Basically it is not necessary to run it again, thus I have removed that and simplified overall code.

BM: in the second useEffect(), I'm wondering why `newLegendList` isn't a dependency for the second useEffect()
DK: `newLegendList` should not be in the dependencies of the 2nd useEffect() in this case. When rendering this `BarplotViz` component, `newLegendList` has false value initially: it is the same for both cases, making new Viz or re-entering a pre-existing Viz. 
If the dependency is set, then it causes an issue when re-entering a pre-existing Viz. Re-entering Viz means that pre-existing `vizConfig.checkedLegendItems` will be utilized, but if the dependency is set, the second useEffect() is executed for update  because `newLegendList` is changed from `false` to `true` inside this 2nd useEffect() (`else setNewLegendList(true)`), which overrides `vizConfig.checkedLegendItems` with a full legendList, not preserving vizConfig.checkedLegendItems.

moontrip:
@jtlong3rd Thank you for your suggestion. I will try to make them be a custom hook, but it still needs a copy-paste anyway 😄 ? I mean, calling the custom hook still needs to be at each Viz.

jtlong3rd:
> @jtlong3rd Thank you for your suggestion. I will try to make them be a custom hook, but it still needs a copy-paste anyway 😄 ? I mean, calling the custom hook still needs to be at each Viz.

Yup, that's right - you'll still call the custom code at each Viz that need it, but there will be less duplicate code. (1 line for importing the custom hook, and ~ 1 line for calling the custom hook.)

moontrip:
@jtlong3rd as you suggested, a custom hook is made for that part. For now, this approach is applied to barplot and histogram Vizs: the leftovers such as boxplot and scatter plot Vizs would be next after review.

bobular (commented):
Well it works! In basic situations at least.  I guess the "missing" dependencies in a useEffect are not a serious as with a useMemo or useCallback (because the values of those variables are not frozen in the closure)

If you toggle the count/proportion control, then the checkbox state is reset, which might be kind of annoying?  Same happens when you add or remove a facet variable.

I'm not so sure about the "include xxx with no data" toggle.  That can **sometimes** change the number of items in the legend (but not always, after #676 is resolved).

My feeling is that **only** when `legendItems` changes, then `vizConfig.checkedLegendItems` should be updated with all of its values.  Unless of course we are coming from a thumbnail and `vizConfig.checkedLegendItems` is already non null.  But how to do this, I'm not sure!

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -371,10 +372,29 @@ function BarplotViz(props: VisualizationProps) {
...
 
-  // use this to set all checked
+  // set this useState to determine whether a new checkedLegendItems should be made
+  const [newLegendList, setNewLegendList] = useState<boolean>(false);
+
+  // run once when rendering and check if vizConfig.checkedLegendItmes is defined
+  // this will handle the case of pre-existing plot
+  useEffect(() => {
+    if (vizConfig.checkedLegendItems != null) {
+      onCheckedLegendItemsChange(vizConfig.checkedLegendItems);
```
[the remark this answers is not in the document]
moontrip: @bobular well you might have not seen the new commits 😃 It is now removed as you suggested before, which is handled at the custom hook

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -371,12 +376,13 @@ function BarplotViz(props: VisualizationProps) {
...
   }, [data]);
 
-  // use this to set all checked
-  useEffect(() => {
-    if (data != null) {
-      onCheckedLegendItemsChange(legendItems.map((item) => item.label));
-    }
-  }, [data, legendItems]);
+  // calling a custom hook to preserve the status of checked legend items
+  const checkedLegendItemsStatus = useCheckedLegendItemsStatus(
```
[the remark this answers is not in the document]
moontrip: I just assigned a variable but perhaps can simply call the hook only.

bobular:
> I guess the "missing" dependencies in a useEffect are not a serious as with a useMemo or useCallback (because the values of those variables are not frozen in the closure)

Or is that true?

This documentation https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects suggests that stale values could be used (like I fully understand with useMemo and useCallback) but I'm not sure here.  Surely if the dependencies change then a fresh first argument to useEffect will be called?

jtlong3rd:
> > I guess the "missing" dependencies in a useEffect are not a serious as with a useMemo or useCallback (because the values of those variables are not frozen in the closure)
> 
> Or is that true?
> 
> This documentation https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects suggests that stale values could be used (like I fully understand with useMemo and useCallback) but I'm not sure here. Surely if the dependencies change then a fresh first argument to useEffect will be called?

Yes, unfortunately, "missing" dependencies in a `useEffect` are just as problematic as for `useMemo` and `useCallback`.

This example does a great job illustrating: https://overreacted.io/a-complete-guide-to-useeffect/#each-render-has-its-own-everything

moontrip:
> Well it works! In basic situations at least. I guess the "missing" dependencies in a useEffect are not a serious as with a useMemo or useCallback (because the values of those variables are not frozen in the closure)
> 
> If you toggle the count/proportion control, then the checkbox state is reset, which might be kind of annoying? Same happens when you add or remove a facet variable.
> 
> I'm not so sure about the "include xxx with no data" toggle. That can **sometimes** change the number of items in the legend (but not always, after #676 is resolved).
> 
> My feeling is that **only** when `legendItems` changes, then `vizConfig.checkedLegendItems` should be updated with all of its values. Unless of course we are coming from a thumbnail and `vizConfig.checkedLegendItems` is already non null. But how to do this, I'm not sure!

@bobular thank you for your reviews and tests. Yes basically it only covers up the case, out-of-the-plot and re-entering, more specifically targeting the case from the thumbnail. It is the main goal of the ticket, citing Danielle's "when ive minimized a visualization (i was looking at scatter) and gone back to where i can add another, its reasonable a user might be looking to have the configurations they set when expanded/full screen to still apply."

So, any in-plot operations like changing plot-related options would recreate the legends with all checked: I didn't think about them.

bobular:
> It is the main goal of the ticket, citing Danielle's "when ive minimized a visualization (i was looking at scatter) and gone back to where i can add another, its reasonable a user might be looking to have the configurations they set when expanded/full screen to still apply."

That's true, but depending on the approach taken, we might have got the extra functionality (that I bring to the table late in the day) for free.  I'm a little uneasy with not being able to understand how your hook works, and the missing dependencies - future maintainability will be an issue.  It's too late for me to devote any useful thought to it now though...  I'll have another look tomorrow!

bobular:
> This example does a great job illustrating: https://overreacted.io/a-complete-guide-to-useeffect/#each-render-has-its-own-everything

Interesting stuff.  Though I was stuck for a while not noticing the difference between `setTimeout` and `setInterval` lol...

Interesting how they introduce `useReducer` before `useCallback`

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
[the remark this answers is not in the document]
jtlong3rd: Yup, it's OK to simply call the hook.

src/lib/core/hooks/checkedLegendItemsStatus.ts
```diff
@@ -0,0 +1,42 @@
...
+import { HistogramDataWithCoverageStatistics } from '../components/visualizations/implementations/HistogramVisualization';
+
+/**
+ * A custom hook to preserve the status of the checked legend items
+ */
+
+export function useCheckedLegendItemsStatus(
+  data: PromiseHookState<
+    BarplotDataWithStatistics | HistogramDataWithCoverageStatistics | undefined
+  >,
```
jtlong3rd, line 14: To avoid hardcoding different Viz types, I would use a type parameter:

```suggestion
export function useCheckedLegendItemsStatus<T>(
  data: PromiseHookState<T>,
```

dmfalke:
I have yet to look at this, but I will do so tonight.

src/lib/core/hooks/checkedLegendItemsStatus.ts
moontrip: @jtlong3rd thank you for your comments. As Bob also suggested, I have changed to simply call the custom hook without assigning it as a const. 

As for the generic, yes I also thought of the case as, without the generic, I should have added all other data types for different plot Vizs 😄 Those are committed - will remove 'export data type' at barplot and histogram Vizs sooner or later.

moontrip:
@bobular Here are some detailed explanation how the current implementation, using useState and two useEffect hooks, work. Previously, it was like
```
  useEffect(() => {
    if (data.value != null) {
        onCheckedLegendItemsChange(legendItems.map((item) => item.label));
    }
  }, [data, legendItems]);
```
This will always update checkedLegendItems whenever data (and legendItems accordingly) is changed. 

Now we need to consider the case "re-entering from thumbnail or subsetting tab for pre-existing plot". For this purpose, I have introduced one useState and one additional useEffect which the latter runs only once (dependency array is empty, []): current implementation. The initial newLegendList (at useState) is set to false, so whenever the page is rendered (create new plot, re-entering pre-existing plot, refreshing browser, etc.), it always has initial value of false. And the new useEffect (1st useEffect) checks whether checkedLegendItems exists via vizConfig.checkedLegendItems. If exists, pre-existing checkedLegendItems  is used (no-op) and newLegendList remains false. If non-existent, i.e., new plot Viz, then set newLegendList to be true. 

Now moving to the 2nd useEffect, the newLegendList value is used to check whether new checkedLegendItems is needed or not. One issue in this approach was that there was a case when changing variable, plot options, etc. within the plot: note that after re-entering from thumbnail, newLegendList = false, thus although data may be changed by changing variable, checkedLegendItems is not updated due to the if-condition of newLegendList = true. To avoid this case, else-condition is added so that newLegendList becomes true, which leads to updating checkedLegendItems whenever data is changed (`else setNewLegendList(true)`). 

If we add newLegendList as a dependency in the 2nd useEffect, what will happen is that if re-entering from thumbnail, pre-existing vizConfig.checkedLegendItems is utilized via 1st useEffect, and then onCheckedLegendItemsChange() in the 2nd useEffect is not called as newLegendList remains false. However, since newLegendList is false, the else-condition is executed (`else setNewLegendList(true)`). Consequently, since newLegendList is changed, onCheckedLegendItemsChange() in the 2nd useEffect is executed (useEffect's update), which overrides vizConfig.checkedLegendItems with fully checked legend items, so we will lose the status of the previous checkedLegendItems.

I wish this would be helpful for your better understanding.

dmfalke:
I'm not sure this is the best approach to the retain the checked legend items. I would instead store the list of checked legend items as a part of the visualization's UI state. The list should be optional, and default to all items. I don't think there is any need for component state, with this approach. Let me know what you think.

moontrip:
@dmfalke Thank you for sharing your thought. Definitely I don't think this is the best solution for that, but not quite sure what you have in mind. Any better approach would be welcome. 👍

dmfalke:
Sorry, I see that the list is being stored in the viz config. Before this PR, I see that the list was being updated to all items, regardless of what was already stored in the config:
https://github.com/VEuPathDB/web-eda/blob/ee16163651665b322d2053e2e0343c340d1d3565/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx#L361-L366.

I think a simple solution would be to remove the bit of code above, and if `vizConfig.checkedLegendItems` is `undefined`, pass all legend items to components that need it:
```typescript
            <Barplot
              data={data.value}
              ref={plotRef}
              // for custom legend: pass checkedLegendItems to PlotlyPlot
              checkedLegendItems={vizConfig.checkedLegendItems || allLegendItems}
              {...plotProps}
            />
```

Let me know what you think.

bobular:
Aha yes that looks like a nice simple solution!

Any reason why not using `??` ?

moontrip:
@dmfalke Thank you for your suggestion 👍 First of all, it is a minor thing that I have made that useEffect to be a custom hook, so it seems like you are looking at old code: it does not matter much though 😄 

So, getting back to your suggestion, unfortunately, it would not work for satisfying all four user's behavior I have noted at [my top comment](https://github.com/VEuPathDB/web-eda/pull/753#issue-1072796901): so I think that it is not that simple. For example, your case covers up my 1) case for sure. I think you had in mind like this:

`checkedLegendItems={vizConfig.checkedLegendItems != null ? vizConfig.checkedLegendItems : legendItems.map((item) => item.label)}`

So, what if vizConfig.checkedLegendItems is defined and changing variable ? Then the new legend would have no checked items because vizConfig.CheckedLegendItems is not updated with new legend list - this is the reason why I have used the approach like custom hook

bobular:
Dave's suggestion omitted the `map((item) => item.label)` but you're right, it doesn't address changing overlay variable.

A simple fix is to reset `vizConfig.checkedLegendItems` every time *any* variable is changed
```
  const handleInputVariableChange = useCallback(
    (selectedVariables: VariablesByInputName) => {
      const {
        xAxisVariable,
        overlayVariable,
        facetVariable,
      } = selectedVariables;
      updateVizConfig({
        xAxisVariable,
        overlayVariable,
        facetVariable,
	checkedLegendItems: undefined,
      });
    },
    [updateVizConfig]
  );
```
But this could be refined to check for only a change in `overlayVariable`

moontrip:
@bobular thank you for your suggestion. Unfortunately, it would not work as well because it means that vizConfig.checkedLegendItems always becomes undefined - plot will be shown but the legend would always have no checked items, like 

barplot1

moontrip:
@bobular seems like the above can be overcome. Working for the possibility.

dmfalke:
@moontrip how did you get into that state? Recall that part of the suggestion is that if `vizConfig.checkedLegendItems == null`, default to all legend items.

moontrip:
@dmfalke so, it would probably need to change PlotLegend component as well - I am checking.

dmfalke:
I don't think anything needs to change in `PlotLegend`.

In the viz, you can do something like this

```typescript
const checkedItems = useMemo(() => {
  return vizConfig.checkedLegendItems ?? legendItems.map(item => item.label)
}, [vizConfig.checkedLegendItems, legendItems[)
```

moontrip:
@dmfalke sorry for making a confusion. I meant that the value of a prop for PlotLegend should be changed. 

I have tested the new approach quite extensively and it seems to work fine: it would be great if you can test this as well. Thank you all for contributing great suggestions and ideas, which is an advantage of crowd sourcing! 👍

bobular (approved):
Works great! Approved assuming you'll roll out to the other vizs ;-) 

Shall we do "Only reset checked items when overlay variable is changed" as another ticket?

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -371,12 +373,12 @@ function BarplotViz(props: VisualizationProps) {
...
   }, [data]);
 
-  // use this to set all checked
-  useEffect(() => {
-    if (data != null) {
-      onCheckedLegendItemsChange(legendItems.map((item) => item.label));
-    }
-  }, [data, legendItems]);
+  // set checkedLegendItems
+  const checkedLegendItems = useMemo(() => {
```
bobular, line 377: We could implement this as a custom hook, right?

jtlong3rd: Yup, definitely.

moontrip:
@bobular Regarding  "Only reset checked items when overlay variable is changed", not quite sure if it would be better than the current one, using handleInputVariableChange callback you suggested. I might miss something but off the top of my head, the way to run & check the change of overlay variable is to use useEffect hook and call updateVizConfig() function inside it. 

This useEffect hook will need to have dependency of [updateVizConfig, vizConfig.overlayVariable], which causes a couple of issues: a) infinite loop by updateVizConfig dependency - need to introduce useRef bypass technique to avoid it, so another hook should be used; b) need appropriate condition or logic not to override pre-defined vizConfig.checkedLegendItems when coming from thumbnail. So I feel that this may make more hassles ?

bobular:
> So I feel that this may make more hassles ?

Sounds like a new ticket!

moontrip:
@bobular @jtlong3rd @dmfalke I have make the part be a custom hook and also applied this approach to other Vizs such as boxplot and scatter plot Viz.

Unfortunately, I think that there is an issue at Scatter plot with this approach - it does not work correctly when changing plot options (like Raw, Smoothed mean, Best fit). With the new approach, it does not account the change of plot options, which can potentially be a serious issue. For example, even without overlay variable, assuming that I choose Best fit, which shows two legend items such as Data and Best fit in general, and deselect Data. If I change plot option to be Raw, then empty plot is shown without legend (we hide legend for single data). Another case is inconsistent display between smoothed mean and best fit when deselecting some of legend items and switching option. 

I might miss something but I could not find another solution for the scatter plot case. Just in case I have also checked my previous approach on the scatter plot and at least it works fine for that. So, it is another crowd sourcing time if you don't like my solution for scatter plot 😄 Any suggestions are welcome!

dmfalke:
Can you add logic to those change handlers to reset checkedLegendItems?

dmfalke:
For instance, you can add a boolean argument to onChangeHandlerFactory, named something like “resetCheckedLegendItems”. If this argument is true, then set checkedLegendItems to undefined, otherwise leave it as is. 

```
const onChangeHandlerFactory = useCallback(
    < ValueType,>(key: keyof ScatterplotConfig, resetCheckedLegendItems: boolean) => (newValue?: ValueType) => {
      const newPartialConfig = resetCheckedLegendItems ? {
        [key]: newValue,
        checkedLegendItems: undefined
      } : {
        [key]: newValue
      };
             updateVizConfig(newPartialConfig);
    },
    [updateVizConfig]
  );
```

moontrip:
> For instance, you can add a boolean argument to onChangeHandlerFactory, named something like “resetCheckedLegendItems”. If this argument is true, then set checkedLegendItems to undefined, otherwise leave it as is.
> 
> ```
> const onChangeHandlerFactory = useCallback(
>     < ValueType,>(key: keyof ScatterplotConfig, resetCheckedLegendItems: boolean) => (newValue?: ValueType) => {
>       const newPartialConfig = resetCheckedLegendItems ? {
>         [key]: newValue,
>         checkedLegendItems: undefined
>       } : {
>         [key]: newValue
>       };
>        updateVizConfig(newPartialConfig);
>     },
>     [updateVizConfig]
>   );
> ```

@dmfalke Thank you for your information! 👍 When I was thinking of an alternative way, I also had in mind to combine checkedLegendItems: undefined in the onChangeHandlerFactory as the problem I noticed with the new approach could be solved if I can add checkedLegendItems: undefined at onValueSpecChange function. Your suggestion is what I wanted to implement 👍 I will try it!

moontrip:
@dmfalke I have changed onChangeHandlerFactory function for all Vizs (for consistency even if Mosaic does not use the custom legend): a slight change on your code above was made which you may not easily notice 😄. This allows more flexibility for plot controls like plot options for Scatter, showMissingness toggle switch.

I also noticed that the showMissingness causes similar issue with what I have observed with plot options, so I also set true option for the onShowMissingnessChange function.

It seems to me that now this functionality works as expected.

dmfalke:
@moontrip this looks great. I don't see the difference, though! :smile:

I do think this could be dried up a bit, but we can wait on that until we generally address refactoring visualizations.

moontrip:
> @moontrip this looks great. I don't see the difference, though! 😄
> 
> I do think this could be dried up a bit, but we can wait on that until we generally address refactoring visualizations.

Thank you for your review. Yeah the diff is too small to find 😄 ->  resetCheckedLegendItems?: boolean (added ooptional symbol to avoid setting all pre-existing on-functions to have the argument)

dmfalke:
Aha, great!

moontrip:
@bobular @jtlong3rd I guess you tentatively agree with my changes (with Dave's suggestion) as you already did approve this PR before ? 😃

jtlong3rd:
> @bobular @jtlong3rd I guess you tentatively agree with my changes (with Dave's suggestion) as you already did approve this PR before ? 😃

I'll re-approve for clarity's sake!

jtlong3rd (approved):

bobular:
Hi @moontrip - Sorry I haven't re-approved. Go ahead and merge though! No red flags in the comments here from me. Good catch with the show-missing and scatterplot modes requiring a reset of the checked legend items. The only thing I ask is to add a ticket in this repo if changing **any** variable causes the legend items to be reset (ideally only changing overlay variable will trigger the reset, I think!)

moontrip:
@bobular thank you for your comments! 👍 Will merge this.

I didn't carefully check the case of your last sentence suggesting only relying on overlay change, which I have no idea how to do that off the top of my head. But we will see.

bobular:
> I didn't carefully check the case of your last sentence suggesting only relying on overlay change, which I have no idea how to do that off the top of my head. But we will see.

Actually, maybe it's OK to reset on **any** variable change.  Let's see if outreach/QA think it's an issue.

PR #754, opened 2021-12-07
Enable user login

### LINKED ISSUE #755: Dev: Allow user login on local dev sites
Local development sites do not currently support user login. This would be useful for developing various features.

dmfalke:
This PR adds the ability to login to a WDK REST service. Note that I removed support for providing a custom `wdk_check_auth` cookie value. I also removed the way we were handling `JSESSIONID`, to simplify the logic in `setupProxy`. The EDA service container is currently using a path cookie, so it is not conflicting with the WDK service cookie.

closes #755


jtlong3rd (approved):
Thanks for doing this!

PR #756, opened 2021-12-07
set node version

dmfalke:
CI builds are failing since https://github.com/actions/runner was updated. Hoping setting an explicit node version will address this.


jtlong3rd (approved):
🤞

moontrip (approved):
@dmfalke thank you for your work on this. Yeah I noticed that the CI/Build uses node 16 so I wondered it as well. I think it would work fine 👍 🤞

moontrip:
I could confirm with my PRs that this fix solved the github CI/Build error.

PR #770, opened 2021-12-07
More sophisticated showMissing behaviour and reused more entity counts

### LINKED ISSUE #676: Visualizations: no data detection needs to be more sophisticated now we have overlay and facet vars

### LINKED ISSUE #752: Visualizations: `VariableCoverageTable` can take a `filteredCounts` prop instead of `filters`
At the moment (correct me if I'm wrong) the `VariableCoverageTable` component takes a `filters` prop and uses `useEntityCounts(filters)` to fetch the filtered counts.  We should be able to pass `filteredCounts` in directly from the calling viz components.  Should save some back end calls too.

bobular:
Resolves #676 
Resolves #752 

No data issue are fixed for all visualizations (mosaic doesn't need it because only one stratification var), as well as the rationalisation of the back end entity count calls.

There is one remaining UX niggle, which will need some discussion and a future ticket: when viewing a faceted plot, cancelling a subsetting filter via the global filters popup changes the layout of the "skeleton spinners" to single un-faceted layout. When the data comes from the back end, the layout reverts again to faceted.  This is to do with waiting for `facetedCounts` to no longer be pending.


bobular:
Ready for review please. Seems to work fine with latest merge from main.

dmfalke (approved):
I looked into the skeleton issue a little more and couldn't find a solution in the short time I spent on it. Let's go with this for now and revisit. I suspect the solution will involve ensuring `filters` and `EntityCounts` props that are passed to visualizations are consistent. It's late and I don't see an obvious way to do this, but I'm confident there is a solution.

moontrip (approved):
I have tested a couple of cases and they looked fine 👍

bobular:
> I looked into the skeleton issue a little more and couldn't find a solution in the short time I spent on it. Let's go with this for now and revisit. I suspect the solution will involve ensuring `filters` and `EntityCounts` props that are passed to visualizations are consistent. It's late and I don't see an obvious way to do this, but I'm confident there is a solution.

Thanks both for reviewing.

I've added a ticket for the skeleton issue. Please change if you don't think it's `low cost`

PR #771, opened 2021-12-08
Use new FacetedPlot variants

chowington:
Fixes https://github.com/VEuPathDB/web-components/issues/259

A couple of outstanding tasks:

- [x] Center the modal plots better
- [x] Fix the broken modal close button icon

### DIFF
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -291,3 +291,9 @@ $data-table-inner-border: 1px solid #888;
     border-bottom: $data-table-inner-border;
   }
 }
+
+.ReactModal__Content {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -291,9 +291,3 @@ $data-table-inner-border: 1px solid #888;
     border-bottom: $data-table-inner-border;
   }
 }
-
-.ReactModal__Content {
-  display: flex;
-  align-items: center;
-  justify-content: center;
-}
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.11.4",
+    "@veupathdb/components": "^0.11.5",
     "@veupathdb/core-components": "^0.2.45",
     "@veupathdb/http-utils": "^1.0.1",
     "@veupathdb/study-data-access": "^0.1.2",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.11.2",
+    "@veupathdb/components": "^0.11.3",
     "@veupathdb/core-components": "^0.2.45",
     "@veupathdb/http-utils": "^1.0.1",
     "@veupathdb/study-data-access": "^0.1.2",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@types/debounce-promise": "^3.1.3",
-    "@veupathdb/components": "^0.11.3",
+    "@veupathdb/components": "^0.11.4",
     "@veupathdb/core-components": "^0.2.45",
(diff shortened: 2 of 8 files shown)


chowington:
@bobular @dmfalke Do y'all see the close button icon or is it broken for you as well? Here's what it looks like for me:

image

chowington:
I also noticed that the mosaic y-axis labels are overflowing their container again sometimes, after multiple attempts on my part to fix that. It'll have to be a separate ticket since I've already merged the web-components portion of this work.

moontrip:
@chowington Yes the close (X) marker is not shown for me neither, however, this works for me (use fa instead of fas)
`<i class="fa fa-times fa-3x"></i>`

Intentionally used fa-3x for checking it: you can use fa-lg, fa-2x, etc for sure :)

chowington:
That's simple enough! Thanks, DK!

moontrip:
@chowington as for vertical centering, though there may be a better approach, one way is to add flex and align-items for the following div class, 

`<div class="ReactModal__Content ReactModal__Content--after-open" ...>`

so, like adding the following two css properties in there.
```
display: flex
align-items: center
```

This will make vertical center as screenshot below (worked at both chrome and firefox), but be caution that this may affect all react-modal without specifically pointing out the target as you know well :) 

snap1

I didn't check it carefully but this means that perhaps it can be configured/controlled from Mike's FullScreenModal component ? I mean, since it utilizes react-modal, there may be a way to set vertical center: looks like customStyles at [the module's example](https://github.com/reactjs/react-modal#examples) may also align horizontally and vertically ?

moontrip:
@chowington Sorry I forgot to say one additional change at FacetedPlot component in conjunction with above. Wrapping modalPlot with div as well. So like,

```
      {modalComponentProps && (
        <FullScreenModal visible={modalIsOpen}>
          <button
            onClick={() => setModalIsOpen(false)}
            style={{
              position: 'absolute',
              top: 30,
              right: 30,
              backgroundColor: 'white',
              cursor: 'pointer',
              border: 'none',
              zIndex: 2000,
            }}
            title="Close expanded plot"
          >
            <i className="fas fa-times fa-lg"></i>
          </button>
          <div> {/* DK: here */}
            {modalPlot}
          </div> {/* DK: here */}
        </FullScreenModal>
      )}
```

chowington:
Thanks DK! That worked really well

chowington:
PR is ready for review! @dmfalke @bobular @moontrip

chowington:
In this commit (https://github.com/VEuPathDB/web-eda/pull/771/commits/6af36b32458accacdcefad75f6c94746e1944543) I fixed some unrelated TS warnings in the same files, and in the process I noticed that adding some dependencies to dependency arrays as TS recommends actually breaks the viz. A couple of comments already in the code hinted at that. I only committed the changes that didn't seem to break the code, but if anyone else is more familiar please do review the changes I made.

moontrip (approved):
@chowington I have just tested two cases, vertical center and X icon and they worked fine: I like the full screen modal to see a big plot 😄 

Certainly not for this ticket but it comes to my mind that such a full screen modal would be quite useful for non-faceted plots as well, especially the cases that lots of stratifications make one difficult to see details in a normal small-size plot. In my opinion, it deserves to be a new task :)

bobular (commented):
Apart from the interactive thing discussed on slack, this looks great.

Do you intend the mosaic legend/axis labels to be all bunched up? I think they look better spaced out.
image

Also, I'm curious what DK's div wrapping suggestion did? @moontrip

chowington:
> Apart from the interactive thing discussed on slack, this looks great.
> 
> Do you intend the mosaic legend/axis labels to be all bunched up? I think they look better spaced out. image
> 
> Also, I'm curious what DK's div wrapping suggestion did? @moontrip

@bobular They're bunched up because `containerStyles.height = '100%'`. I couldn't find a way to determine the plot height in px from non-px values, which is required to know how tall the not-legend can be, so the safe fallback is to not space them out at all

moontrip:
Hi @bobular 👋 I saw you called me 😄 

So, when using display:flex, it consists of flex container (so-called parent) and child div(s). It is often obscure which ones should be flex container (parent) and children, especially when using library-based component in React. So, simply put, the <div> that wraps plotly plot is a sort of declaration/clarification that it is a child of a flex container.

chowington:
@bobular How about my lastest commit? Does that work for you?

src/lib/core/components/visualizations/Visualizations.scss
```diff
@@ -291,3 +291,9 @@ $data-table-inner-border: 1px solid #888;
     border-bottom: $data-table-inner-border;
   }
 }
+
+.ReactModal__Content {
```
eikonomega, line 295: @chowington This might not be the best idea. If this is a global definition, it would affect all modals.

chowington: It no longer exists :)

chowington:
The faceted plot spacing isn't perfect, but this needs to go ahead and get merged. We can add another ticket for the plot spacing if we need to. Merging now.

PR #773, opened 2021-12-13
Enable Public Analyses

eikonomega:
## What's New
- Enables users to make an analysis public.

## What's Updated
- Various name changes with new version of CoreUI.


dmfalke (commented):
This is looking great! Code looks good. I did notice a few things that would be good to address before merging.

1. Tooltips on buttons block content when hidden. It looks like they are hidden with `opacity: 0`. Maybe `display: none` would work better?
image

2. The close icon in the modal is not a button. For a11y, I think it might be good to wrap this in a `<button>`, or add appropriate aria tags.
image

Let me know what you think. Thanks!

dmfalke (approved):

dmfalke:
Looks good to go!

Issue #653, opened 2021-10-29
Inconsistency in appearance of 400 errors for renaming an analysis with too long a name


SheenaTomko:
Just finished qa of https://github.com/VEuPathDB/EdaUserService/issues/7 and noticed the following:
1) If you try and rename an open analysis with a name that's too long, you get a pop up with the 400 error message, but if you dismiss the box, the long name persists until you go the "My analyses" page, homepage, etc.
<img width="709" alt="Screen Shot 2021-10-29 at 2 27 09 PM" src="https://user-images.githubusercontent.com/54597319/139492129-0add2557-e4a0-4f96-9ba4-51f1b0d7b856.png">

<img width="884" alt="Screen Shot 2021-10-29 at 2 27 35 PM" src="https://user-images.githubusercontent.com/54597319/139492150-b8425155-abd6-4cdb-bdbc-6e1f16a7012b.png">

2) If you try and rename an analysis from the "My analyses" page with a name that's too long, text appears at the top of the page. This is more subtle than the pop-up. However, the name itself reverts to what it was before. The error message on the page persists even after you've finished correctly renaming the strategy.

<img width="1111" alt="Screen Shot 2021-10-29 at 2 28 14 PM" src="https://user-images.githubusercontent.com/54597319/139492250-bd2502fc-5e23-4739-b938-9b66caaff837.png">

I think this is a low-priority ticket

chowington:
The real problem here is that the frontend sends the bad request in the first place, so the solution is to prevent the user from attempting to rename the analysis a name that's too long for the backend. That has been accomplished with the linked PR.

SheenaTomko:
This looks great. Now, whether I'm trying to name it from the open analysis or the analysis page, I can't type >50 characters. Much more obvious UX

PR #777, opened 2021-12-14
Max analysis name length

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/653


src/lib/workspace/AllAnalyses.tsx
```diff
@@ -78,6 +78,7 @@ const useStyles = makeStyles({
 });
 
 const UNKNOWN_DATASET_NAME = 'Unknown study';
+export const ANALYSIS_NAME_MAX_LENGTH = 50;
```
jtlong3rd, line 81: This `ANALYSIS_NAME_MAX_LENGTH` constant should probably go into a utility file - maybe [here](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/core/utils/analysis.ts)?

jtlong3rd (approved):
I have one small inline suggestion; other than that, this should be good to merge!

jtlong3rd (approved):

Issue #774, opened 2021-12-14
Map visualization: implement very basic non-semantically zoomable entity count markers


bobular:
- [x] create `io-ts` rules/validators for `map-markers` pass-thru plugin
- [x] create a new `MapVisualization` (akin to `BarplotVisualization`) which calls the `map-markers` endpoint with some things semi-hardcoded (later they will become more user-configurable!)
      - outputEntity should be the root entity of the study
      - the lat and long variables should be OBI_0001620 and OBI_0001621 respectively (again using the Study's root entity)
      - geoAggregateVariable should be geohash_3 EUPATH_0043205 (or whichever level works for early testing) again with study root entity
      - filters should be passed to the back end as usual (
      - note: the viz will NOT have any variable pickers (`<InputVariables>`)
- [x] use the MapVEu component demonstrated in [this story](https://veupathdb.github.io/web-components/?path=/story/map-donut-markers--first-request) to render simple white circle markers with entity counts - and, for now, initialize the map with a whole world or whole Africa center/zoom level 
      - the size of the map might need to be bigger than other visualizations - let's see how it looks the same size for now, though
- [x] figure out z-plane for map and global filters modal/popup to co-exist nicely (popup is currently behind the map)

We should be able to test this with **UMSP Surveillance** and **SCORE Seasonal Transmission S. haematobium Cluster Randomized Trial** studies that have recently been loaded with geolocation variables (on the root entity).

nkittur-uga:
Tested in SCORE Seasonal, can see simple white circle markers with entity counts, and map is initialized with with a whole world center/zoom level. Passes QA.

PR #782, opened 2021-12-17
Map viz first ticket #774

bobular:
<s>Requires https://github.com/VEuPathDB/web-components/pull/282</s> MERGED to 0.11.12
<s>Requires https://github.com/VEuPathDB/WDKClient/pull/171</s> MERGED to 0.5.4

Next up:

- [x] remember center and zoom level
- [x] implement io-ts types
- [x] wire in basic back end call and render basic DonutMarkers
- [x] resolve issue with CSS inclusion (I think this prevents overlapping markers rising to top on mouse-over)
- [x] make zoomlevel-to-geohash-level mapping a prop that can be passed to the map component (currently the mappings are defined once on the EDA side, and another time on the web-components side (for the map grid))


src/lib/core/hooks/visualizations.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param configData - raw JSON response from back end
+ * @param configDecoder - io-ts decoder object
+ * @param createDefaultConfig - function to create default config js object
+ * @param updateConfiguration - callback to set/update entire configuration
+ *
+ * @return vizConfig - the configuration
+ * @return updateVizConfig - partial update function
+ */
+export function useVizConfig<ConfigType>(
```
bobular, line 17: Note: my first custom hook?  This is a small step towards DRYing up the visualizations.

src/lib/core/components/GlobalFiltersDialog.scss
```diff
@@ -4,7 +4,7 @@
   left: 0;
   width: 100vw;
   height: 100vh;
-  z-index: 100;
+  z-index: 1100;
```
bobular, line 7: Some map components go up to z=1000

chowington:
@bobular What study can be used to test this on? I just looked at a couple, but no markers show up, which I assume meant the study didn't have geographical data.

bobular:
> @bobular What study can be used to test this on? I just looked at a couple, but no markers show up, which I assume meant the study didn't have geographical data.

Oh yes, sorry, UMSP!

src/lib/core/components/computations/PassThroughComputation.tsx
```diff
@@ -41,6 +42,7 @@ const visualizationTypes: Record<string, VisualizationType> = {
   // densityplot: scatterplotVisualization,
   barplot: barplotVisualization,
   boxplot: boxplotVisualization,
+  ['map-markers']: mapVisualization,
```
chowington, line 45:
```suggestion
  'map-markers': mapVisualization,
```

Clears up an eslint warning

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -0,0 +1,242 @@
...
+    visualization,
+    updateConfiguration,
+    updateThumbnail,
+    filters,
+    dataElementConstraints,
+    dataElementDependencyOrder,
+    starredVariables,
+    toggleStarredVariable,
+    totalCounts,
+    filteredCounts,
```
chowington, line 81: These cause eslint unused variable warnings, though I assume we will use them soon. `updateThumbail` does too, but I'm gonna use it in my PR, of course

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+  const [vizConfig, updateVizConfig] = useVizConfig(
+    visualization.descriptor.configuration,
+    MapConfig,
+    createDefaultConfig,
+    updateConfiguration
+  );
+
+  const handleViewportChanged = useCallback(
+    ({ center, zoom }) => {
```
chowington, line 100:
```suggestion
    ({ center, zoom }: { center?: number[] | null; zoom?: number | null }) => {
```

I suggest adding typing to prevent `center` and `zoom` being `any`, though the inline type isn't very pretty

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -0,0 +1,242 @@
...
+    }, [
+      studyId,
+      filters,
+      dataClient,
+      tempOutputEntityId,
+      latitudeVariableDetails,
+      longitudeVariableDetails,
+      boundsZoomLevel,
+      computation.descriptor.type,
+      defaultAnimationDuration,
```
chowington, line 213:
```suggestion
```

Clearing another eslint warning for unnecessary dependency

chowington (approved):
Just some eslint/TS suggestions, but otherwise, this looks great and works great!

src/lib/core/components/computations/PassThroughComputation.tsx
bobular: Ooh thanks.  I've used the square brackets before.  Not sure why not this time!

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
bobular: Yeah I think we should leave them in for now.  This is just one step along the road.

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
bobular: `defaultAnimationDuration` is passed as a prop to the `DonutMarker` component, so I think it's needed?

src/lib/core/components/computations/PassThroughComputation.tsx
chowington: eslint just said they're not necessary for this key value. Nothing major 👍

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
chowington: While it is used in the callback, I believe the idea is that `defaultAnimationDuration` is a module import and so it won't ever change in the component's lifecycle

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
bobular: That's weird - my emacs IDE doesn't highlight this as a problem/warning, though the type of the callback is
```
const handleViewportChanged: ({ center, zoom }: any) => void
```
So it should warn about `any`s.

I'll try and make a prettier fix, but don't want to import `react-leaflet`'s `Viewport` type into the client.  I think I have an idea.

bobular: I made a [fix here](https://github.com/VEuPathDB/web-eda/pull/782/commits/a71986cf0a37208818dc23790ef9ad4b781026f2) but I don't really see why we have to do it. I'm sure I haven't had to do that kind of thing before...

chowington: Ah, nice one! Yeah, IDK why eslint doesn't complain about those `any`s

Issue #761, opened 2021-11-30
Numbers in table cells should be right aligned


bobular:
Something that has been bugging me...

For example in the Mosaic data tables

Currently
![image](https://user-images.githubusercontent.com/308639/144078629-5fbdf638-c642-4130-84a7-bd0504cf9952.png)

After my quick hack in dev tools
![image](https://user-images.githubusercontent.com/308639/144078393-6c89fc45-7e49-4d48-a442-8a3b6f1b9b26.png)

Right aligned numbers make more sense - they are easier to compare.
It needs to be done more carefully than my hack.  The table headings are also right-justified which isn't ideal.

aaronwlsong:
looks good now. Closing ticket.

![image](https://user-images.githubusercontent.com/61077845/150411125-0af735d6-b997-4551-baa0-bf3e1a475b0b.png)

![image](https://user-images.githubusercontent.com/61077845/150411414-c203ba88-4713-453c-929b-36baf8e2af69.png)

PR #786, opened 2021-12-21
Right align table numbers

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/761


bobular (approved):
All looks great - including the p-value tables.  Thanks.

PR #792, opened 2022-01-11
New NPM dependency structure

jtlong3rd:
This PR:

1. Upgrades `@veupathdb/study-data-access` (which, due to https://github.com/VEuPathDB/web-study-data-access/pull/5, eliminates a transitive dependency on `@veupathdb/wdk-client`)
2. Makes `@veupathdb/wdk-client` a peer and dev dependency (to reduce the risk of introducing transitive dependencies of `@veupathdb/wdk-client` to consumers of `@veupathdb/web-eda`)
3. Applies the `ify-loader` to only JS/TS assets. (This resolves an issue where the `ify-loader` was mis-bundling images from `@veupathdb/web-common`.)

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -38,6 +38,7 @@ module.exports = function override(config, env) {
         // a dev dependency.
         {
           test: /\.(js|jsx|ts|tsx)$/,
+          include: /node_modules/,
           loader: 'ify-loader',
         },
       ],
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -37,7 +37,7 @@ module.exports = function override(config, env) {
         // This also needs the bubleify packages installed as
         // a dev dependency.
         {
-          test: /node_modules/,
+          test: /\.(js|jsx|ts|tsx)$/,
           loader: 'ify-loader',
         },
       ],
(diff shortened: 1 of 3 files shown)


config-overrides.js
```diff
@@ -37,7 +37,8 @@ module.exports = function override(config, env) {
         // This also needs the bubleify packages installed as
         // a dev dependency.
         {
-          test: /node_modules/,
+          test: /\.(js|jsx|ts|tsx)$/,
+          include: /node_modules/,
```
jtlong3rd, line 41: @dmfalke Just to clarify, this change to the `ify-loader` config is a twofer. It:

1. Addresses that issue where images from `@veupathdb/web-common` were being mis-bundled. (By not applying the `ify-loader` to images.)
2. Preemptively addresses other bugs which could result from the `ify-loader` being applied to non-JS assets. (The [plotly documentation](https://plotly.com/javascript/getting-started/#webpack-user) insinuates that the `ify-loader` only needs to be applied to JS assets.)

I was concerned that our original solution (“apply `ify-loader` to the entire contents of `node_modules`, except for @veupathdb dependencies”) might exclude files in `@veupathdb/components` that need to handled with `ify-loader`.

(Contributing to this ambiguity over which files should be handled with `ify-loader`: currently, our web-eda local dev environment, as well as “WDK sites built from npm modules,” both seem to build fine without the `ify-loader` in place. It would be great, if the long term, we can remove the `ify-loader`, as that's adding 30-40 seconds to the `wb site` build time.)

dmfalke (approved):

PR #794, opened 2022-01-12
axis range controls for histogram, bar, box, and scatter plot vizs

moontrip:
Update: To use different size/width of axis range controls between histogram filter and Vizs, some changes are made here and web-components. Thus, now this PR relies on [the PR at web-components](https://github.com/VEuPathDB/web-components/pull/292) (`yarn link` is required)

This is the first step for adding zoom & truncation axis range controls to Vizs: tried to Histogram Viz only. I did not clean up the codes as it needs some updates but it would work out in most cases as of now.

Introduced/converted to custom hooks and functions that are intended for the use at other Vizs: but not that generic now, which will be changed.

Mostly works fine but there are known issues that need to be addressed: a) adjusting input form size as plot area is rather small; b) an instant discrepancy during multiple re-renderings between local value and vizConfig value. I need a help for b) from you, @dmfalke, @jtlong3rd, @bobular and others who are interested.

The issue I have is that computed `defaultIndependentAxisRange` and `defaultDependentAxisRange` have unsync with corresponding `vizConfig` (i.e., `vizConfig.independentAxisRange` and `vizConfig.dependentAxisRange`) at a point among multiple re-renderings. Please note that they eventually become the same at the end of re-rendering, so from the operation viewpoint, it is not an issue. But it causes to pop up an erroneous truncation warning message due to the instant discrepancy. Actually I could resolve the case of the `independentAxisRange` but still has the issue for `dependentAxisRange`. One exemplary use case is when switching the plot option from the Proportion to the Count: it is because the condition to pop up the warning message is `defaultDependentAxisRange` > `vizConfig.dependentAxisRange`. 
 
It would be great if you can check and suggest any idea to overcome the above issue.

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -570,6 +570,8 @@ function HistogramPlotWithControls({
         selectedRange={selectedRange}
         selectedRangeBounds={selectedRangeBounds}
         onSelectedRangeChange={handleSelectedRangeChange}
+        //DKDK parentComponentName
+        parentComponentName={'HistogramFilter'}
       />
       <Histogram
         {...histogramProps}
@@ -629,6 +631,8 @@ function HistogramPlotWithControls({
             }}
             allowPartialRange={false}
             containerStyles={{ minWidth: '400px' }}
+            //DKDK parentComponentName
+            parentComponentName={'HistogramFilter'}
           />
           {/* truncation notification */}
           {truncatedDependentAxisWarning ? (
@@ -678,6 +682,8 @@ function HistogramPlotWithControls({
             onRangeChange={handleIndependentAxisRangeChange}
             valueType={data?.valueType}
             containerStyles={{ minWidth: '400px' }}
+            //DKDK parentComponentName
+            parentComponentName={'HistogramFilter'}
           />
           {/* truncation notification */}
           {truncatedIndependentAxisWarning ? (
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -5,7 +5,7 @@ import { preorder } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
 import { getOrElse } from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/function';
 import * as t from 'io-ts';
-import { useCallback, useMemo } from 'react';
+import { useCallback, useMemo, useState, useEffect } from 'react';
 
 // need to set for Scatterplot
 
@@ -80,12 +80,7 @@ import {
   ColorPaletteDefault,
   ColorPaletteDark,
 } from '@veupathdb/components/lib/types/plots/addOns';
-// import variable's metadata-based independent axis range utils
-import { defaultIndependentAxisRange } from '../../../utils/default-independent-axis-range';
-import { axisRangeMargin } from '../../../utils/axis-range-margin';
 import { VariablesByInputName } from '../../../utils/data-element-constraints';
-// util to find dependent axis range
-import { defaultDependentAxisRange } from '../../../utils/default-dependent-axis-range';
 import { useRouteMatch } from 'react-router';
 import { Link } from '@veupathdb/wdk-client/lib/Components';
 import PluginError from '../PluginError';
@@ -99,8 +94,24 @@ import FacetedXYPlot from '@veupathdb/components/lib/plots/facetedPlots/FacetedX
 import * as ColorMath from 'color-math';
 // R-square table component
 import { ScatterplotRsquareTable } from '../../ScatterplotRsquareTable';
-//DKDK a custom hook to preserve the status of checked legend items
+// a custom hook to preserve the status of checked legend items
 import { useCheckedLegendItemsStatus } from '../../../hooks/checkedLegendItemsStatus';
+
+// concerning axis range control
+import { NumberOrDateRange } from '../../../types/general';
+import { padISODateTime } from '../../../utils/date-conversion';
+// reusable util for computing truncationConfig
+import { truncationConfig } from '../../../utils/truncation-config-utils-viz';
+// use Notification for truncation warning message
+import Notification from '@veupathdb/components/lib/components/widgets//Notification';
+import Button from '@veupathdb/components/lib/components/widgets/Button';
+import AxisRangeControl from '@veupathdb/components/lib/components/plotControls/AxisRangeControl';
+import { UIState } from '../../filter/HistogramFilter';
+// change defaultIndependentAxisRange to hook
+import { useDefaultIndependentAxisRange } from '../../../hooks/computeDefaultIndependentAxisRange';
+// for scatter plot, use another custom hook different from other Vizs
+import { useDefaultDependentAxisRange } from '../../../hooks/computeNumberDateDefaultDependentAxisRange';
+import LabelledGroup from '@veupathdb/components/lib/components/widgets/LabelledGroup';
 
 const MAXALLOWEDDATAPOINTS = 100000;
 const SMOOTHEDMEANTEXT = 'Smoothed mean';
@@ -126,8 +137,8 @@ const modalPlotContainerStyles = {
   margin: 'auto',
 };
 
-// define XYPlotDataWithCoverage
-interface XYPlotDataWithCoverage extends CoverageStatistics {
+// define XYPlotDataWithCoverage and export
+export interface XYPlotDataWithCoverage extends CoverageStatistics {
   dataSetProcess: XYPlotData | FacetedData<XYPlotData>;
   // change these types to be compatible with new axis range
   yMin: number | string | undefined;
@@ -174,6 +185,9 @@ export const ScatterplotConfig = t.partial({
   showMissingness: t.boolean,
   // for vizconfig.checkedLegendItems
   checkedLegendItems: t.array(t.string),
+  // axis range control
+  independentAxisRange: NumberOrDateRange,
+  dependentAxisRange: NumberOrDateRange,
 });
 
 function ScatterplotViz(props: VisualizationProps) {
@@ -248,6 +262,16 @@ function ScatterplotViz(props: VisualizationProps) {
     vizConfig.facetVariable,
   ]);
 
+  // set the state of truncation warning message
+  const [
+    truncatedIndependentAxisWarning,
+    setTruncatedIndependentAxisWarning,
+  ] = useState<string>('');
+  const [
+    truncatedDependentAxisWarning,
+    setTruncatedDependentAxisWarning,
+  ] = useState<string>('');
+
   // TODO Handle facetVariable
   const handleInputVariableChange = useCallback(
     (selectedVariables: VariablesByInputName) => {
@@ -269,7 +293,13 @@ function ScatterplotViz(props: VisualizationProps) {
             : vizConfig.valueSpecConfig,
         // set undefined for variable change
         checkedLegendItems: undefined,
+        // set independentAxisRange undefined
+        independentAxisRange: undefined,
+        dependentAxisRange: undefined,
       });
+      // close truncation warnings here
+      setTruncatedIndependentAxisWarning('');
+      setTruncatedDependentAxisWarning('');
     },
     [updateVizConfig, findEntityAndVariable, vizConfig.valueSpecConfig]
   );
@@ -277,16 +307,27 @@ function ScatterplotViz(props: VisualizationProps) {
   // prettier-ignore
   // allow 2nd parameter of resetCheckedLegendItems for checking legend status
   const onChangeHandlerFactory = useCallback(
-    < ValueType,>(key: keyof ScatterplotConfig, resetCheckedLegendItems?: boolean) => (newValue?: ValueType) => {
+    < ValueType,>(key: keyof ScatterplotConfig, resetCheckedLegendItems?: boolean, resetTruncationWarnings?: boolean) => (newValue?: ValueType) => {
       const newPartialConfig = resetCheckedLegendItems
         ? {
             [key]: newValue,
-            checkedLegendItems: undefined
+            checkedLegendItems: undefined,
+            // set independentAxisRange undefined
+            independentAxisRange: resetTruncationWarnings ? undefined : defaultIndependentRange,
+            dependentAxisRange: resetTruncationWarnings ? undefined : defaultDependentAxisRange,
           }
         : {
-          [key]: newValue
-        };
+            [key]: newValue,
+            // set independentAxisRange undefined
+            independentAxisRange: resetTruncationWarnings ? undefined : defaultIndependentRange,
+            dependentAxisRange: resetTruncationWarnings ? undefined : defaultDependentAxisRange,
+          };
        updateVizConfig(newPartialConfig);
+      if (resetTruncationWarnings) {
+        // close truncation warnings
+        setTruncatedIndependentAxisWarning('');
+        setTruncatedDependentAxisWarning('');
+      }
     },
     [updateVizConfig]
   );
@@ -294,6 +335,7 @@ function ScatterplotViz(props: VisualizationProps) {
   // set checkedLegendItems: undefined for the change of both plot options and showMissingness
   const onValueSpecChange = onChangeHandlerFactory<string>(
     'valueSpecConfig',
+    true,
     true
   );
   const onShowMissingnessChange = onChangeHandlerFactory<boolean>(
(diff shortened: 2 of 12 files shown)


moontrip:
Here are screenshots to explain the issue

- Histogram Viz plot: you can see truncation warning message (switched from Proportion to Count)

histogramviz1

- console.log: in the middle, there exists a discrepancy between (data-driven) `defaultDependentAxisRange` and `vizConfig.dependentAxisRange` and flag/boolean of `truncationConfig` at the last item which is for the comparison of max value in the dependent axis- eventually become the same, though.

histogramViz2

moontrip:
FYI @dmfalke, @jtlong3rd, and @bobular. Although I may need more tests to confirm, I think that I could fix the above issue myself last night. So please ignore my request to help 😃 During that process, I also found another exception that is similar to the above, but I think it is much more manageable. Happy weekend and MLK day, all!

moontrip:
Now axis range controls are available for histogram and barplot Vizs. Several bugs were found/fixed and faceted plot issue was also addressed: screenshots are attached in the following (tested with GEMS1 Case Control study). From now on, this PR needs [the corresponding web-components PR](https://github.com/VEuPathDB/web-components/pull/292):

histogram-viz-control1

barplot-viz-control1

moontrip:
Axis range control for Boxplot Viz is done. A screenshot is attached for faceted plot mode:

boxplot-viz-control1

moontrip:
Scatterplot Viz now has axis range controls for both independent and dependent axes. Several bugs are also fixed.
Cleanup is made for review. A screenshot is attached.

scatter-viz-facetd-control1

bobular:
Will leave a proper review ASAP.  While testing I noticed that bar plot log scale isn't a dependency for the thumbnail (re)generation. There may be more.  Looks really great otherwise! @moontrip

moontrip:
> Will leave a proper review ASAP. While testing I noticed that bar plot log scale isn't a dependency for the thumbnail (re)generation. There may be more. Looks really great otherwise! @moontrip
> 
> Also seems that barplot y-axis range isn't being stored in the saved analysis config?

Thank you for your initial review and comments, @bobular. I will take a look at them.

bobular:
Hi @moontrip - I spotted an issue with code that wasn't in the diff, so here is a standalone comment.

In Scatterplot viz, user clicks on the overlay legend checkboxes were causing the axis ranges to be reset.  I found the `onChangeHandlerFactory` function confusing (why call the argument `resetTruncationWarnings`?) so I figured the following would work better.


```
  // prettier-ignore
  const onChangeHandlerFactory = useCallback(
    < ValueType,>(key: keyof ScatterplotConfig, resetCheckedLegendItems?: boolean, resetAxisRanges?: boolean) => (newValue?: ValueType) => {
      const newPartialConfig = {
	[key]: newValue,
	...(resetCheckedLegendItems ? { checkedLegendItems: undefined } : {} ),
	...(resetAxisRanges ? {
	  independentAxisRange: defaultIndependentRange,
	  dependentAxisRange: defaultDependentAxisRange,
	} : {} ),
      };
      updateVizConfig(newPartialConfig);
      if (resetAxisRanges) {
        // close truncation warnings
        setTruncatedIndependentAxisWarning('');
        setTruncatedDependentAxisWarning('');
      }
    },
    [updateVizConfig]
  );
```

moontrip:
> Hi @moontrip - I spotted an issue with code that wasn't in the diff, so here is a standalone comment.
> 
> In Scatterplot viz, user clicks on the overlay legend checkboxes were causing the axis ranges to be reset. I found the `onChangeHandlerFactory` function confusing (why call the argument `resetTruncationWarnings`?) so I figured the following would work better.
> 
> ```
>   // prettier-ignore
>   const onChangeHandlerFactory = useCallback(
>     < ValueType,>(key: keyof ScatterplotConfig, resetCheckedLegendItems?: boolean, resetAxisRanges?: boolean) => (newValue?: ValueType) => {
>       const newPartialConfig = {
> 	[key]: newValue,
> 	...(resetCheckedLegendItems ? { checkedLegendItems: undefined } : {} ),
> 	...(resetAxisRanges ? {
> 	  independentAxisRange: defaultIndependentRange,
> 	  dependentAxisRange: defaultDependentAxisRange,
> 	} : {} ),
>       };
>       updateVizConfig(newPartialConfig);
>       if (resetAxisRanges) {
>         // close truncation warnings
>         setTruncatedIndependentAxisWarning('');
>         setTruncatedDependentAxisWarning('');
>       }
>     },
>     [updateVizConfig]
>   );
> ```

@bobular thank you for your suggestion. I will take a look at it later 👍

src/lib/core/hooks/computeDefaultDependentAxisRange.ts
```diff
@@ -0,0 +1,251 @@
...
+  useLayoutEffect(() => {
+    if (!data.pending)
+      updateVizConfigRef.current({
+        dependentAxisRange: defaultDependentAxisRange,
+      });
+    else
+      updateVizConfigRef.current({
+        dependentAxisRange: undefined,
+      });
+  }, [data, defaultDependentAxisRange]);
```
bobular, line 124: As far as I can tell, this section is not needed (and no need to pass `updateVizConfig` as a prop).
I have tested with bar, scatter and boxplot, and the dependent axis seems to behave properly, and the vizConfigs are saved properly too.

Please could you describe what doesn't work (if any) if this code is removed?

bobular:
> > Hi @moontrip - I spotted an issue with code that wasn't in the diff, so here is a standalone comment.
> > In Scatterplot viz, user clicks on the overlay legend checkboxes were causing the axis ranges to be reset. I found the `onChangeHandlerFactory` function confusing (why call the argument `resetTruncationWarnings`?) so I figured the following would work better.
> > ```
> >   // prettier-ignore
> >   const onChangeHandlerFactory = useCallback(
> >     < ValueType,>(key: keyof ScatterplotConfig, resetCheckedLegendItems?: boolean, resetAxisRanges?: boolean) => (newValue?: ValueType) => {
> >       const newPartialConfig = {
> > 	[key]: newValue,
> > 	...(resetCheckedLegendItems ? { checkedLegendItems: undefined } : {} ),
> > 	...(resetAxisRanges ? {
> > 	  independentAxisRange: defaultIndependentRange,
> > 	  dependentAxisRange: defaultDependentAxisRange,
> > 	} : {} ),
> >       };
> >       updateVizConfig(newPartialConfig);
> >       if (resetAxisRanges) {
> >         // close truncation warnings
> >         setTruncatedIndependentAxisWarning('');
> >         setTruncatedDependentAxisWarning('');
> >       }
> >     },
> >     [updateVizConfig]
> >   );
> > ```
> 
> @bobular thank you for your suggestion. I will take a look at it later +1

There's a problem with my suggestion - I reckon `defaultIndependentRange` and `defaultDependentAxisRange` should be dependencies of the useCallback but they are not defined until later in the code (for good reasons), so we can't use them here. But they can be used in the callback.  I've played around a bit with closures in the node shell and only half understand what's going on here.  Might need to ask @jtlong3rd for a lesson!

bobular (commented):
I also noted a few things not triggering thumbnail updates and possibly settings not being saved in the config, but I'd like to do a full test again after we've sorted the issues I've raised in this review.

moontrip:
> I also noted a few things not triggering thumbnail updates and possibly settings not being saved in the config, but I'd like to do a full test again after we've sorted the issues I've raised in this review.

@bobular It is quite important part to update vizConfig.dependentAxisRange. Please try to comment out the code part,  make a new plot Viz, e.g., barplot Viz, etc., and see its behavior. For example, changing Count and Proportion, changing Variable (via input variable picker), etc. then it would not work properly.

bobular:
> > I also noted a few things not triggering thumbnail updates and possibly settings not being saved in the config, but I'd like to do a full test again after we've sorted the issues I've raised in this review.
> 
> @bobular It is quite important part to update vizConfig.dependentAxisRange. Please try to comment out the code part, make a new plot Viz, e.g., barplot Viz, etc., and see its behavior. For example, changing Count and Proportion, changing Variable (via input variable picker), etc. then it would not work properly.

Ah yes, I tested it again and I see some behavioural problems.

I still think the useLayoutEffect is unnecessary added complexity and could bite us in the future.  Here is a suggestion to handle it in a more boring way https://github.com/VEuPathDB/web-eda/pull/830

bobular:
> > I also noted a few things not triggering thumbnail updates and possibly settings not being saved in the config, but I'd like to do a full test again after we've sorted the issues I've raised in this review.
> 
> @bobular It is quite important part to update vizConfig.dependentAxisRange. Please try to comment out the code part, make a new plot Viz, e.g., barplot Viz, etc., and see its behavior. For example, changing Count and Proportion, changing Variable (via input variable picker), etc. then it would not work properly.

Ah yes, I tested it again and I see some behavioural problems.

I still think the useLayoutEffect is unnecessary added complexity and could bite us in the future.  Here is a suggestion to handle it in a more boring way https://github.com/VEuPathDB/web-eda/pull/830

moontrip:
> > > I also noted a few things not triggering thumbnail updates and possibly settings not being saved in the config, but I'd like to do a full test again after we've sorted the issues I've raised in this review.
> > 
> > 
> > @bobular It is quite important part to update vizConfig.dependentAxisRange. Please try to comment out the code part, make a new plot Viz, e.g., barplot Viz, etc., and see its behavior. For example, changing Count and Proportion, changing Variable (via input variable picker), etc. then it would not work properly.
> 
> Ah yes, I tested it again and I see some behavioural problems.
> 
> I still think the useLayoutEffect is unnecessary added complexity and could bite us in the future. Here is a suggestion to handle it in a more boring way #830

@bobular thank you for your comment and your suggestion with a new PR. Although I didn't look at/test it, it looks like your approach is the way what we have used at the subsetting's histogram filter, setting dependentAxisRange as undefined. At this stage I may have misunderstood your approach, but if so, please note that histogram filter and Vizs concerning the  dependentAxisRange has one big difference: Vizs should have defined range all the time. It is because you have suggested having it so that all faceted plots show the same dependent ranges

bobular:
> **Update**: Your PR seems to work at faceted plot too! As commented in your PR, will test more and update Vizs if works fine. +1

That's good to hear :+1:

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -570,6 +570,8 @@ function HistogramPlotWithControls({
         selectedRange={selectedRange}
         selectedRangeBounds={selectedRangeBounds}
         onSelectedRangeChange={handleSelectedRangeChange}
+        //DKDK parentComponentName
+        parentComponentName={'HistogramFilter'}
```
bobular, line 574: Hi @moontrip  - I was going to give this PR a quick review, but I noticed it doesn't compile because the `parentComponentName` props are still here.  Does that mean it's not ready for review?  I know this is second-priority stuff due to the b56 deadline but I thought yesterday you said it was done.

bobular: Oh, I see.  You're now working in the branch `pr794-suggestion`

I thought you would merge my suggestions into your branch, and continue there.

But it's OK, I have it working now!

src/lib/core/utils/axis-range-margin.ts
```diff
@@ -6,40 +6,42 @@ export function axisRangeMargin(
   valueType?: string | undefined
 ): NumberOrDateRange | undefined {
   // compute truncated axis with 5 % area from the range of min and max
-  if (valueType != null && valueType === 'date') {
-    // find date diff (days) between range.min and range.max, take 5 % of range, and round up!
-    const dateRangeDiff = Math.round(
-      DateMath.diff(
+  if (valueType != null) {
```
bobular, line 9:
```
if (valueType != null && axisRange != null) {
```
Checking for non-null `axisRange` would avoid the need for all the `?`s in the `axisRange?.min` and `axisRange?.max` below, I think?

bobular (approved):
This looks and works great and I bet we could get this QA'ed before the new delayed release date.

I suggest using the `useVizConfig` hook to remove a bit of repetition - see MapVisualization for how to use it.

Another suggestion is to truncate the decimal places (to 4?) of any range bounds that we have computed (e.g. when extending by +/- 5%).  Here's an example I got with boxplot

image

Don't truncate unless we've done some calculations on the numbers, because the back-end provided numbers should already be display-ready (I think!).

Couple of other inline suggestions.

bobular:
Just going to do some testing with date variables...

bobular:
I think with @danicahelb and colleagues' help we will want to fine tune the "reset custom axis range" behaviour.  At the moment we are being quite conservative - so any variable change causes the custom axes to be reset. Which I think is a wise starting point.

In some combinations, we might not want to: e.g. Histogram, changing the overlay variable, because it's a stacked histogram, the y-axis is unchanged.  But in a barplot, the overlay variable affects the y-axis max counts.

Anyway, I suggest we go ahead with the conservative approach to reseting axis controls (maybe even getting it out for b56) and revisit the fine details if/when anyone has any spare time.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -433,6 +466,8 @@ function ScatterplotViz(props: VisualizationProps) {
       visualization.descriptor.type,
       outputEntity,
       filteredCounts,
+      // get data when changing independentAxisRange
+      vizConfig.independentAxisRange,
```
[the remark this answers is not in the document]
moontrip: Thank you for your detailed reviews and comments, @bobular 👍 Will handle the decimal points. As for this one, actually I added it in the very beginning of scatterplot work concerning axis range control, and forgot to remove it even if I knew that it was unnecessary. Thank you for finding it!

src/lib/core/utils/axis-range-margin.ts
moontrip: yep I think so, but will check and test it!

bobular:
> Just going to do some testing with date variables...

I think I tested all the plots that handled dates and the new controls are working great!

image

moontrip:
sounds great! 🎉 Thank you for your tests and confirmation, @bobular 👍

moontrip:
@bobular I have addressed your comments and tested extensively: couldn't find an issue from my tests. Here are changes I have made:

- use useVizConfig hook for all Vizs
  - clean up unnecessary imports
- Scatterplot Viz: remove dependency of vizConfig.independentAxisRange for data
- Scatterplot Viz: add dependency of vizConfig.independentAxisRange for thumbnail hook (plotRef)
- adjust axisRangeMargin function to remove undefined type assertion
- adjust decimal points to 4
  - made a utility function to deal with it

Please note that Danica is asking the change of truncation color from gray to yellowish ([this issue](https://github.com/VEuPathDB/EdaNewIssues/issues/145 )). I have suggested some colors with screenshots, so it may be included in the corresponding web-components PR  or at the separate one depending on whether this axis range work will be included in b56 or not. It would be great if you can check the issue and suggest a color.

FYI, example screenshots of decimal points with Scatter plot Viz are attached in the following:

# before
scatter-decimal-before

# after (4 decimal points but trailing zeros were ignored)
scatter-decimal-after

src/lib/core/utils/number-decimal-point.ts
```diff
@@ -0,0 +1,9 @@
+// util to return a number with specific decimal points
+export function numberDecimalPoint(
+  value: number,
+  decimalPoint: number
+): number {
+  return Number(
+    Math.round(parseFloat(value + 'e' + decimalPoint)) + 'e-' + decimalPoint
```
bobular, line 7: Hi @moontrip 

I'm worried that this function will fail for numbers that get rendered to "scientific/exponential notation" as strings.

For example, 
```
numberDecimalPoint(500000000000000000, 3)
500000000000000000
numberDecimalPoint(5000000000000000000, 3)
NaN
```
Would something like `Number(value.toPrecision(decimalPoint))` be more robust?

bobular: Except maybe `toFixed` !!

moontrip:
@bobular I just committed the change to use `Number(value.toFixed(decimalPoint))`

bobular:
@moontrip - don't merge this...

moontrip:
> Reason for not merging is that we're not planning to put this out for b56

@bobular Yep! Thank you for reminding this again! 👍 Will also switch the corresponding PR at web-components to a draft

moontrip:
Found a bug on the scatterplot range (https://github.com/VEuPathDB/web-components/issues/311) thus the fix is also applied to this branch

bobular:
Hi @moontrip - I've just done some more testing on this and I think it can merge any time - I don't think it will break anything for Tuesday's demo, but I guess we could wait until just after.

I saw the y-axis range automatically adjust to take in the smoothed mean confidence interval, which is nice!

moontrip:
@bobular Thank you for your further tests and reconfirmation! 👍 Yeah I agree that we can wait after the demo, just in case

Issue #544, opened 2021-10-11
Variable tree: order featured variables based on tree and display order


dmfalke:
Featured variables are currently being ordered using the ordering that the subsetting service produces. Instead, they should be ordered the same as in the tree, using display order.

chowington:
@dmfalke Is this still an issue? It appears to me that the variable tree and the featured variables are using the same ordering. This is both from eye tests on multiple studies and from looking at the code. In the code, I believe the variable tree and the featured variables both derive their order from the ordering that `getStudyMetadata` is doing here: https://github.com/VEuPathDB/web-eda/blob/c0335f832c4bf17ed19f758765fcbebfc9471041/src/lib/core/api/SubsettingClient/index.ts#L50

chowington:
@jtlong3rd See my comment above, thanks!

bobular:
@danicahelb to do QA on this issue

danicahelb:
This still seems to be an issue:

![image](https://user-images.githubusercontent.com/25450900/149237815-c2dbe939-c371-40b2-b346-3ad73c7d742a.png)

danicahelb:
Also an issue in PRISM ICEMR Cohort:

![image](https://user-images.githubusercontent.com/25450900/149238774-2c97012c-a791-4cfb-b69e-2c353399c190.png)

danicahelb:
And PRISM2 ICEMR Cohort

![image](https://user-images.githubusercontent.com/25450900/149238862-49478977-bbbe-454c-a9ab-147115ee01b7.png)

danicahelb:
And SCORE Burundi Mapping Cross sectional

![image](https://user-images.githubusercontent.com/25450900/149239141-7ff238ba-b897-479d-a730-ca1f2919572d.png)

danicahelb:
And SCOORE Rwanda Mapping CX

![image](https://user-images.githubusercontent.com/25450900/149239292-1fb66445-d29b-4ab2-9119-7cdd9433a7be.png)

danicahelb:
And SCORE Zanzibar S. haematobium Cluster Randomized Trial

![image](https://user-images.githubusercontent.com/25450900/149239442-8f866a67-a710-4b11-83d1-a1f4ab4d6d93.png)

dmfalke:
@danicahelb are all of these issues that the order of the featured variables is not the same as the tree? If so, would you say that the order in the tree is correct?

danicahelb:
[...]

PR #800, opened 2022-01-19
Fix incorrect featured variables ordering

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/544

First pass fix; kind of a roundabout solution that seems a bit inefficient. I might try to improve it a bit. Will also clean code up

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "uuid": "^8.3.2"
   },
   "peerDependencies": {
-    "@veupathdb/wdk-client": ">=0.5.4"
+    "@veupathdb/wdk-client": ">=0.5.5"
   },
   "sideEffects": [
     "*.scss"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.1.54",
+  "version": "1.1.55",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.1.53",
+  "version": "1.1.54",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
@@ -76,7 +76,7 @@
     "@veupathdb/prettier-config": "^1.0.0",
     "@veupathdb/react-scripts": "^1.0.1",
     "@veupathdb/tsconfig": "^1.0.1",
-    "@veupathdb/wdk-client": "^0.5.4",
+    "@veupathdb/wdk-client": "^0.5.5",
     "@veupathdb/web-common": "^0.1.12",
     "bubleify": "^2.0.0",
     "http-proxy-middleware": "^1.0.6",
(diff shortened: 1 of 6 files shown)


chowington:
Note to reviewers: the latest commit has two solutions to fix this problem. The result of the first is `sortedFeaturedFields` and the second is `featuredFieldsFromTree`. They should be identical. What isn't identical is the algorithms, and I'd appreciate feedback on which is best.

Solution 1 (`sortedFeaturedFields`) (inefficient):
1. (existing code) Get incorrectly ordered `featuredFields` from `useFeaturedFields`, which processes `entities`
2. (existing code) Get `fieldTree` from `useFieldTree` that processes `flattenedFields` (which is derived from `entities`). `fieldTree` is correctly ordered but does not know which fields are featured (fields are missing `isFeatured` prop)
3. Get correctly ordered featured field terms from `fieldTree` by cross-referencing with `featuredFields` to know which fields are featured
4. Correctly sort `featuredFields` by cross-referencing order from featured field terms obtained from `fieldTree`

Solution 2 (`featuredFieldsFromTree`) (not type safe):
1. Retain `isFeatured` prop for fields in `fieldTree`
2. Use `as` typecasting to silence TypeScript complaining about this
3. Get correctly ordered featured fields directly from `fieldTree`

I will add a new commit with only the solution that's best from these two after getting feedback. Also open to a better solution than either of these!

dmfalke (commented):
Hi @chowington, conceptually, I like the second approach. I don't really like the casting being done, since it can introduce bugs in the future. Would you be up for updating [getTree](https://github.com/VEuPathDB/WDKClient/blob/3d0f8df2c12ac625ae050a7c07fc87e886050dcc/Client/src/Components/AttributeFilter/AttributeFilterUtils.ts#L151) such that it's type definition looks like this:

```typescript
export function getTree<T extends Field>(ontologyEntries: Iterable<T>, options: TreeOptions = defaultTreeOptions): TreeNode<T>
```

I think this change would be self-contained to `AttributeFilterUtils.ts`, but will require changing `FieldTreeNode` to a more generic `TreeNode<T>` interface. This would allow our code in web-eda to have nodes with additional properties.

Let me know what you think!

chowington:
@Dave that sounds good. I'm trying to figure out the details. I have a PR (https://github.com/VEuPathDB/WDKClient/pull/174) on wdk-client that isn't quite working. I'd appreciate your eyes on it over there!

chowington:
Requires https://github.com/VEuPathDB/WDKClient/pull/174

package.json
```diff
@@ -76,7 +76,7 @@
     "@veupathdb/prettier-config": "^1.0.0",
     "@veupathdb/react-scripts": "^1.0.1",
     "@veupathdb/tsconfig": "^1.0.1",
-    "@veupathdb/wdk-client": "^0.5.4",
+    "@veupathdb/wdk-client": "^0.5.5",
```
jtlong3rd, line 79: You'll also want to change the `@veupathdb/wdk-client` `peerDependency` version from `>=0.5.4` to `>=0.5.5`. 

This will signal to `https://github.com/VEuPathDB/ClinEpiWebsite` (the repo where all the ClinEpi frontend assets are bundled together) that the bundled version of `@veupathdb/wdk-client`, declared [here](https://github.com/VEuPathDB/ClinEpiWebsite/blob/fea955c5332fce5ecb694d1ec67e3af952608753/Site/package.json#L14), must be at least `0.5.5`.

(In our various packages that depend on `@veupathdb/wdk-client`, we express the dependency of `@veupathdb/wdk-client` as a `peerDependency`, and not a regular `dependency`, to avoid introducing multiple copies of `@veupathdb/wdk-client`. More info about `peerDependencies` [here](https://flaviocopes.com/npm-peer-dependencies/).)

src/lib/core/types/study.ts
```diff
@@ -231,3 +232,9 @@ export const StudyMetadata = t.intersection([
     rootEntity: StudyEntity,
   }),
 ]);
+
+export type ExtendedField = Field & {
+  precision: number;
+  variableName: string;
+  isFeatured?: boolean;
+};
```
jtlong3rd, line 240: It might also be worth declaring a type alias like `type FieldWithMetadata = Field | ExtendedField`, and using that new alias in place of `Field | ExtendedField`. That should help if we ever change the field metadata.

(I'm not sold on the name `FieldWithMetadata`.)

jtlong3rd (approved):

jtlong3rd (approved):

PR #804, opened 2022-01-20
Bump web-components version

chowington:
Fixes https://github.com/VEuPathDB/web-components/issues/264


bobular (approved):
I hereby approve.  Note that in #782 we go all the way up to 0.11.12!

PR #808, opened 2022-01-20
Update client to use the new hideFrom property

### LINKED ISSUE #801: Update client to use the new hideFrom property.
the possible values (so far) are 'everywhere', 'downloads' and 'variableTree'

jtlong3rd:
Resolves #801 

For example, in Amazonia ICEMR Brazil Cohort, the `Settlement` variable has `hideFrom: ["variableTree"]`, so it is downloadable in the "view and download" modal, but does not appear in the variable trees for the subsetting and visualization tabs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
--- a/src/lib/core/components/filter/MultiFilter.tsx
+++ b/src/lib/core/components/filter/MultiFilter.tsx
@@ -74,7 +74,7 @@ export function MultiFilter(props: Props) {
       // This function uses {entity.id}/{variable.id} to generate a field's term
       // and parent property value. That is not desired here, so we have to do
       // some post-processing to use {variable.id} for those properties.
-      entitiesToFields([entity])
+      entitiesToFields([entity], 'variableTree')
         .filter((field) => !field.term.startsWith('entity:'))
         .map((field) => ({
           ...field,
--- a/src/lib/core/components/variableTrees/MultiSelectVariableTree.tsx
+++ b/src/lib/core/components/variableTrees/MultiSelectVariableTree.tsx
@@ -34,7 +34,7 @@ export default function MultiSelectVariableTree({
 }: MultiSelectVariableTreeProps) {
   const entities = useStudyEntities(rootEntity);
   const valuesMap = useValuesMap(entities);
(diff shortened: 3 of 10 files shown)


dmfalke (commented):
I've taken a cursory look at this, and it looks really good. I'll be more thorough later, but have a question. Is it possible for a non-leaf node to have a hideFrom property set? And if so, how will this impact tree construction? Eg, what do we do if we can't find the parent of a node?

jtlong3rd:
> I've taken a cursory look at this, and it looks really good. I'll be more thorough later, but have a question. Is it possible for a non-leaf node to have a hideFrom property set? And if so, how will this impact tree construction? Eg, what do we do if we can't find the parent of a node?

Good question! I'm not sure if the service permits `hideFrom` to be a set on a non-leaf node. I left a [question](https://github.com/VEuPathDB/web-eda/issues/801#issuecomment-1020467757) on the associated GitHub issue.

As currently implemented, if a non-leaf node has the `hideFrom` property set, the node and all of its descendants will be omitted from the field tree.

(Assuming we agree with this behavior, I should also update `useFlattenedFields` and `useFeaturedFields` to omit _all_ hidden fields, not just the ones which were explicitly marked with "hideFrom.")

jtlong3rd:
> > I've taken a cursory look at this, and it looks really good. I'll be more thorough later, but have a question. Is it possible for a non-leaf node to have a hideFrom property set? And if so, how will this impact tree construction? Eg, what do we do if we can't find the parent of a node?
> 
> Good question! I'm not sure if the service permits `hideFrom` to be a set on a non-leaf node. I left a [question](https://github.com/VEuPathDB/web-eda/issues/801#issuecomment-1020467757) on the associated GitHub issue.
> 
> As currently implemented, if a non-leaf node has the `hideFrom` property set, the node and all of its descendants will be omitted from the field tree.
> 
> (Assuming we agree with this behavior, I should also update `useFlattenedFields` and `useFeaturedFields` to omit _all_ hidden fields, not just the ones which were explicitly marked with "hideFrom.")

@steve-fischer-200 confirmed that "also hide the descendants" is the expected behavior for "`hideFrom` on non-leaf nodes."

`useFieldTree` behaves as expected; I'll also need to update `useFlattenedFields` and `useFeaturedFields` to omit not just fields which were marked with `hideFrom`, but also their descendants.

dmfalke:
This sounds great, @jtlong3rd!

jtlong3rd:
@dmfalke I've now updated `useFlattenedFields` and `useFeaturedFields` to accommodate `hideFrom` on non-leaf nodes.

There will likely be merge conflicts with @chowington's [outstanding PR](https://github.com/VEuPathDB/web-eda/pull/800), but they should be straightforward to resolve.

dmfalke (approved):

PR #829, opened 2022-02-05
Data table modal todos

### LINKED ISSUE #656: Bugs in subsetting results data table modal & subset download files
- [x] Uncomment button to invoke modal
- [x] Update styling of close button for modal (use an X in the top-right corner)
- [x] Update styling of variable selector toggle (possibilities: change label on state change; add a close icon; or make look even more like a drop-down menu)

QA Tasks (each tagged w/ assignee)
- [x] Incorrect URL is hardcoded for the tabular download (@dmfalke)
- [x] Reflect starred variables in the variable selector drop-down. (@dmfalke)
- [x] Allow user to add/remove starred variables from the selector drop-down. (@dmfalke)
- [x] Have the tree expanded by default in the selector drop-down. (@dmfalke)
- [x] Look into the data loading errors mentioned by Danica. (@dmfalke)
- [x] It would be good to have the featured variables box above the variable selector. (@dmfalke)
- [x] See if we can remove the "select all | clear all | expand all | collapse all" footer to give more real estate to the variable tree. (@dmfalke)
- [ ] When a user is not logged on / does not have an access request approved, they cannot open the subsetting results table. Instead they see the popup instructing them to log in or request access. We had previously talked about enabling users to see 20 rows of data when they don't have access. We should discuss what functionality we'd like during an upcoming ux meeting (@dmfalke)
- [x] Re-position the variable selector window so that it is more obviously a pop-over. (@eikonomega)

dmfalke:
This PR addresses the remaining issues in #656.

depends on https://github.com/VEuPathDB/web-http-utils/pull/4
closes #656

### DIFF
--- a/src/lib/core/api/SubsettingClient/index.ts
+++ b/src/lib/core/api/SubsettingClient/index.ts
@@ -1,9 +1,7 @@
 import { array, number, type } from 'io-ts';
-import { memoize } from 'lodash';
 import { saveAs } from 'file-saver';
 
 import { preorder } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
-import { WdkService } from '@veupathdb/wdk-client/lib/Core';
 import {
   createJsonRequest,
   FetchClientWithCredentials,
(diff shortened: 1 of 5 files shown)


dmfalke:
@eikonomega I noticed that the buttons for this feature are not using the same theme as elsewhere on the site. Any idea why?

image

jtlong3rd (approved):

dmfalke:
> @eikonomega I noticed that the buttons for this feature are not using the same theme as elsewhere on the site. Any idea why?
> 

I realized this isn't an issue, since the button won't appear here when we release.

PR #830, opened 2022-02-06
Handle dependent range without further useEffects

bobular:
Suggestion how to avoid the useRef and useLayoutEffect and keep things simple

The main things I did were
1. pass `vizConfig.dependentAxisRange ?? defaultDependentAxisRange` to both the plot component and plot controls - this means we can reset vizConfig.dependentAxisRange to undefined which makes things cleaner.
2. add a new flag to the onChangeHandlerFactory to reset the dependentAxisRange for valueSpec and showMissingness handlers (onVariableChange also resets the dependent axis range, which I think is a good idea)

I also updated the thumbnail dependencies.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
@@ -202,6 +202,7 @@ function BarplotViz(props: VisualizationProps) {
         facetVariable,
         // set undefined for variable change
         checkedLegendItems: undefined,
+        dependentAxisRange: undefined,
       });
       // close truncation warnings
       setTruncatedDependentAxisWarning('');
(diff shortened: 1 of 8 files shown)


src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -210,30 +211,33 @@ function BarplotViz(props: VisualizationProps) {
...
-       updateVizConfig(newPartialConfig);
+    < ValueType,>(key: keyof BarplotConfig, resetCheckedLegendItems?: boolean, resetAxisRanges?: boolean) => (newValue?: ValueType) => {
+      const newPartialConfig = {
+        [key]: newValue,
+        ...(resetCheckedLegendItems ? { checkedLegendItems: undefined } : {}),
+	...(resetAxisRanges ? { dependentAxisRange: undefined } : {}),
+      };
+      updateVizConfig(newPartialConfig);
+      if (resetAxisRanges)
+	setTruncatedDependentAxisWarning('');
```
bobular, line 223: Sorry the formatting is terrible.  The `prettier-ignore` is/was needed for the `< ValueType,>` bit (I don't remember why - I think maybe it upsets the code colouring in my emacs...).  We could try removing the `prettier-ignore` directive.

moontrip:
@bobular I am checking and testing this now and it looks good so far 👍 Will test intensively and if it works fine, then I will update other Vizs.

bobular:
Looks great, but merging this into the parent branch, where I will complete the review :-) 
I would like all the new code in the same PR.

PR #851, opened 2022-02-09
Bump @veupathdb/eda and @veupathdb/components versions

moontrip:
With the custom legend styling work at [web-components](https://github.com/VEuPathDB/web-components/pull/300), both @veupathdb/eda and @veupathdb/components versions are up in this PR.


moontrip:
@jtlong3rd as described above, I need to update web-eda and web-components versions too 😃

Issue #844, opened 2022-02-07
Styling fix needed: pencil, copy and delete icons should be the same color


danicahelb:
Styling issues were identified in the 2/7/2022 EDA UX meeting.

Pencil icon for renaming an analysis is blue, all other icons are teal → these should all be the same color

<img width="316" alt="image" src="https://user-images.githubusercontent.com/25450900/152875262-cded48d0-c49a-4da1-be21-150d36c6e781.png">

TODO: make all one color (teal is fine)

SheenaTomko:
Color looks good now. Marking as complete

Issue #843, opened 2022-02-07
Styling fix needed, reposition copy, delete, and show all filters buttons


danicahelb:
Styling issues were identified in the 2/7/2022 EDA UX meeting.

Copy and trash icons look to be related to the pencil icon (ie, renaming the analysis), as opposed to being related to the analysis as a whole. Also, “Show all filters” button should not be next to the “Share analysis” button

<img width="1409" alt="image" src="https://user-images.githubusercontent.com/25450900/152875620-99152809-aa91-49aa-9c13-4ed322cfd87a.png">

TODO: Move copy & delete icons near the “Share analysis” button and move the “Show all filters” button to the left. The fix does not need to exactly replicate what is on the current beta site, but "Show all filters" should generally be on the left side of the screen and there should be some space between the copy and delete icons (it was suggested to group copy & delete with share)

<img width="690" alt="image" src="https://user-images.githubusercontent.com/25450900/152875655-c2966a1b-5cdf-4bde-b9ad-ce5cc4795da5.png">

SheenaTomko:
The position of the different buttons and links looks good. However, the UX is a little strange and needs discussion. When a user first opens a study, they see only the edit name icon and Share analysis button
<img width="1176" alt="Screen Shot 2022-02-14 at 8 14 44 AM" src="https://user-images.githubusercontent.com/54597319/153881253-9cace0c6-9b22-48b2-aefb-2566b3fcce71.png">

If they start a visualization, the copy and delete icons appear and push the share analysis button to the left
<img width="1187" alt="Screen Shot 2022-02-14 at 8 15 01 AM" src="https://user-images.githubusercontent.com/54597319/153881374-93ec93cb-44ed-4d2e-bdc8-39f8791660a9.png">

If they add a filter, the Show all filters button appears and the copy and delete icons appear and push the share analysis button to the left
<img width="1185" alt="Screen Shot 2022-02-14 at 8 15 22 AM" src="https://user-images.githubusercontent.com/54597319/153881481-234cad7b-4b69-4571-b43c-f543f19fae44.png">

Recommendation: The "Share analysis" button, like the copy and delete buttons, should only appear after the user has either applied a filter or created a visualization

chowington:
@SheenaTomko I think that makes sense! I'll make it happen 👍

SheenaTomko:
Looks great! Marking as complete

PR #853, opened 2022-02-09
Improve analysis heading button styles

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/843
Fixes https://github.com/VEuPathDB/web-eda/issues/844

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.3.9",
+  "version": "1.3.10",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


chowington:
image

eikonomega (changes requested):
Hi @chowington this all makes sense. Just a couple of small things.
1. It looks like this needs to have `main` merged back into it before it is merged itself.
2. Can you update the package version in `package.json` to me, this looks like a bug fix release, so it should just increment the last number of the version by 1.

eikonomega (approved):

PR #854, opened 2022-02-10
Ensure onModalOpen is referentially stable across renders

dmfalke:
I discovered that an infinite render loop also occurs when the modal is open. I added a `useCallback` hook and made sure its dependencies are referentially stable when possible.


bobular:
Functionally working for me :-)

Issue #815, opened 2022-01-12
Subsetting tab: Long original variable name problem


SAK22:
The show more/show less button seems to have disappeared from variables in the Kalifabougou Malaria Cohort.
See screenshots below.
<img width="1275" alt="Screenshot 2022-01-12 at 18 33 52" src="https://user-images.githubusercontent.com/22981342/149202581-7983c66a-2a9f-4f44-8dc7-fc5d069db94c.png">
<img width="1263" alt="Screenshot 2022-01-12 at 18 35 45" src="https://user-images.githubusercontent.com/22981342/149202595-08ea270f-b6cf-4f1c-a506-760c1e20c284.png">

SAK22:
Looks good!

PR #857, opened 2022-02-10
Show more for long list of variable names at Subsetting

moontrip:
This will address #815 

Previously, long list of variable names at Subsetting was treated by introducing Show/Hide more link, but it was targeted to multifacet variables only. Now it is extended to general variable that has more than three list.

Somehow, Kalifabougou Malaria Cohort is not available so I could not test the example case shown in the issue #815, however, I think that this change would work just fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.3.12",
+  "version": "1.3.13",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
--- a/src/lib/workspace/Variable.tsx
+++ b/src/lib/workspace/Variable.tsx
@@ -42,6 +42,9 @@ export function VariableDetails(props: Props) {
         variable,
         groupBy(entity.variables, (variable) => variable.parentId)
       ).length
+    : // array string to array: for generalizing Show/Hide more
+    isJsonString(variable.providerLabel)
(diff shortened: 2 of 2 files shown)


dmfalke:
One thing I noticed is that we are processing `providerLabel` three separate times in this component:
https://github.com/VEuPathDB/web-eda/blob/02ec541993b38f39296f33a6cdb56e10d22498df/src/lib/workspace/Variable.tsx#L40-L119. In all three cases, we are branching on the exact same logic, and we are processing in the exact same way. This should be factored such that we only do this once.

@moontrip I don't know if this is something you want to take on now. If not, we should address this soon, if only to avoid duplicate code. I think the reality is that we will not hit a huge performance penalty, unless we get a lot of successive renders of this component due to incoming props changing. I'm happy to help, if you decide to do it.

jtlong3rd (approved):
Aside from the `providerLabel` DRYing up that Dave indicated, this should be good to merge.

moontrip:
Thank you for your reviews and comments, @dmfalke @jtlong3rd 👍 
Yes I noticed the repetition so I will think of DRYing up 😃

moontrip:
@dmfalke and @jtlong3rd I have made changes to minimize repetition. It would be great if you review this again

dmfalke (approved):
This looks great, @moontrip!

jtlong3rd (approved):

moontrip:
thank you for your quick reviews @dmfalke @jtlong3rd 👍

PR #858, opened 2022-02-11
Use submitAsForm utility

dmfalke:
This PR uses the utility function `submitAsForm`. It also shows how you can append the `Auth-Key` search param to the form url. You can adapt this function to also return an URL that can be used with an `<a>` tag.


eikonomega (approved):
Thank you @dmfalke for all your great work as always.

PR #860, opened 2022-02-14
Reverse the order of entity nodes in diagram, but not in variable tree

### LINKED ISSUE #859: Bug: Community Repeated measures should come directly after Community
The order of entities in the variable tree for WASHb BNG is incorrect. It should always be in this order:

- Community
- Community Repeated Measures
- Household
- Household Repeated Measures
- Participants
- Participant Repeated Measures
- Samples

image

dmfalke:
The ordering of nodes has been stabilized on the backend, when constructing the entity tree. This requires an adjustment to how the tree is traversed when displaying the variable tree and the entity diagram.

fixes #859

### DIFF
--- a/src/lib/core/components/EntityDiagram.tsx
+++ b/src/lib/core/components/EntityDiagram.tsx
@@ -34,7 +34,7 @@ export function EntityDiagram(props: Props) {
       (entity) => entity.children ?? [],
       studyMetadata.rootEntity
     );
-  }, []);
+  }, [studyMetadata.rootEntity]);
   const [lastVariableMap, setLastVariableMap] = useState<
     Record<string, string>
   >({});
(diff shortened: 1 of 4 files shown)


src/lib/core/components/EntityDiagram.tsx
```diff
@@ -22,6 +25,16 @@ interface Props {
...
+  const entityTree = useMemo((): StudyEntity => {
+    return mapStructure(
+      (node, children) => ({
+        ...node,
+        children: children.slice().reverse(),
+      }),
+      (entity) => entity.children ?? [],
+      studyMetadata.rootEntity
+    );
+  }, []);
```
jtlong3rd, line 37: Missing `studyMetadata` dependency?

jtlong3rd (commented):
Aside from one inline suggestion, should be good to merge!

src/lib/core/components/EntityDiagram.tsx
dmfalke: Thank you, yes!

PR #863, opened 2022-02-14
findFirstVariable: search only "scoped" variables

### LINKED ISSUE #861: "findFirstVariable" should respect hideFrom
As initially reported by @SheenaTomko: https://github.com/VEuPathDB/web-eda/issues/801#issuecomment-1028289935.

The cause is that the underlying [findFirstVariable](https://github.com/VEuPathDB/web-eda/blob/721b00964968a6e0c179bd7d024f29b375890551/src/lib/workspace/Utils.ts#L12) utility is searching through all variables, even those which are supposed to be `hid[den]From` the `"variableTree"` scope.

(Note: [I confirmed](https://github.com/VEuPathDB/web-eda/issues/801#issuecomment-1028314980) this bug was also present when we were hiding variables based on whether the variable's `displayType` is `"hidden"`. A more fundamental issue with how we're handling state / derived data is at play here.)

jtlong3rd:
Resolves #861 

Future work: instead of passing a WDK field tree to `findFirstVariable`, pass a variable tree with EDA-specific metadata.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.4.1",
+  "version": "1.4.2",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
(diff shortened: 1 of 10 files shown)


dmfalke (approved):
Great!

jtlong3rd:
@asizemore This is merged now - hope it helps!

PR #864, opened 2022-02-15
Stand alone study page

dmfalke:
This PR introduces a standalone study page route, in the context of the EDA workspace application.

**Details:**
- For private and prerelease, if the user does not have access, they will be redirected from the analysis page to the standalone study page
- For private and prerelease, the "New analysis" and "My analyses" buttons are removed

**Other things to consider:**
- What should be the behavior for importing/sharing? I'm not able to test right now due to instability on the backend, but I believe we will currently show the more traditional modal.

### DIFF
--- a/src/lib/workspace/AnalysisPanel.tsx
+++ b/src/lib/workspace/AnalysisPanel.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import { uniq } from 'lodash';
+import Path from 'path';
 import {
   Redirect,
   Route,
@@ -9,11 +10,9 @@ import {
 } from 'react-router';
 
 // Functions
-import { getAnalysisId } from '../core/utils/analysis';
 import { cx } from './Utils';
 
 // Definitions
-import { AnalysisState } from '../core';
 import { Status } from '../core';
 
 // Hooks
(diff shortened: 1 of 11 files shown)


src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -42,6 +41,12 @@ import NotesTab from './NotesTab';
 import DownloadsTab from './DownloadsTab';
 import { Alert } from '@material-ui/lab';
 import ShareFromAnalysis from './sharing/ShareFromAnalysis';
+import { string } from 'fp-ts';
```
jtlong3rd, line 44: Stray import?

src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -165,174 +181,190 @@ export function AnalysisPanel({
       </div>
     );
   if (analysis == null) return <Loading />;
+  if (approvalStatus === 'not-approved')
+    return <Redirect to={Path.normalize(routeBase + '/..')} />;
```
jtlong3rd, line 185: This approach makes sense to me!

jtlong3rd (approved):
Nicely done!

src/lib/workspace/AnalysisPanel.tsx
dmfalke: Yep! I don't know why this happens. It's usually when I'm making a new type definition. Somehow, ts thinks I want a value instead of a type.

dmfalke:
```suggestion
```

src/lib/workspace/AnalysisPanel.tsx
dmfalke: :smile:

PR #867, opened 2022-02-17
Handle entities without categories

asizemore:
After learning that mbio has some entities that have no categories (and sometimes no variables!), we looked for a way to gracefully handle the situation.

This PR implements a change that searches all roots, instead of just the first root, for a good variable (credit @jtlong3rd and @dmfalke). Additionally, if the root is not a category, then it is a variable and so should be returned as-is. 

@jtlong3rd @dmfalke is there a reason we wouldn't want to include these top level variables (non-category variables that sit directly under the entity)? It would be great to fix this problem for mbio, but if we're breaking how we expect the data to look, i'm happy to wait for the data to be reloaded without this entity.

### DIFF
--- a/src/lib/workspace/Utils.ts
+++ b/src/lib/workspace/Utils.ts
@@ -24,7 +24,7 @@ export function findFirstVariable(variables: VariableTreeNode[]): Variable {
     throw new Error('Tree is broken: cannot determine root nodes.');
 
   // Traverse first branch of tree and find first non-category node
-  // If the branch only consists of the root, and the root is not a category, return the root variable.
+  // If the root is not a category, then it is a variable. In this case, return the root variable.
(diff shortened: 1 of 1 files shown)


dmfalke:
Hi @asizemore! I just approved https://github.com/VEuPathDB/web-eda/pull/863, which will likely obviate this PR. When @jtlong3rd shared a link to this PR, I didn't realize it was awaiting my review! I think this PR will also allow for entities with only a single child. If you find this to be the case, then please close this PR. Thanks!

asizemore:
Will do! Thanks so much @dmfalke!!

asizemore:
Empty entities and those with a single child were removed from the mbio data. Closing this issue.

Issue #575, opened 2021-09-29
Visualizations: documentation on variable constraints in the “Learn more” link is out of date


danicahelb:
using the WASH-b Kenya study, I made a box plot of LAZ (participant repeated measure entity) for each study arm (community entity). I'd like to overlay this by either "study timepoint" or "diarrhea in last 7 days" (both under participant repeated measure entity) but I can only select overlay variables in the community entity

![image](https://user-images.githubusercontent.com/25450900/135361826-6c093b9c-7a4f-4820-9663-6cf6f566999e.png)

danicahelb:
Variable constraint documentation has been added, but it is out of date now that we are no longer inferring data shape. The text needs to be updated:

<img width="1080" alt="image" src="https://user-images.githubusercontent.com/25450900/153089784-368278be-dad1-4938-8028-e88da4e1b31d.png">

SheenaTomko:
Moving this to failed qa for now because I made some changes to the text at https://github.com/VEuPathDB/web-eda/blob/variable-constraints/src/lib/core/components/docs/variable-constraints.tsx. 
Danielle will review and submit the pull request.

SheenaTomko:
@d-callan Now when I try to click on the "Learn more" link, I don't get the expected pop-up with the documentation on variable constraints.

d-callan:
its just worked for me on qa.restricted, and appears to have your updates there.. where did you try and what did you get?

danicahelb:
it works for me too. @SheenaTomko please provide more info on how to recreate. thanks!

SheenaTomko:
Closing this. Reviewed the page that Danielle provided a link for and the text looks good. Created a separate ticket for the issue with the pop-up not appearing (so far no one else has been able to reproduce)

https://github.com/VEuPathDB/EdaNewIssues/issues/162

PR #868, opened 2022-02-17
Variable Constraints Documentation

d-callan:
thought id make a PR just as a place to discuss.. i wonder primarily:

1. if there is anything else we need to add/ update (@danicahelb @asizemore others?) 
2. if we can talk about the styling of this document. i have this general feeling it could be improved but no specific feedback.. (@asizemore @eikonomega others?)


d-callan:
actually another question.. if mbio plans to use this page as well, do we want clinepi in the title?

asizemore:
In terms of styling, i think the simplicity is nice. The only thing i'd consider adding is a line or two to break up the page. Specifically between the tldr and the rest of the page. 

Larger margins on the left and bottom would be a big help i think, and keep this [page](https://beta.clinepidb.org/ce.beta/app/workspace/analyses/documentation/variable-constraints) more consistent with other help [pages](https://qa.clinepidb.org/ce.b19/app/static-content/ClinEpiDB/tutorials.html). I'd also try making the plot titles h3 instead of h2, and increasing the paragraph text size a tad.

d-callan:
@dmfalke thoughts on making this page available under the help tab as well? is that all static content stuff?

d-callan:
resolves #575

dmfalke (approved):

Issue #654, opened 2021-10-25
GEMS1A subsetting histogram may not render at very small minimum bin width


bobular:
GEMS1A
Subsetting
Participant repeated measures: Age (months)

If you reduce the binwidth to the minimum 0.062 the spinner seems to spin for ever (I waited a minute or so)

Update: subsetting is [calculating its own min/max allowed values](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/core/components/filter/HistogramFilter.tsx#L789) for the binWidth slider, while the visualization is getting them (more sensible values) from the back-end response.

IIRC the `distribution` endpoint that drives subsetting is stripped down and does not calculate things like this for a reason.

chowington:
This is still happening as of today. @bobular Do you see this as a frontend or backend problem?

chowington:
Our fontend logic calculates the min bin width as the width that would give us 1,000 bins, which is currently far too many for this particular variable endpoint. The smallest bin width for this variable that doesn't cause QA to crash is somewhere around 1, which equates to about 50 bins given this variable's range. I'd hate to force all histograms to have at most 50 bins; ideally we should have a smarter way to calculate the min bin width since it's not currently given to us by the backend for the subsetting histogram. @d-callan Do you have any input on this?

bobular:
Update: in https://github.com/VEuPathDB/web-eda/pull/874 I have added integer-based bin slider constraints.

I don't see any performance problems at minimum bin width (original 1000 bin calculations) with floating point variables, such as GEMS1A -> PRM -> Weight, Height, BMI etc.  Same amount of data (35,187 Participant repeated measures have data for this variable).

bobular:
Resolved by #874

sweverschulman:
In testing this ticket - I noticed there are no performance issues. The bin width is not allowed to go lower than 1. Please @bobular confirm this functionality is correct. If so, we can move this ticket to complete

PR #869, opened 2022-02-17
Drastically reduce max number of bins in histogram filter

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/654 (primarily)
Fixes VEuPathDB/web-eda#873 (incidentally)

This is far from an ideal solution. Please see and contribute to conversation on the above ticket  :)

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -794,8 +794,9 @@ function computeBinSlider(
     case 'number': {
       const { min: rangeMin, max: rangeMax } = range as NumberRange;
       const rangeSize = Math.round((rangeMax - rangeMin) * 100) / 100;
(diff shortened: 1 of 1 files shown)


bobular (commented):
This is pretty drastic!
From up to 1000 bins down to 60...
I tested with 200 using LLINEUP  Participant->Age (there are >100k data points) and it seemed fine.
But I guess I should try more than one study...
(Have done so, and continued discussion on [issue ticket](https://github.com/VEuPathDB/web-eda/issues/654#issuecomment-1043586412))

d-callan:
these are presumably integers.. can we start by just saying integers cant have a bin width smaller than 1 and see where that gets us perhaps? seems like we should be doing that anyway, so if it also fixes this issue i think thatd be a win.

d-callan:
ive since seen #874 and favor that to this

Issue #590, opened 2021-10-09
General: revisit back arrow functionality


danicahelb:
I often hit the back button because I want to go to the last screen i looked at (ie, go from viz tab to subset tab). i know i don't _need_ to do this, but using the back button in this way seems more intuitive than bouncing me entirely out of the analysis

Edit (Connor):
Per conversation with Danica, the following should be "backable"
- [x] Analysis tabs (study details, subsetting, visualize, notes)
- [x] Changing entities via the entity diagram
- [x] Changing variables via the variable tree
- [x] Going into/out of an existing visualization
- [x] Going into/out of the new viz selector screen

chowington:
@danicahelb I tested this, and it seems undoing analysis tab navigation is an easy fix; undoing navigation into and out of visualizations is not so easy. What are the specific reqs for this ticket?

danicahelb:
Hey @chowington i'm not sure what you mean by "analysis tab navigation". did you mean undoing (pressing back on) **_browse & subset tab_** (not analysis tab?) navigation? 

If I am in the browse & subset tab, I'd want the back arrow to either remove filters I just applied to the current variable or, if i hadn't applied filters but navigated between variables I would want it to take me to the last variable I looked at. It is very annoying that the back arrow currently completely bounces me out of my current analysis.

If I am clicking through tabs, I'd like the back arrow to bring me to the place i last visited in the tab. For example, if I am looking at "Study timepoint" on the browse & subset tab and then I clicked onto the notes tab, I'd want the back arrow to bring me back to the "study timepoint" variable on the browse & subset tab (I'm not sure if this is one of the things you indicated was difficult).

I guess in general, I don't want the back arrow to bounce me out of my analysis unless the last thing I was doing was looking at something else on the site (outside of my current analysis)

Maybe we should find a time to chat?

bobular:
I've been testing Connor's branch.

And now reading this thread.

@danicahelb wrote
> If I am in the browse & subset tab, I'd want the back arrow to either remove filters I just applied to the current variable

This is a major feature request that we can't just add to this ticket lightly (short version: filters have never been in the URL so it will not be trivial to implement this requested behaviour, because back/forward button browser history uses the URL)

What Connor has done already seems to avoid "backing out of the analysis entirely" which I think is a good step forward :-)

chowington:
[...]

PR #870, opened 2022-02-17
Fix back button functionality

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/590


src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -32,6 +39,18 @@ import { GeoConfig } from '../../types/geoConfig';
 
 const cx = makeClassNameHelper('VisualizationsContainer');
 
+// Doesn't seem to work
```
bobular, line 42: Do you mean the TS typing doesn't work?

chowington: The typing works fine. This component simply doesn't fix the problem I wanted it to!

chowington:
Do test this "in anger" (@bobular ;)), as in my own earlier testing, the `locationKeys` and `pathnamesRef` arrays would sometimes get into weird states, opening up the potential for unexpected history navigation behavior. That's why I added the sanity checks to the code. My latest testing seemed to work well though.

bobular (approved):
Extensively tested, including the "back into new" scenario.

Looks nice - can't find any problems :-)

dmfalke:
Apologies, I started testing the before I went on vacation and found a small bug. I thought I made a comment about it, but I don't see it. I'll provide more details soon.

dmfalke (approved):
I'm not able to reproduce the issue I remember hitting (that's what I get for not reporting it immediately :slightly_smiling_face:), so I think this is good to go! Nice work!

src/lib/workspace/WorkspaceRouter.tsx
```diff
@@ -62,6 +63,76 @@ export function WorkspaceRouter({
...
+
+          // Sanity check
+          if (!(pathnamesCursorRef.current > 0)) return;
+
+          pathnamesCursorRef.current--;
+
+          const lastPathname =
+            pathnamesRef.current[pathnamesCursorRef.current + 1];
+          const newAnalysisRegex = /eda\/.*\/new\/.*/;
+          const savedAnalysisRegex = /eda\/[^/]*\/(?!new\/)([^/]*)\/.*/;
```
jtlong3rd, line 114: The base url for the EDA workspace isn't necessarily `/eda` (for example, the ClinEpi website uses `/workspace/analyses`), so these regexes will need to be dynamic based on the base url.

jtlong3rd (commented):
Just one inline suggestion, and then this is likely good to merge!

src/lib/workspace/WorkspaceRouter.tsx
dmfalke: Good catch! :detective:

chowington: @jtlong3rd Thanks for the info! How does it look now?

jtlong3rd: The regexes look great! Just to be safe, could you also add `path` to the effect dependencies?

jtlong3rd (approved):

PR #874, opened 2022-02-17
Integer-based bin slider limits for integer vars, reduced precision of limit labels, max binwidth now half what it was

### LINKED ISSUE #871: Non-integer (less than 1) binWidths cause `distributions` back end to time out/fail
Tested with LLINEUP

"Persons" variables (e.g. Persons <5 years living in house) - only integer binwidths work - non-integers cause a proxy error (and cause performance problems for subsequent queries if you send too many of these).

Solution: disallow client from requesting non-integer binWidths for integer variables.

May require some slider component work to enforce integer values from both slider and text box.

### LINKED ISSUE #873: Subsetting histogram filter: min/max for binWidth slider has too much precision
See image
image

Visualizations round to 4 decimal places. Seems wise here.

bobular:
Resolves #871 
Resolves #873 

@chowington - note that variables like GEMS1A PRM Age are `type==='integer'` - so this PR will take precedence over your changes in #869 

Actually I can't find a problem with 1000 bins with floating point variables but will comment on your ticket about that.

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -791,6 +791,13 @@ function computeBinSlider(
       return { min: 1, max: 60, step: 1 };
     }
     case 'integer':
+      const { min: rangeMin, max: rangeMax } = range as NumberRange;
+      const rangeSize = rangeMax - rangeMin;
+      const maxBins = 100;
+      const stepSize = Math.floor(rangeSize / maxBins);
+      const min = stepSize < 1 ? 1 : stepSize;
+      const max = Math.floor(rangeSize / 2);
+      return { min, max, step: min };
     case 'number': {
       const { min: rangeMin, max: rangeMax } = range as NumberRange;
       const rangeSize = Math.round((rangeMax - rangeMin) * 100) / 100;
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -786,6 +786,7 @@ function computeBinSlider(
   type: HistogramVariable['type'],
   range: NumberOrDateRange
 ) {
+  const [minBins, maxBins] = [2, 1000];
   switch (type) {
     case 'date': {
       return { min: 1, max: 60, step: 1 };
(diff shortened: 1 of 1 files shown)


src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -791,6 +791,13 @@ function computeBinSlider(
       return { min: 1, max: 60, step: 1 };
     }
     case 'integer':
```
ryanrdoherty, line 793: Two things: For consistent code styling, maybe put curly braces around the integer case code (to match the others)?  And also, I'm curious why we always start with the min size rather than some aesthetically pleasing number of bins e.g. 10? (assuming it doesn't violate the min/max).

bobular: Thanks Ryan - completely missed the curly braces - surprised it worked without!

Actually we don't always start with min bin size.  The recommended binWidth is supplied with variable metadata and we use that.  These calculations just provide the limits and step-size for the slider.

d-callan:
seems a little odd to me to allow different numbers of bins depending on data type..

chowington:
> > i also wonder if, rather than having similar but not quite identical logic in two places, we might instead consider adding the recommended binslider specs to our variable metadata along w the recommended bin width? then the logic would be guaranteed to be consistent, and only have to exist in one place (plot.data) ?
> 
> Hi @d-callan I agree with adding the binslider specs to the variable metadata - seems like a longer term solution though (post b56 at least!)
> 
> For now we could have consistent 1000-bin logic throughout and integer vars have a minimum binwidth of 1?
> 
> What do you think, @chowington ?

That sounds great! I think it would fix our Gems1A Age problem, as bin width 1 works for it 👍 Much better than imposing a strict limit on the max number of bins!

chowington:
As an update, Ryan found the problem was the backend flooring bin values between 0 and 1 for integer variables to 0, causing an infinite loop. Totally makes sense based on what we all were seeing! This fix takes care of it quite well from our side :)

d-callan:
one other thing to consider, i think w this logic its still possible to have non-integer binSlider values that are greater than 1?

bobular:
@ryanrdoherty @chowington or @d-callan - please review at your convenience :smiley:

ryanrdoherty (approved):

bobular:
Thanks for all the feedback guys.

As soon as this is deployed, we should see fewer OOM errors on the back end!

PR #876, opened 2022-02-18
My analyses: fix search bar when HTML tags are in study name, also solve endless datasets request

### LINKED ISSUE #780: Studies with HTML tags, such as italicised species, are not filtering properly in the "My analyses" list
For example 
Study **SCORE Seasonal Transmission S. haematobium Cluster Randomized Trial**

Make a new analysis - name it or change a filter or something so it saves - then click "My analyses"
image

The list will be empty, but it shouldn't be!

bobular:
Resolves #780 

Also resolves a hitherto unreported bug (endless loop of study list requests)

The only issue with this solution is that if a user starts typing HTML, they find that it is impossible.  Alternatively we can keep the HTML tags in the search box, and add `'displayNameHTML'` to `searchableDatasetColumns`.

### DIFF
--- a/src/lib/core/hooks/study.ts
+++ b/src/lib/core/hooks/study.ts
@@ -96,10 +96,11 @@ export function useWdkStudyRecord(datasetId: string): HookValue | undefined {
 
 const DEFAULT_STUDY_ATTRIBUTES = ['dataset_id', 'eda_study_id'];
 const DEFAULT_STUDY_TABLES: string[] = [];
+const EMPTY_ARRAY: string[] = [];
 
 export function useWdkStudyRecords(
-  attributes: AnswerJsonFormatConfig['attributes'] = [],
-  tables: AnswerJsonFormatConfig['tables'] = []
+  attributes: AnswerJsonFormatConfig['attributes'] = EMPTY_ARRAY,
+  tables: AnswerJsonFormatConfig['tables'] = EMPTY_ARRAY
 ): StudyRecord[] | undefined {
   return useWdkService(
     (wdkService) =>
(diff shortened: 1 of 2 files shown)


src/lib/core/hooks/study.ts
```diff
@@ -96,10 +96,11 @@ export function useWdkStudyRecord(datasetId: string): HookValue | undefined {
 
 const DEFAULT_STUDY_ATTRIBUTES = ['dataset_id', 'eda_study_id'];
 const DEFAULT_STUDY_TABLES: string[] = [];
+const EMPTY_ARRAY: string[] = [];
 
 export function useWdkStudyRecords(
-  attributes: AnswerJsonFormatConfig['attributes'] = [],
-  tables: AnswerJsonFormatConfig['tables'] = []
+  attributes: AnswerJsonFormatConfig['attributes'] = EMPTY_ARRAY,
+  tables: AnswerJsonFormatConfig['tables'] = EMPTY_ARRAY
```
jtlong3rd, line 103: Good catch!

src/lib/workspace/AllAnalyses.tsx
```diff
@@ -94,7 +94,8 @@ export function AllAnalyses(props: Props) {
   const classes = useStyles();
 
   const queryParams = new URLSearchParams(location.search);
-  const searchText = queryParams.get('s') ?? '';
+  const searchParam = queryParams.get('s') ?? '';
+  const searchText = stripHTML(searchParam); // matches stripHTML(dataset.displayName) below
```
jtlong3rd, line 98: If we go with this solution over the alternative solution, let's `useMemo`-ize `searchText`.

jtlong3rd (approved):
I don't have strong feelings about this solution vs. the alternative solution! I suspect that our typical user won't be trying to type HTML into the search bar, so this solution is likely OK.

Issue #839, opened 2022-02-08
Styling of subsetting results modal


danicahelb:
- [x] Position of the "Download" and "Select variables" buttons: "Download button should be in the upper right hand side of the modal (where "Select variables" is currently). "Select variables" button should be moved to the left, near the entity name.

- [x] "Select variables" button should not be filled in but instead should look more similar to the "My analyses" button on Ann's latest study card mockups.

danicahelb:
Steve, via slack:
- the Select button should be a standard action button
- the Downloads button should be a "high-attention" button

danicahelb:
As discussed in 2/14/2022 EDA meeting:

- [ ] Please rename "Select variables" button to "Add columns"

danicahelb:
looks good, closing this ticket

PR #878, opened 2022-02-18
Subset datagrid ux tasks  2022 02

### LINKED ISSUE #656: Bugs in subsetting results data table modal & subset download files

dmfalke:
closes #656

- When a variable has units, the units should be displayed in the column header for that variable (ex: "Hemoglobin (g/dL)" in LLINEUP)
- Always show all relevant ID columns (Participant ID, Household ID, etc) even when no variables have been added to the table
- When just ID columns are in the table (ie, when no variables have been added), please display the text: To configure this table, click on the "Add columns" button. above the table (Note that the "Select variables" button is being changed to "Add columns" in this https://github.com/VEuPathDB/web-eda/issues/839)
- Never pre-populate the table with any variables (including featured and starred variables). Default view of the subsetting results table should just show columns for the relevant IDs
- In https://github.com/VEuPathDB/web-eda/issues/839, we indicate that the "Select variables" button will be moved to the left of the entity name. The variable selector tree should open below this button. Please position the variable selector tree so that we can see the table behind it (We should see the table on top of the variable selector box and also to the left and right of it)
- If I am in a new analysis and I star one of the featured variables from the browse/subset variable tree, I get an error message when I click "View and download"
- All featured variables appear in the featured variables box in the variable selector for every entity, even when they do not belong to a given entity. There is a check box next to each of the featured variables, making it seem as though I can add these variables to my results table. Please grey out and remove the check box from featured variables that do not belong to the entity the user has selected.


dmfalke:
I just discovered that we're making two requests when the modal is open, and the user has selected variables. I will try to address this before merging.

dmfalke:
I think this latest commit will do it.

dmfalke:
@eikonomega just want to alert you to this PR, since it likely overlaps with your related work on the **Downloads** tab. Feel free to review, but I will likely go for a merge sooner rather than later :slightly_smiling_face:.

jtlong3rd (approved):

dmfalke:
@eikonomega I'm planning to merge this shortly. I'm happy to work out any merge conflicts that you might run into on https://github.com/VEuPathDB/web-eda/tree/shifu__758__downloads_tab

PR #894, opened 2022-02-24
Allow variables to be (un)checked by clicking on the name.

### LINKED ISSUE #891: In Subsetting download modal custom column selector, clicking on text name of variable doesn't check the corresponding box
This is despite the hover tooltip that says "Select this variable", so it's a bit misleading.

dmfalke:
This is a quick fix to allow variables to be (un)checked by clicking on the name, when in multiselect mode.

Note that the tooltip still says "Select this variable", even if the variable is checked. Addressing this will be a little more involved.

fixes #891

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.6.4",
+  "version": "1.6.5",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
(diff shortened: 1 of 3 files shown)


bobular (approved):
Looks/works great to me.

One easy temporary fix to the tooltip issue would be to use the text

`Select/deselect this variable`

dmfalke:
That's only true for multiselect mode. I can address this a little differently.

dmfalke:
@bobular I pushed a change, if you want to take a look. I removed the tooltip when in multiselect mode.

Issue #897, opened 2022-02-24
Bug in visualization constraints pop up


SheenaTomko:
On macbook pro, using Google Chrome, using qa.clinepidb.org (which redirects to https://qa.clinepidb.org/ce.b19/app/, which I don't know if it's an issue that the URL says b19 rather than b21).

I clicked on the "Amazonia ICEMR Brazil Cohort", went to the Visualize tab, opened a Mosaic plot: 2x2 table, clicked on the X-axis dropdown, moused-over "Only show compatible variables", moused-over "Learn more" and took the following screenshot, then tried to click "Learn more", but the expected white box containing the visualization constraints information never appeared.

Tested with a few different studies, tried adding a subset before mousing over, tried adding a variable to the x-axis before clicking learn more, tried clicking Learn more from the drop downs of other parameters (overlay and y-axis) in other visualization types. I was able to see the pop up once, but not able to reproduce how I saw it.

![Screen Shot 2022-02-24 at 9 47 39 AM](https://user-images.githubusercontent.com/54597319/155559333-bff46203-e5c9-47a1-a389-825e5ff93d46.png)

moontrip:
@dmfalke FYI I could confirm this as well: actually all variable tree selector at Viz had the same issue. Looked at some relevant codes but didn't notice specific issue at a glance. Thus suspected that it might be related to material-ui's issue and noticed a possibly related one via googling. 

I tried it and it seemed to resolve the issue. Add a new prop at <HtmlTooltip> in the VariableList component, `disableFocusListener={true}`

More details can be found in https://github.com/mui/material-ui/issues/30973

dmfalke:
@moontrip thank you for looking into this. I was able to reproduce and it does seem your solution works. There are some other strange interactions going on with the variable selector pop up and the documentation modal. Fixing might be somewhat involved, so I'm wondering if we should open the documentation in a new tab for now? It's possible we can fix the modal before release. 

@SheenaTomko Thoughts on this?

SheenaTomko:
@dmfalke I'm fine with opening the documentation in a new tab for now

dmfalke:
Thanks, @SheenaTomko. I will make a new issue to temporarily remove the modal behavior.

PR #903, opened 2022-02-25
resolving link inside material-ui tooltip

moontrip:
This would possibly resolve the issue #897 
Detailed explanation is explained at my comment in the ticket, https://github.com/VEuPathDB/web-eda/issues/897#issuecomment-1050119406

I expected that Dave would make this kind of PR with my suggestion but he might have been hectic so I made this instead 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.6.7",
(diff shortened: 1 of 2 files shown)


dmfalke (approved):

moontrip:
This is also related to #905

PR #916, opened 2022-02-28
handle exceptions in legend (barplot and boxplot Vizs)

### LINKED ISSUE #915: Visualization: missing data element seems to cause issues on legend
Found this bug by chance.
While reviewing #911 , which was initially linked to https://github.com/VEuPathDB/plot.data/issues/134 incorrectly, I realized that there existed some exceptions (used the 134's variables in part) that threw error popup, seemingly related to legend.

Here are some example cases that showed errors (barplot viz with GEMS1 Case Control): 
a) Main: Country, Overlay: Household Wealth Index, Categorical, Facet: Floor material
b) Main: Age group, Overlay: Household Wealth Index, Categorical, Facet: Drinking water container

Will examine the issue and try to resolve it

moontrip:
This would resolve #915 

The issue was that somehow backend does not return certain data element in facet case (data.value.facets[].data.series[]) when all value array elements (data.value.facets[].data.series[].value) are undefined (I presume). It is just fine for plotting, but causes issue when making the custom legend: array functions like some() and map() would throw error in that case as they do not accept undefined. To resolve, I made a condition to check the case.

Note that this solution is intended to resolve issues found in barplot Viz as described in the ticket #915, but since boxplot Viz utilizes similar logic, I have added the same condition in there as a safety net, even if I could not find the case. Histogram and Scatterplot use different logic and data formats and in part it was already treated so only barplot and boxplot Vizs are changed.


bobular:
Hi @moontrip 
Isn't the faulty legend caused by a faulty back end response? See https://github.com/VEuPathDB/plot.data/issues/134

I will check out your web-eda branch to see if it also solves it

bobular:
Yeah, I'm still getting missing labels in the legend when on `handle-legend-exceptions` branch


image

moontrip:
@bobular Thank you for your review and tests. Actually, the missing/empty label at the legend is not what I tried to resolve in this PR: there was no data regarding the name so it is certainly backend issue as you linked.  

What I was trying is to avoid a case of missing data (array element) when generating legend.  So if you can see plots under this branch, then it means it is resolved. If you test the same case at other branch like main, then you would get error popup.

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -398,9 +398,12 @@ function BarplotViz(props: VisualizationProps) {
               : data.value?.facets
                   .map((el: { label: string; data?: BarplotData }) => {
                     // faceted plot: here data.value is full data
-                    return el.data?.series[index].value.some(
```
bobular, line 401: I'd suggest the following. I think it's more readable?

```
                    return el.data?.series[index]?.value.some(
```

bobular (approved):
I have validated the behaviour (I didn't facet before, sorry) and approve these fixes @moontrip - it's always good to catch array indexing problems.

See my minor suggestion to condense the code a bit.

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
moontrip: @bobular thank you for your reviews and tests. I agree with your suggestion so I adopted it and made commit 👍

PR #917, opened 2022-02-28
Fix documentation modal

### LINKED ISSUE #896: Documentation: Various issues with modal and variable selector interactions
There are a few issues with how the documentation modal and the variable tree selector interact. Both are effectively "modal" components, and some of their behaviors are inconsistent with one another.

1. The documentation modal is appearing beneath the variable selector. It should be on top of it.
2. The window's scrollbar disappears if the variable selector is closed before the documentation modal. These components are using different libraries for the modal behavior, and it seems that some logic is getting tripped up. Ideally, the documentation modal would prevent interaction with anything beneath it, include key presses. It may be that the documentation modal is not getting the proper focus, or it may be that it should be using a "capture" event, and prevent the event from propagating further down the DOM tree.

dmfalke:
In this PR, the `Modal` component from material-ui is used for documentation, rather than the `Modal` component from `@veupathdb/coreui`. Since the `VariableTreeDropdown` component also uses a material-ui modal-like component, this addresses interop issues.

fixes #896

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.7.0",
+  "version": "1.7.1",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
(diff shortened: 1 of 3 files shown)


moontrip (approved):
@dmfalke yes I thought of the same thing. Your fix to match the origin of modals seems to resolve the conflict(s) - tested 👍

Issue #311, opened 2021-08-16
Use plural form of entity name where appropriate


dmfalke:
The property `displayNamePlural` should be used where appropriate:
- Entity diagram
- Count labels
- Filter labels

Not sure about the variable tree.

danicahelb:
Plurals haven't been implemented in 3 places:

1. Browse/subset histogram legend labels (ex: "Subset of Household" and "Household")
2. Browse/subset histogram mouseover labels (but I don't think we even need these labels, the overall histogram label may be good enough.)
3. Mouseover information for subset table download

![image](https://user-images.githubusercontent.com/25450900/132931724-2cde793c-60c1-4e4b-bf70-9848d61d21be.png)

![image](https://user-images.githubusercontent.com/25450900/132931735-8440637b-0002-4eb0-87c5-8bbb6a0c54e4.png)

danicahelb:
I created https://github.com/VEuPathDB/EdaNewIssues/issues/120 to remove the text labels from the histogram mouseover, so don't worry about fixing item 2 above

danicahelb:
looks good, thanks!

PR #921, opened 2022-03-01
fixing bug on scatterplot range

moontrip:
This resolves [#311 at web-components](https://github.com/VEuPathDB/web-components/issues/311)

I have examined this issue carefully, perhaps too deep unnecessarily, but found that there was an error to use Math.min/max. Simply put, Math.min was wrongfully used to calculate Math.max. 

Here is a screenshot after fix: used the same example shown in the ticket and now it shows correct dependent range (see Olyset Plus plot that looks like a fish 😄)

LLINEUP scatterplot range bug

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.7.2",
+  "version": "1.7.3",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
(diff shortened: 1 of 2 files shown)


jtlong3rd (approved):
I can't begin to tell you how many times I made "min" and "max" typos in my (math) grad school days!

moontrip:
@jtlong3rd thank you for your quick review! 👍 Ironically, I spend several hours to identify the culprit: also it is too small to find in my vscode 😄

PR #944, opened 2022-03-03
Adjust min/max for selectedRangeBounds in Histogram subsetting

jernestmyers:
Relates to Issue #385

Changed the min/max of selectedRangeBounds to refer to the first bin's binStart and the last bin's binEnd since the desired behavior is to be able to subset the "visible" range (versus the actual data's range).

When testing with ISO dates, the selection is not quite capturing the maximum bin. I am looking into this now.

image


jernestmyers:
@bobular FYI, I haven't pushed updated code since we last talked about this. @dmfalke seems to have a different opinion on how the subset range relates to the blue selection range.

dmfalke:
@bobular Jeremy explained to me your reservations about the selected range max exceeding the data max. I don't entirely agree that it's problematic. I believe your primary concern has to do with cases where the x-axis represents something like probability or percentage. We've hit similar situations in a genomics context, and I think we decided that we need to handle this type of data differently.

Do you know if we have cases like this presently, or on the horizon? Maybe this needs to be discussed w/ the dataviz folks?

bobular:
If you mean are there percentages in the data already, then yes, a few

image

but note that that one already goes over 100 due to curation/processing issues 😬

dmfalke:
Thanks @bobular. What do you think about my assertion that the selected range inputs should match the bounds of the selection rectangle, even if it exceed the extremes of the data?

bobular:
> Thanks @bobular. What do you think about my assertion that the selected range inputs should match the bounds of the selection rectangle, even if it exceed the extremes of the data?

That would be OK, I suppose.  In actual fact there is more of a problem with my proposed approach of adjusting the blue rectangle to the end of the right-most bin... This is wholly inappropriate if the user subsequently changes the bin-width.

@jernestmyers - go ahead with your original approach!

jernestmyers:
The latest commit includes the original solution without all the console logs. Regarding the screenshot above, am I correct to assume that we want the blue selector to be able to select the very last day when dealing with date objects?

bobular (approved):
This is a good fix, although it leaves me wondering if we could remove `selectedRangeBounds` altogether.  It would allow people to enter crazy ranges in the text inputs (and these would get translated into crazy subsetting filters) and there may be other consequences. Best not to poke the bear maybe...

With regard to dates, I zoomed in to GEMS1 observation date and selected up to the end, and there's a tiny sliver of blue missing.  I thought it was going to be a whole day out, but it's just an hour, and I think I see why.  The `rightCoordinate` of the blue rectangle is `"2011-05-08T22:59:59.999Z"` (from react dev tools, component inspector).  Looks like a time zone issue (arrrghhhhhhh).  The `rightCoordinate` is calculated here using `DateMath.endOf('day')` - it's one microsecond to 11pm, instead of midnight.  This is odd because I'm in GMT/UTC at the moment.  Worth trying on your US-based browser. It would be great if you could investigate this - as a separate ticket + PR if you prefer.

image

I didn't see any problems with date range selections at other scales. Do you still see them?

bobular:
Before you merge, @jernestmyers - notice in the above screenshot there is no yellow axis truncation warning.

It turns out that all the maths done in `extendAxisRangeForTruncations` is at day resolution. When you are zoomed in to days, this ends up being rounded to zero days.

Please could you change it to hours throughout that function.  Do some quick tests and merge it.  It's never been reported (though I've seen it in the past and was too busy to report).  You can change the name of the PR to include the extra fix if you like.

dmfalke (approved):

jernestmyers:
Closed #385

PR #946, opened 2022-03-04
Add scatterplot to alpha diversity app

### LINKED ISSUE #304: Apps: alpha div app vizs
this needs a box and scatter plot.

asizemore:
This PR adds a scatterplot visualization to the alpha diversity app. When finished, this resolves #304 

The approach was the same as that for boxplot, with one small addition. When the yAxisVariable is undefined, the logic called it a date. Since our computed variable is on the y axis, we are always in this case. So there were a few additions to handle this scenario.

At this point i've learned that this will get messier when we add the next app :) The reason is that the computation is going to specify _two_ variables instead of one. Perhaps we could have the app tell the front end which variables it will override?

### DIFF
--- a/src/lib/core/components/computations/plugins/alphaDiv.tsx
+++ b/src/lib/core/components/computations/plugins/alphaDiv.tsx
@@ -4,12 +4,14 @@ import { useStudyEntities } from '../../../hooks/study';
 import { VariableDescriptor } from '../../../types/variable';
 import { findEntityAndVariable } from '../../../utils/study-metadata';
 import { boxplotVisualization } from '../../visualizations/implementations/BoxplotVisualization';
+import { scatterplotVisualization } from '../../visualizations/implementations/ScatterplotVisualization';
 import { ComputationConfigProps, ComputationPlugin } from '../Types';
 
 export const plugin: ComputationPlugin = {
   configurationComponent: AlphaDivConfiguration,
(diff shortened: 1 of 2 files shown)


dmfalke:
> At this point i've learned that this will get messier when we add the next app :) The reason is that the computation is going to specify _two_ variables instead of one. Perhaps we could have the app tell the front end which variables it will override?

Yeah, this is a big piece I've been trying to work out, although it's been several weeks since I thought about it.

One possibility is that all of our visualization components could optionally be told which variables to use for which axes. This moves the current visualization component logic of pulling variables off of the computation, to having the computation component pass in the variables to the visualization component.

The other piece of the puzzle is to have a utilize that takes the _incoming variables_ and the _input variables_, and returns a normalized set of variables used by the app. This would allow us to get rid of code like this: https://github.com/VEuPathDB/web-eda/blob/3bf646f4f3a5e1d21897dc50a163d426c239e30a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L743-L747

dmfalke (approved):

asizemore:
Thanks @dmfalke ! Since this branch is only getting merged in to `initial-apps-run`, do i need to do any versioning or additional organization-type things before merging?

dmfalke:
Nope! Merge freely!

Issue #948, opened 2022-03-08
Map visualization: add enlarged pie-plot on mouse-over functionality


bobular:
This will need to branch off (or follow merger of) #929.

All the components should be in web-components already - should be mostly wiring!

SheenaTomko:
Have to click the magnifying glass on the map, then can mouse over for enlarged pie plots with counts

PR #313, opened 2021-08-16
Disallow multivalued variables in visualizations

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/279

Note that the default is to not allow multivalued variables

PR #949, opened 2022-03-08
map improvement - add scale

moontrip:
This resolves [#313](https://github.com/VEuPathDB/web-components/issues/313)

Relies on [#315](https://github.com/VEuPathDB/web-components/pull/315) at web-components

This shows how to use showScale prop. Branched off from `add-pieplot-to-map`  - update: since `add-pieplot-to-map` was already merged, this will be merged into `main` branch


moontrip:
map mouse tool (mouseover popup) is added in this PR, which resolves #948. Details can be found at https://github.com/VEuPathDB/web-components/pull/315#issuecomment-1062522926

bobular:
> map mouse tool (mouseover popup) is added in this PR, which resolves #948. Details can be found at [VEuPathDB/web-components#315 (comment)](https://github.com/VEuPathDB/web-components/pull/315#issuecomment-1062522926)

Hi @moontrip - looks nice but I see that the pop-left/down/right when a marker is near the edge is not working. That is CSS-driven IIRC.  I'm in a call now, but we have a way to get CSS into the npm module...

bobular:
Update: seems that the pop-down behaviour does work but only when my browser window is small and the map is scrolled up to the top of the window.  Therefore it seems that the measurement of "distance from edge" is using the browser window edge rather than the container div edge?

bobular (approved):
Approving this `web-eda` PR though.  Conditional map scale works well and the threshold seems sensible to me!

bobular:
One issue I just discovered...

If there is no **Categorical overlay** variable, then the enlarged versions of the simple white donuts look a bit odd.

I'm not sure what the best UX solution is here though...  I don't think we want to add a thin black outline to the donut, do we? That would interfere with tiny slices. Though we could add a black outline if the number of segments == 1?

Or disable the mouse-over enlargement mode when no strat var is selected? (I don't like this very much)

moontrip:
> Update: seems that the pop-down behaviour does work but only when my browser window is small and the map is scrolled up to the top of the window. Therefore it seems that the measurement of "distance from edge" is using the browser window edge rather than the container div edge?

@bobular I think that @chowington can answer this question as he implemented such a nice functionality 😃 
 
> I thought I might be able to reproduce this in storybook, by putting the storybook controls/console panel on the right, but it still seems to work correctly. Anyway, at least it's not a CSS issue...

If assuming it depends on browser window then perhaps this is because the map at storybook is full-screen mode ? (100% width and height)

PR #957, opened 2022-03-09
update line viz selector icon

### LINKED ISSUE #884: Viz Icons: cement general line plot viz icon, knowing three more line plots are coming someday
Previously we had a viz icon we were planning to use for the line plot. However, since we're breaking up the line plot into multiple types, we'll need to be prepared to have a different icon for each type.

Planned types:
1. General line (b56)
2. Frequency line
3. Cumulative frequency
4. Kaplan Meier survival plots

See line plot mockups for examples of each.

asizemore:
Resolves #884 

New viz icon for general line plot

Screen Shot 2022-03-09 at 1 36 20 PM


bobular (approved):
Looks great. Sorry for the delay approving!

PR #959, opened 2022-03-10
filter apps by projectId

### LINKED ISSUE #955: Apps: filter apps list by projectId
Each app has a list of projects with which it is compatible. Have dataClient use the projectId to filter the list of all apps to only those that are compatible with that project.

asizemore:
Resolves #955 

In the data service, each app is assigned a list of projects for which it is valid. The code changes in this PR use that project array and filter the apps to show only those that are valid for the current project.

### DIFF
--- a/src/lib/core/types/visualization.ts
+++ b/src/lib/core/types/visualization.ts
@@ -111,6 +111,7 @@ export const ComputationAppOverview = intersection([
   Thing,
   partial({
     visualizations: array(VisualizationOverview),
+    projects: array(string),
   }),
 ]);
 
--- a/src/lib/workspace/ComputationRoute.tsx
+++ b/src/lib/workspace/ComputationRoute.tsx
@@ -9,6 +9,7 @@ import { createComputation } from '../core/components/computations/Utils';
 import { PromiseResult } from '../core/components/Promise';
 import { EntityCounts } from '../core/hooks/entityCounts';
 import { PromiseHookState, usePromise } from '../core/hooks/promise';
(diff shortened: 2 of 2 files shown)


dmfalke (commented):
This is good, but I have a small concern. It's possible that this code can run before `projectId` has loaded. When that happens, `apps` will not be filtered. The user might then see the unfiltered list for a moment. Once `projectId` is loaded, `apps` will be loaded again and filtered.

I suggest something like this:
```typescript
import { useNonNullableContext } from '@veupathdb/wdk-client/lib/Hooks/NonNullableContext';                                                                                                                                                                                     
import { WdkDependenciesContext } from '@veupathdb/wdk-client/lib/Hooks/WdkDependenciesEffect';                                                                                                                                                                                 

//...

  const { wdkService } = useNonNullableContext(WdkDependenciesContext);
  const promiseState = usePromise(
    useCallback(async() => {
      const { projectId } = await wdkService.getConfig();
      const { apps } = await dataClient.getApps()
      return apps.filter(app => app.projects.includes(projectId));
    }, [dataClient, wdkService])
  );
```

This moves all async tasks into the `usePromise` callback, which ensures that the resulting value is settled only when all tasks are complete. Let me know what you think.

asizemore:
Thanks @dmfalke !! This makes a lot of sense!

I made one modification just to check for an empty apps array before returning. There's no reason we should ever have an app that doesn't have any projects defined - that would mean we have a data service issue :) - so not sure if this check is worth losing the `const` on `apps`. Let me know!

dmfalke (approved):
This is great! I think the additional check for an empty apps array is a great idea!

Issue #838, opened 2022-02-08
Bug: functionality of starred view with featured variables


danicahelb:
If i have clicked the toggle to just see starred variables in the variable tree, I can still see variables in the featured variable box. The featured variables have the ability to be starred, but if I click to star them, they do not automatically show up in the view of starred variables in the variable tree. If I toggle the starred variable view off and then on, now the featured variables appear.

Please fix this behavior so that -- if I star a featured variable while in the starred variables view -- the newly starred featured variable immediately becomes visible in the starred view of the variable tree below.

this needs to be fixed on the variable tree in the tabs as well as the variable tree in the subsetting results modal

<img width="305" alt="image" src="https://user-images.githubusercontent.com/25450900/153087561-4d114ddb-bc42-4637-9a5e-922d46a425b0.png">

sweverschulman:
checked functionality over three different studies and can confirm that as soon as I click the star variable it shows up automatically in the variable tree when filtered for star vars in the tree.

PR #961, opened 2022-03-10
Fix bug with updating variable tree, featured vars only

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/838

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.7.12",
+  "version": "1.7.13",
   "dependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "@veupathdb/prettier-config": "^1.0.0",
     "@veupathdb/react-scripts": "^1.0.1",
     "@veupathdb/tsconfig": "^1.0.1",
-    "@veupathdb/wdk-client": "^0.5.5",
+    "@veupathdb/wdk-client": "^0.5.10",
```
jtlong3rd, line 79: You'll also want to update the `@veupathdb/wdk-client` `peerDependency` to `>=0.5.10`.

jtlong3rd (approved):
Works great! Just one inline comment from me.

Issue #954, opened 2022-03-08
Apps: add singleAppMode to switch between clinepi and mbio views


asizemore:
In `workspaceRouter`, add `singleAppMode` prop with value the name of the app to be used. For clinepi, name is 'pass'. If we have this set, the user only lives in one instance of the app always. Otherwise, user gets to choose from a list of apps (mbio view)

Important to get this out quickly so that we can start working on mbio apps on the main branch.

asizemore:
Closed with #962

SheenaTomko:
@asizemore, does this need qa or can we move to "Complete"?

asizemore:
@SheenaTomko this is one the clinepi team is best to QA, but QA is straightforward: if the QA site is showing the correct pass app and it looks just like it always did, then this ticket can be moved to Complete

SheenaTomko:
QA.clinepi looks like normal on the visualize tab. Moving to complete.

PR #962, opened 2022-03-10
Single app mode 954

asizemore:
Addresses #954

The goal of this PR is to offer an idea of how we might implement `singleAppMode`. This PR doesn't have to be the PR that actually adds this functionality.

# To Dos

- [x] redirect /pass/ to /pass-through/ for single app mode using the pass app
- [x] when in single app mode, dont save app instance until something has been actually done. Currently as soon as we _make_ an app instance it gets saved. We want to stay on /new/ for a while instead.


asizemore:
@dmfalke marking you as a review for sake of organization. If you like the direction this is going i can make it into a real PR, or definitely I'd welcome any thoughts about other strategies!

src/index.tsx
```diff
@@ -46,6 +46,10 @@ const dataServiceUrl = '/eda-data-service';
 const userServiceUrl = '/eda-user-service';
 const downloadServiceUrl = '/eda-user-service';
 
+// Set singleAppMode to the name of one app, if the eda should use one instance of one app only.
+// Otherwise, let singleAppMode remain undefined or set it to ''
+const singleAppMode = '';
```
dmfalke, line 51: I think we should allow this to be set in `.env`, just like `exampleAnalysesAuthor` below.

src/lib/core/components/EDAWorkspaceContainer.tsx
```diff
@@ -21,6 +21,7 @@ import { FieldWithMetadata, StudyMetadata } from '..';
 
 import { useFieldTree, useFlattenedFields } from './variableTrees/hooks';
 import { DownloadClient } from '../api/DownloadClient';
+import { useWdkService } from '@veupathdb/wdk-client/lib/Hooks/WdkServiceHook';
```
dmfalke, line 24: Unused import?
```suggestion
```

src/lib/core/components/computations/Utils.ts
```diff
@@ -7,11 +7,15 @@ export function createComputation(
...
+  computationId?: string
 ): Computation {
-  const computationId = createNewId(
-    new Set(computations.map((c) => c.computationId))
-  );
+  if (!computationId) {
+    computationId = createNewId(
+      new Set(computations.map((c) => c.computationId))
+    );
+  }
```
dmfalke, line 17: This can be abbreviated. See [logical nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment).

```suggestion
    computationId ??= createNewId(
      new Set(computations.map((c) => c.computationId))
    );
```

dmfalke (commented):
I think this approach is great! I left some inline comments.

asizemore:
Ready for review!!

My biggest struggle was preventing the singleAppMode from saving the analysis just because it started a computation instance. Curious if there's a more elegant way to do this!

Otherwise, my largest concern of course is messing up clinepi's website. So if you wouldn't mind double checking that all works as expected with the clinepi backend, that would be appreciated!

Looking forward to your feedback!

asizemore:
Note: please set
`REACT_APP_SINGLE_APP_MODE=pass`
in .env.local in order to use the pass app with the single app mode.

dmfalke:
> My biggest struggle was preventing the singleAppMode from saving the analysis just because it started a computation instance. Curious if there's a more elegant way to do this!

I was thinking about this exact thing last night, when I got the reviewer notification :slightly_smiling_face:. I will take a close look at your solution!

dmfalke (commented):
@asizemore this is great. I left some suggestions inline.

I think your approach to preventing a new analysis from being created is pretty good. I'm partly wondering if all setters should have an optional flag to prevent creating a new analysis.

Another approach I had thought about was passing in a computation object to [makeNewAnalysis](https://github.com/VEuPathDB/web-eda/blob/a1158003626a7fe17f12042c2f0acb86218a72ae/src/lib/core/types/analysis.ts#L133-L151). This would obviate the need for the new setter function.

What are your thoughts?

moontrip:
@asizemore I have tested this PR at both ClinEpiDB (dfalke-b) and Mbio (feature.mbio) configurations by changing `.env.local` 

- ClinEpiDB
When pressing Vizualize tab, yes it goes directly to our conventional interface, selecting New Viz interface, i.e., without showing Apps page (like Mbio case). Didn't test numerous cases but Vizs seem to work fine overall.

- Mbio
That being said, somehow I could not even get to the page that shows tabs (Subsetting, Viz, Notes, etc.) after selecting a study: it shows `Page not found` error instead. Tried to use my up-to-date branch, `initial-apps-run-update`, but it went well in there: so it may not a backend issue. I guess that Mbio is not considered yet or there may be some misconfiguration for Mbio case ? FYI, I did not set `REACT_APP_SINGLE_APP_MODE` in this Mbio test

asizemore:
> I'm partly wondering if all setters should have an optional flag to prevent creating a new analysis.

@dmfalke I think this is great and would prefer it to the current solution probably. I messed with having `createAnalysisOnChange` be optional and something that could be passed to the setter, but struggled with the implementation. So if we want to go this route i have a few questions :)

> passing in a computation object to [makeNewAnalysis](https://github.com/VEuPathDB/web-eda/blob/a1158003626a7fe17f12042c2f0acb86218a72ae/src/lib/core/types/analysis.ts#L133-L151).
Ahh interesting. I like this even more, because I think this solution would solve the new analysis problem more directly than updating the setters. I'll give it a go!

asizemore:
@moontrip thank you for your testing!! Hmm i wonder why the mbio didn't work... i'll take another look at it!

Did you add the permissions hack for mbio when testing? It's not in the branch so that i don't accidentally merge it, but it is still needed for a bit longer in order to access studies.

moontrip:
@asizemore FYI, I have checked both branches, this one and my initial-apps-run-update, but I did not use a hack (AnalysisPanel component) at both. 

In retrospect, it is a bit surprised that at some point I could use Mbio without the hack you did let me know before: idk from when because there were several merges from main.

asizemore:
ahhh yes :) i'm guessing it was a recent upgrade of the `study-access-package` in web-eda. That would explain why your `initial-apps-run-update` branch is working but this `singleAppMode-954` one is not 😄 

In retrospect i should have probably branched off of your updated branch!

dmfalke:
> I couldn't figure out how to get the computation without calling the whole `dataClient.getApps()` sequence of events, which led to duplicate code (we're doing the same thing in `analyses.ts` and `ComputationRoute.tsx`).

It looks like the only reason we need the overview in `createComputation` is to get the computation `name`: https://github.com/VEuPathDB/web-eda/blob/9fd57847e2ffc91caee16ccc63adab4d0d969a79/src/lib/core/components/computations/Utils.ts#L23. We already know the computation name from the value of `singleAppMode`, so I think we should change the signature of `createComputation` so that it just takes the computation `name`. I think this would remove the need to fetch the computation overviews in the analysis hook.

One downside to this is if the value of `singleAppMode` does not refer to a valid app, then we will still create this instance. I suppose that's okay since the analysis isn't saved yet, and we should probably throw an error when loading app overviews when this happens, if not already.

asizemore:
@dmfalke works like a charm!! Thank you for the help!

dmfalke (approved):

asizemore:
@moontrip just wanted to check in to see if you are happy with this PR as well? No rush, I just didn't want to merge if you're still thinking about it!

moontrip (approved):
@asizemore I think that I have extensively tested this enough before. Your additional changes are for improving codes per Dave so I think it is good 👍

PR #964, opened 2022-03-11
Lineplot: handle ordinal x variables

### LINKED ISSUE #963: Visualizations: Line plot front end does not handle ordinal x-vars
Ordinal x-vars, such as "Age group" `["0-11 months", "12-23 months", "24-59 months"]` currently break the front end.

If we support these (which I guess we will) it may make x-axis zooming controls more complicated (heads up to @moontrip) - unless we decide not to have zooming controls on x.

UPDATE: date variables on the y-axis were causing a 500 on the back end, but we decided not to calculate error bars for date variables, and so the PR for this ticket also disables error bars for date y-variables.

NOTE for testers: The only ordinal variable (there are very few) that seems to require reordering is "Age group" from SCORE Seasonal Transmission S. haematobium Cluster Randomized Trial

bobular:
Fixes #963 

image

Binning is completely disabled (even the binning on/off button) for these variables.

### DIFF
--- a/src/lib/core/api/DataClient/types.ts
+++ b/src/lib/core/api/DataClient/types.ts
@@ -323,20 +323,24 @@ export type LineplotResponse = TypeOf<typeof LineplotResponse>;
 export const LineplotResponse = type({
   lineplot: type({
     data: LineplotResponseData,
-    config: type({
-      completeCasesAllVars: number,
-      completeCasesAxesVars: number,
-      binSlider: BinWidthSlider,
-      binSpec: BinSpec,
-      xVariableDetails: type({
-        variableId: string,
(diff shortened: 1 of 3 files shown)


bobular:
Github's convert to draft button is broken, but it looks like more work is needed. The order of ordinal variables will need correcting (using the xVariable.vocabulary)

image

bobular:
There aren't many ordinal variables to test. Here's one that works!

India ICEMR Fever Surveillance 

image

bobular:
Also sorted a new/related issue with error bars breaking for date y-vars. 

Ready for review now (but no need to push out before the demos)

PR #968, opened 2022-03-14
Lineplot yscale

### LINKED ISSUE #965: Visualizations: lineplot - provide global y-axis scale functionality
This is needed for faceting mainly.  For now I will NOT ensure that the y-axis starts from zero (when all values are positive), but we can add this later after discussion with outreach/dataviz.

I get started on this now.

bobular:
Branched off `lineplot-ordinalx-963` #964  (so some of the diff is for that PR, sorry)
Fixes #965 

@d-callan - because all the code was basically there already (thanks @moontrip) I just enabled the same y-axis handling as we already have in scatterplot.  That means the range will be the annotated range, but if the data goes outside this (for example with error bars) then it is extended.  There's not explicit "zero y axis" behaviour as far as I can tell.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -859,7 +859,7 @@ function LineplotWithControls({
         </LabelledGroup>
         <LabelledGroup label="Y-axis">
           <Switch
-            label="Show error bars"
+            label="Show error bars (95% C.I.)"
             state={showErrorBars}
             onStateChange={onShowErrorBarsChange}
             disabled={neverShowErrorBars}
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -1195,7 +1195,19 @@ function processInputData(
   });
 
   const xValues = dataSetProcess.flatMap<string | number>((series) => series.x);
-  const yValues = dataSetProcess.flatMap<string | number>((series) => series.y);
+  // get all values of y (including error bars if present) in a kind of clunky way...
+  const yValues = dataSetProcess
+    .flatMap<string | number>((series) => series.y)
+    .concat(
+      dataSetProcess
+        .flatMap((series) => series.yErrorBarLower ?? [])
+        .filter((val): val is number | string => val != null)
+    )
+    .concat(
+      dataSetProcess
+        .flatMap((series) => series.yErrorBarUpper ?? [])
+        .filter((val): val is number | string => val != null)
+    );
 
   return {
     dataSetProcess: {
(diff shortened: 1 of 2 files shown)


d-callan:
I haven't looked at this in any detail but some questions:
1. Is there yet/ are we anticipating specific logic in line plot for the fact it takes both continuous and categorical y vars? I think what you describe is reasonable for the continuous y vars.
2. Will this scale all facets consistently regardless of error bars in individual panels? 
3. For the categorical y vars can we similarly borrow logic from barplot maybe?

moontrip:
@bobular glad to hear that it works with minimal changes 😉 I have looked at your changes and it looks fine to me. Will test it's functionality later.

Although you don't need to change here, just FYI, actually I have changed some logics for calculating dependent axis range (for axis range control). However, it would still be compatible with line plot as I could manipulate to merge main branch's line plot to axis range control with some (not significant) changes at line plot viz - does not mean that axis range control for line plot viz is available though. :)

moontrip:
@bobular I have tested lineplot for y-axis scale for faceted plot and it works fine. But it raises a couple of questions regarding UI. First, all other Vizs and histogram filter have plot controls of y-axis (left) and x-axis (right) in order, while this lineplot has x-axis and y-axis controls in order. Do you have a plan to switch the order ? Second, I noticed that y-axis aggregation option (mean and median) is shown on top of plot. I wonder if it was determined to be located in there. Perhaps it can be placed at y-axis control on the bottom of the plot, e.g., below Show error bars toggle ?

bobular:
> I haven't looked at this in any detail but some questions:
> 
>     1. Is there yet/ are we anticipating specific logic in line plot for the fact it takes both continuous and categorical y vars? I think what you describe is reasonable for the continuous y vars.

Sure - the y-axis range for the computed values from categorical variables will need to be different. Probably 0 to 1 but I haven't given it much thought yet.

>     2. Will this scale all facets consistently regardless of error bars in individual panels?

Yes

>     3. For the categorical y vars can we similarly borrow logic from barplot maybe?

Maybe. Barplot shows counts, and this will show proportions, so we'll see.

bobular:
> @bobular I have tested lineplot for y-axis scale for faceted plot and it works fine 👍 But it raises a couple of questions regarding UI. First, all other Vizs and histogram filter have plot controls of y-axis (left) and x-axis (right) in order, while this lineplot has x-axis and y-axis controls in order. Do you have a plan to switch the order ? 

When you do the axis controls, you are welcome to switch the order to whatever is most appropriate. I agree that the current lineplot layout is not consistent or logical!

> Second, I noticed that y-axis aggregation option (mean and median) is shown on top of plot. I wonder if it was determined to be located in there. Perhaps it can be placed at y-axis control on the bottom of the plot, e.g., below Show error bars toggle ?

Yes, I wondered that too, but when we add handling for categorical variables, that will be the third "aggregation option" so I think up top is OK for now.

PR #973, opened 2022-03-15
Add ranked abundance plugin 972

### LINKED ISSUE #972: Apps: Add Ranked abundance plugin
Add a ranked abundance plugin. Eventually this app will have a scatterplot and boxplot as defined in the [spreadsheet](https://docs.google.com/spreadsheets/d/1oO265vVOVR13MovZDQ9ubG9-nIJGruz54wSyrmCAfQI/edit#gid=0).

This PR is ONLY to add the plugin and configuration page. Nothing more at this time. Adding visualizations will be handled in #966

asizemore:
Resolves #972 

This PR adds the ranked abundance (called "abundance") plugin. Nothing really new was required - it's the same setup as alpha diversity, just with different names and methods.

Also renamed some other plugin files for consistency.


moontrip:
@asizemore thank you for your quick update on the component 👍 I will take a look at and test it.

moontrip (approved):
It works! 👍 I am happy to work with the abundance component as it would help me work my ticket better 😃

dmfalke (approved):
This looks good. I suppose we will want to factor out the layout being used for the configuration components at some point so that we can impose consistency.

asizemore:
Ahh yes factoring out the layout would be ideal. Adding that to my (ever-growing) list of organization tasks!

Issue #966, opened 2022-03-14
Ranked abundance app: update scatter plot and box plot visualizations


dmfalke:
Generally speaking, the changes needed for this issue are
1. Use variables provided by the computation configuration for the visualization configuration.
2. Use variables provided by the computation configuration for plot and facet labels.

For the time being, we can simply omit variables in `InuptVariables` that are provided by the computation. 

**The following visualizations will be used for the ranked abundance app**
- **BoxPlotVisualization**
    - **x-axis** is ranked collection var (*new from alphaDiv*)
    - **y-axis** is computed var
    - *Backend bug requires that an overlay var be included*
- **ScatterPlotVisualization**
    - **y-axis** is computed
    - **overlay** is ranked collection var (*new from alphaDiv*)

**In other words, the following changes need to be made for each visualization when used with the ranked abundance app**
- BoxPlot will use the collection variable as for it's x-axis. This means that `xAxisVariable` does not need to be specified as a part of the visualization configuration, and the collection variable should be used for the x-axis label.
- ScatterPlot will use the collection variable for overlay. This means that `overlayVariable` does not need to be specified as a part of the visualization configuration, and the collection variable should be used for the overlay label. It also means that the user will have to choose an x-axis variable before anything can be displayed (this shouldn't require any additional work, but it is worth mentioning).

_Note that collection variables are just categories in an EDA study's variable tree. This is useful for constructing the associated label(s) in the visualization._

moontrip:
@dave and @asizemore First of all, thank you for your detailed description made in this ticket @dmfalke. This is very helpful and resolves several questions I tried to ask 👍 I have tried to understand initial apps run and what we want to implement - studied some background knowledge about alpha div and ranked abundance too 😄  

Based on the current status and Ann's works and issue, a couple of questions arise regarding this issues and its scope of works. Please correct me if I misunderstand something. 

a) As far as I know, currently we do not have rankedAbu (ranked abundance) component, though we have alphaDiv component.

b) We should accommodate both alphaDiv and rankedAbu at each Viz. Given that the two types of plots have different choice of variables (X, Y, and Overlay) each other, we would need to distinguish them. With the absence of rankedAbu component, I can't check it contents of computation props. Based on alphaDiv component and its Vizs, I guess `computation.descriptor.type` is a unique name that can be used for this purpose. Certainly I may use known computation.descriptor.type for alphaDiv so it may be fine for the time being.

[...]

Issue #953, opened 2022-03-08
Apps: apps wrap vizs and send parameters


asizemore:
The app instance should tell the visualization which variables to expect the computation to provide. All other variables would be user-defined just like normal.

Examples:
alpha div box - compute defines y axis (alpha diversity). All other variables user-selected.
alpha div scatter - compute defines y axis (alpha diversity). All other variables user-selected.
ranked abundance box - compute defines x AND y axes. All other variables user-selected.

Consider having the data service return this information.

moontrip:
@asizemore Will change the codes to accommodate your suggestions at the `initial-apps-run-update` branch 👍

asizemore:
@dmfalke should i close this issue since it's part of the whole refactoring?

dmfalke:
I think that's fine. Thanks!

PR #974, opened 2022-03-15
update Initial apps run to the latest and box and scatter plot for ranked abundance

moontrip:
- Update (3/17/22): 
  - the latest `main` and `initial-apps-run` are merged into this branch
  - the issue #966 is handled in this branch because it was better to use the latest main - with the merge of axis range control, both box and scatter plot Vizs had lots of changes. Details about this b) is commented below separately

This PR is for updating `initial-apps-run` to be the latest with the `main` branch, so it would be merged into the branch, not main. Did various tests but could not find major issue. A minor one is `dependentValueType` prop, which the current implementation sets the initial format of dependent axis range control as date. Not an issue as a functionality but perhaps it is better to be empty layout (not date format) at the beginning. Will revisit this how to avoid that situation later.

Aside from above minor glitch(?), a couple of questions arose during this work: 
a) should we support axis range controls to alphaDiv and rankedAbun (ranked abundance) ? 
b) for scatter plot Viz concerning alphaDiv and rankedAbun, I guess we do not need to have plot option (Raw, Smoothed mean with raw, Best fit line with raw) ?


moontrip:
For @dmfalke and @asizemore information
Here are some details about #966 works, questions, and my thoughts. Some may not work properly as I could not confirm them with backend response, understandably at this stage. 

- defined type for `computation.descriptor.configuration` instead of `unknown`
  - certainly I could simply use it as inline type assertion like `as any` but tried to type it and changed several codes
  - though there are two places as I needed to use any, `ZeroConfiguration` component where it sets null for configuration
  - one concern is that we need to keep changing the type whenever adding new plot type (e.g., betadiv)
- format of X-axis label for abundance
  - currently set to `entity.displayName: variable.displayName` format. Is it alright ?
- I think that legend title for Scatter plot viz in abundance need to follow collection variable ?
  - currently set to the same format like above X-axis label - legend title supports ellipsis with tooltip so its length would not matter 
- need bird eye view in alphadiv and abundance ?
- need VariableCoverageTable in alphadiv and abundance ?
  - if so, it needs some treatment for computed/collection variables
    - e.g., X-, Y-axis, and/or overlay variable name, etc. depending on alphadiv and abundance
- need plot options for Scatter plot in alphadiv and abundance ?
  - not quite sure if alphadiv and abundance will support smoothed mean, best fit
  - I can set a condition to show/hide it
- need axis range controls in alphadiv and abundance ?
  - note that X-axis range is currently determined by variable's metadata, which may be a potential issue for boxplot vis in abundance ?

asizemore:
This is so exciting @moontrip !!! Tomorrow I'll hook this up to my local backend that contains a bunch of patches so i'll be able to give it a good test :) 

For now, a few answers to some questions.

- The X-Axis label for abundance boxplot can just be left blank or given the default 'X-axis' for now. I also really like what you have so no need to change it at all :) This axis title is a little tricky and i'm thinking about adding it to the compute response somehow... so the easy route in web-eda for now i think is good :) 
- Same for the Legend title. Whatever you choose for the boxplot x-axis label is great here as well!
- yes please to birds eye and VariableCoverageTable. However, if there is any difficulty with them, they definitely don't have a to be totally ready before this PR gets merged. They can be in a follow-up issue no problem! 
- both alpha div and ranked abundance should support trendlines like smoothed mean just like any other scatter :) We eventually will have an example of an app that doesn't allow trendlines, but we're starting with the simple apps for now!
- Axis range controls are like the birds eye view in my opinion - if they work out of the box no problem, then excellent. If not, we'll address in a later ticket. The computed variables are given a `displayRangeMin` and `displayRangeMax` in a property called `computedVariableMetadata`, so we're aiming to eventually have range controls. But like i said, if it doesn't work super easily right now, let's just add to a new ticket :)

moontrip:
@asizemore I appreciate your detailed information about many random questions from me 👍 Yes, some of the questions are lack of my env. to test the plot with backend response so maybe your tests reveal something I might miss or made mistakes 😄 The first thing I had in mind while working on this ticket (#966) is not to break pre-existing pass-through stuffs, so I had to test several cases after changing things for alphadiv and abundance 🤞  

It sounds great to hear that such a prop, `computedVariableMetadata`, would be provided in the near future 🥳 It would be quite helpful to fill out missing infos on those plot add-on stuffs in conjunction with main plot (the add-ons I mean are like bird eye, variable coverage table, axis range control, etc.). Actually I did not touch those add-ons in this PR - just left them as are. As you could glance at from my screenshots above, they are shown, but I could not check their functionality without backend response. Also great to know that alphadiv and/or abundance will still offer smoothed mean like stuffs. 

Honestly I doubt most add-ons will work out without changes, but with sufficient amount of backend response to test, we can probably make them happen for other types of apps beyond the pass-through.

Before merging this PR to `initial-apps-run` branch, I will try to keep updating this PR with the `main` and `initial-apps-run` branches 😃

asizemore:
> The first thing I had in mind while working on this ticket (https://github.com/VEuPathDB/web-eda/issues/966) is not to break pre-existing pass-through stuffs
Hahaha I totally agree!!! 

Well I can say you did an amazing job without a backend for testing...

Screen Shot 2022-03-17 at 6 33 46 PM

Screen Shot 2022-03-17 at 6 32 09 PM


I'll do more thorough testing tomorrow but it's looking really good so far!!! Is there anything in particular you'd like me to focus on? Without a backend for you to test, I'll keep feedback for this PR to things that can be worked on and tested without the backend. If I see other things that need addressing (for example, looks like we might have some legend fun ahead :D ) then i'll make new tickets and we can wait for a working backend to address those.

moontrip:
@asizemore Wow 😮 I didn't expect it works that well like that! Thank you for sharing the screenshots that encourage me alot. Yes I could see funny legend and probably it has different data format than pass-through ones. No variable names at the variable coverage table are expected as no xAxisVariable and/or yAxisVariable. 

Perhaps you can check some filtering, faceted plot, and consistency of y-axis range throughout multiple plots in faceted plot mode. One thing I wonder from your boxplot screenshot is whether the plot with the selection of the overlayVariable is correct - guessed it would be multiple plots, but it may differ from typical boxplot in the pass-through app.

But I can say that it is a great start to look at those nice plots 👍 Please feel free to make tickets whenever you find issues!

moontrip:
@asizemore TL;DR I made a commit to possibly resolve the legend issue of the scatter in the abundance app (like screenshot). So, can you please pull this PR and see if the funny legend issue can be resolved when you will be available ?

TL...
I was thinking over the reason of the funny legend at the scatter plot in the abundance app. It looked to me that legend list seemed to be made correctly when assuming that each legend item/text has very long but the visible texts are identical. So, I presumed that the issue might have been the marker color only. Scatterplot, which is arguably the most tricky viz among others, has a condition to check the presence of overlayVariable so I tried to set more general conditions considering both alphadiv and abundance apps: certainly need to check them with backend response later. So, it would be great if you can check both apps.

asizemore:
Thanks @moontrip this is great! Looks like your legend fix made a big difference!

Screen Shot 2022-03-18 at 10 12 27 AM

I've noted that the line of best fit for ranked abundance isn't showing the line colors in the legend (only when not faceted). Looks like they're being added to `legendLabel` which then means `legendLabelColor` is undefined for those. Anyways, a problem for a new ticket :) 

Anyways, faceting and filtering looks good! The axis ranges work, too! Across facets the y-axes are consistent when we're looking at raw data. 

Screen Shot 2022-03-18 at 10 26 45 AM

When we have a scatterplot with the smoothed mean, sometimes the y-axes vary because in one plot the confidence band goes above the expected y axis max. I'm guessing this is the expected behavior?

Screen Shot 2022-03-18 at 10 17 49 AM

Also here's a better boxplot from the abundance app - that overlay variable i used above only had 1 value so it made just the same boxplot haha.

Screen Shot 2022-03-18 at 10 36 53 AM

Looking good! I'll keep testing!

asizemore:
Checking mbio non-computational apps now (mbio has three non-computational apps that house all the vizs from clinepi's pass app). I noticed the scatterplot legend here is repeating colors with and without trend lines. The plot is colored correctly just the legend has this behavior.

Screen Shot 2022-03-18 at 10 40 30 AM

Otherwise all checks out!

asizemore:
Now hooked up to clinepi backend (dfalke). Pass-through app looking good!

moontrip:
@asizemore thank you for your detailed tests! 👍 👍 

glad to see that it is mostly working fine without specific changes. In general, I am off Friday from VEuPathDB, so I will look at issues you found next week if it can be fixed without backend. From your tests I could see three possible issues like legend line issue at scatter plot, faceted plot's y-axis range at a scatter plot (not sure if it is a bug or not at this stage), and non-computational apps. 

The last one, non-computational apps, is probably related to the condition I recently changed because I didn't expect that exceptions like them 😄 - the recent change at the scatterplot Viz was made purely based on the three apps such as pass-through, alphadiv, and abundance, and their unique parameter, `computation.descriptor.type`, is utilized to distinguish them. Honestly I don't know well about what non-computational apps are and what kinds of those apps are used but if it can be tested without backend (I mean to get into those apps to see Viz UIs at least) then perhaps I can find another unique conditions to house those exceptions - will see next week 🤞 

Just for a record, you might notice that I made a commit this morning. It was because I realized that typing for `computation.descriptor.configuration` did not work well with `pass-through` app for `null` case concerning `ZeroConfiguration`. I couldn't find a quick solution for that so just reverted it back to `unknown` for now. Perhaps will revisit this later too.

asizemore (approved):
Amazing work @moontrip !!

Next week sounds great! The non-computational apps don't need the backend fixes, so if you could take a look at those that'd be lovely! I'll make a new issue for them, though. Personally, i'd rather get this merged and then have lots of bugs to fix because it'll give us a better idea of how to fix them when it's all together. But that's just me!

On that note, @dmfalke based on some of these changes i wonder if it'd be helpful for the apps to also send axes titles and legend titles to the visualizations? Almost all the information needed is already spread throughout the computation response, and i can add the other things that are missing (mostly the abundance needs to tell us what to use as the x-axis or legend title).

Thanks again for all your work on this @moontrip and being a good sport without a backend to support you :)

moontrip:
@asizemore my pleasure! Thank you for your tests to confirm the behavior and identify issues 👍

dmfalke:
> On that note, @dmfalke based on some of these changes i wonder if it'd be helpful for the apps to also send axes titles and legend titles to the visualizations? Almost all the information needed is already spread throughout the computation response, and i can add the other things that are missing (mostly the abundance needs to tell us what to use as the x-axis or legend title).

If it's easy and possible for the backed to include labels, I think that would be great. Are you thinking about also doing this for "standard" variables, or just computed/collections?

asizemore:
@dmfalke just for the computed variables. We do have the ability to pass around display labels for any variable, but i think the front end is already gathering that info for regular variables. But yeah i'll look into where all of that info lives and get back to you!

Just to clarify, for now at least i'm envisioning that the compute response has something like `computedVariableDisplayName`. Then the `AbundanceScatterPlot` wrapper component would tell `Scatterplot` to put computedVariableDisplayName` on the y axis title. Is that in line with your vision for the apps/viz component relationship?

moontrip:
three kinds of additional changes are made since last changes. a) ts issue for `computation.descriptor.configuration` with respect to null case at `ZeroConfiguration` is now resolved, so removed `unknown`;  b) scatter plot issue for non-computational app reported by @asizemore, which I think it was occurred at the X-Y relationships app, is now fixed; c) fixed a small bug on R-table for Scatter plot (with best fit line with raw option)

I also tried to check other two issues mentioned at https://github.com/VEuPathDB/web-eda/pull/974#issuecomment-1072514212 but was not quite sure of the origin of the issues without backend response. Will revisit them later.

Some screenshots with feature.mbio and scatter plot using X-Y relationships showing correct legend coloring

XY-GEMS1-noOverlay

XY-GEMS1-overlay

asizemore:
Wow @moontrip this is awesome thanks so much for fixing those bugs!!

moontrip:
@asizemore I have revisited scatter plot viz codes and made some changes which may resolve `a best fit legend bug` when using `abundance app`: also added an exceptional case where single overlay data at the app. Certainly not 100% sure without test with data/response but when you will be available, can you please pull this branch at your local (for updating your dev) and see if the first case in your previous comment works ? https://github.com/VEuPathDB/web-eda/pull/974#issuecomment-1072473338 🤞

Here is the copy of your previous screenshot (from the linked comment) to point out the case I meant 😃 

159019780-b1d44a8a-367f-4837-8fc4-d44260aeb4e1

asizemore:
Screen Shot 2022-03-22 at 4 14 07 PM

Magic! Works great!!

moontrip:
@asizemore 🥳 Honestly, I didn't fully expect it works without further refinements! Thank you for your confirmation! 👍

moontrip:
#### updates
- display legend even for single overlayVariable data for app, which is related to https://github.com/VEuPathDB/EdaNewIssues/issues/202; 
- addressing axes and legend titles suggested at  https://github.com/VEuPathDB/web-eda/issues/953 for alphadiv and abundance except config.computedVariableMetadata.displayName that is not available yet. But placeholders (commented out though) for the displayName are made so it can be implemented easily in the near future.

asizemore:
@moontrip backend is working!! Just made a successful alpha div boxplot while hooked up to the mbio feature backend! 

Let me know if you run into any issues of course!

moontrip:
@asizemore Sounds great! 🥳 I will try and test plots too!

moontrip:
@asizemore here are initial observations from some tests for alphadiv, abundance, and X-Y relationships (non-computational) apps. Mostly I could see plots 🎉 so it is great to test some implementations I have made, and they worked fine for me.

some issues concerning computational apps such as alphadiv and abundance I noticed are: (X-Y relationships app seems to work fine for scatter plot)
a) y-axis range control: will talk about this in detail later
b) birds eye view: I don't know well about this so I can't judge whether it works fine or not. At least I could see the plot anyway
c) variable coverage table: I don't know well about this, but understandably collection/computed variables are not shown correctly in there because some variable info come from computation, not user-defined.

Speaking of a), the issue is that the truncation is not working correctly (yellow shading). Certainly, x-axis range control for scatter plot works fine, but y-axis range control does not show truncation (but axis range itself changes correctly). It is mainly because of no information about variable.type such as number, integer, date, etc. which are used for computing min/max. 

I think that for boxplot, y-axis data are always number or integer, right ? If that is the case, I can add some conditions to compute min/max for alphadiv and abundance. But for scatter plot, pass-through has variables with date in y-axis in some cases. I am not quite sure if it is the case for alphadiv and/or abundance. If those two computational apps always have variables with numeric then I can add conditions like boxplot above. If not, perhaps it is better to get a metadata info concerning that, e.g., config.computedVariableMetadata.displayType or config.computedVariableMetadata.type then I can probably use the info to compute min/max correctly.

I will test more and report here if I find something more.

P.S. Seems like config.computedVariableMetadata.displayName is not ready yet when I check backend response.

asizemore:
So glad it's working for you @moontrip !!

Let's see if i can get you some answers...

b) Birds eye view: seems to be working okay for alpha div but for abundance it looks funky. I'm guessing its on my end in plot.data. 

c) variable coverage table: I haven't looked into the code to understand exactly what's going on here, but i'm guessing there is some sort of check or attempt to pull out the numbers based on the input variables maybe? For alpha div we do send the information back, but the variableId is `alphaDiversity`, which i am guessing is just not being picked matched or something?
Screen Shot 2022-03-24 at 3 46 37 PM
In the abundance app the story is similar but also maybe more complicated because it's returning one row for each of those vars in the collection. Just by inspecting the response, it looks structurally okay, but for example when i make an abundance boxplot i might have a completeCases table with 7 rows. 

> It is mainly because of no information about variable.type such as number, integer, date, etc. which are used for computing min/max
Great point. A few things i can tell you are
- As far as I can see into the future, all computed variables (including alpha div and abundance values) are numeric. 
- I'm not at all opposed to adding new metadata about this, if the logic becomes annoying!
- If helpful for the y axis ranges, the metadata also includes `displayRangeMin` and max
Screen Shot 2022-03-24 at 3 32 50 PM

> P.S. Seems like config.computedVariableMetadata.displayName is not ready yet when I check backend response.
Hmm i'm seeing it come back (in the screenshot above), but it is within an array. Sometimes we'll have multiple computed variables coming back, which is why it's an array of strings instead of just one string. Are you able to see the array or are you getting no display name information at all?

moontrip:
@asizemore thank you for your detailed information 👍 As for displayName stuff, I rechecked the case and realized that it only exists for alphadiv, not abundance: is it right ? I might have checked abundance's backend response before.

Certainly I also saw displayRangeMin/Max in there so perhaps I can use them as one of parameters to compute min/max. It sounds great to hear that the alphadiv and abundance will have numeric for y-axis. 

So, for now I have at least two tasks to do as follows: 
a) considering displayName for y-axis title in the alphadiv
b) resolving truncation issue and considering displayRangeMin/Max too

Will probably do that early next week as I have something to do now and tomorrow is off from VEuPathDB.

asizemore:
> As for displayName stuff, I rechecked the case and realized that it only exists for alphadiv, not abundance: is it right ?

Ahh yes good catch! This is a funny case i'm not sure what to do with yet. Though on the front end the "abundance" variable is intended to be treated like a computed variable, from the perspective of the backend it was not truly computed which is why it doesn't have a displayName in the computedVariableMetadata. I'll have to take that one up with danielle. But all that aside, i think for now if we see a display name let's use it, and if not we can fall back to the logic you had previously - what do you think? Does that sound like a reasonable solution for now?

All the rest sounds great! Thanks again for all of your awesome work on this!!

moontrip:
@asizemore thank you for your detailed information. I have checked the use case of the display Name at #953 again and realized that it is only used for y-axis title at alphadiv as abundance always use Relative abundance title. So it would be fine for now :)

 One unknown is what array element we should use for the title if multiple elements exist as you mentioned the possibility. Just take the first one or concatenate all elements as a single title?

asizemore:
Great question! When we have multiple computed variables come back, we'll show multiple computed vars on the plot. 

The example we have right now is from beta diversity. It will return [computedVar1, computedVar2]. In this case, we'll end up sending computedVar1 to the x axis and computedVar2 to the y axis. 

Beta diversity is an app that we don't have to worry about for a few months though 😄 . Hopefully by then we'll have all the nice component organization where the app tells the viz which variables go on what axes! Should make it easy i'd hope!

moontrip:
Aha I see. I thought alphadiv may also have multiple strings but it is not the case 😃 then I can just use the first element for alphadiv's y-axis title. Thank you for your information 👍

moontrip:
@asizemore I have resolved bugs I reported earlier, displaying y-axis (dependent axis) truncation. Also applied #953 in this new commit, https://github.com/VEuPathDB/web-eda/pull/974/commits/fb57b17b7b8dd26ecc6dc743af8dc062cc953bc5. Note that for alphadiv's y-axis title, it needs to have backend response (`computedVariableMetadata`) so I currently set the default as `alphadiv` like before: it can be easily changed though. Please feel free to let me know if you find a bug/bug(s) or suggestion(s).

Here are some example screenshots for alphadiv and abundance.
- alphadiv scatterplot (for showing y-axis title and y-axis truncation)
alphadiv-scatter-yControl

- abundance boxplot (for showing y-axis truncation and single overlay data to display legend)
abundance-boxplot-yControl

asizemore:
Woooow thanks @moontrip this looks great!! I'll do some testing later this afternoon!

src/lib/core/api/DataClient/types.ts
```diff
@@ -237,22 +227,43 @@ export const ScatterplotResponseData = array(
...
+// typing computedVariableMetadata for computation apps such as alphadiv and abundance
+export type ComputedVariableMetadata = TypeOf<typeof ComputedVariableMetadata>;
+export const ComputedVariableMetadata = partial({
+  displayName: array(string),
+  displayRangeMin: string,
+  displayRangeMax: string,
+  // collectionVariable for abundance app
+  collectionVariable: partial({
+    collectionType: string,
+    collectionValuePlotRef: string, // e.g., yAxisVariable
```
asizemore, line 239: Just thinking for the future, the collectionValuePlotRef and collectionVariablePlotRef will give the information of where collection var data will go. In other words, it is a way for us to get information about what parts of the plot the app will override/supply data for.

moontrip: @asizemore yes I also guessed about that when I checked actual(!) backend response and the Ref values 👍: that's the reason why I could add e.g. texts in there haha 😄 That info may be helpful to fill out VariableCoverageTable's X, Y, overlay variables' labels, I suppose

asizemore:
Tested box and scatter in alpha div, abundance, and the non-compute Distributions and X-Y Relationships apps. The y-axis and truncation look good!

One question i have more generally, do we have a function like "findVariableDetails" or something like that? I ask because in the legend, for example, we see the raw variable ids, instead of their display names. This is understandable since the scatterplot returns the variable Id and entity Id for each entry in the legend. But im curious if we have a function already somewhere that takes the variable id and entity id and returns all that variable's information, including displayName and similar metadata? I looked around in the legend code here but maybe missed it?

moontrip:
> Tested box and scatter in alpha div, abundance, and the non-compute Distributions and X-Y Relationships apps. The y-axis and truncation look good!
> 
> One question i have more generally, do we have a function like "findVariableDetails" or something like that? I ask because in the legend, for example, we see the raw variable ids, instead of their display names. This is understandable since the scatterplot returns the variable Id and entity Id for each entry in the legend. But im curious if we have a function already somewhere that takes the variable id and entity id and returns all that variable's information, including displayName and similar metadata? I looked around in the legend code here but maybe missed it?

@asizemore for pass app, yes there is a function to find variable's metadata, [findEntityAndVariable](https://github.com/VEuPathDB/web-eda/blob/76bd9137c126f216c8b6e073d8210076b0c2b767/src/lib/core/utils/study-metadata.ts#L10), which accepts entities and variable descriptor where the latter contains entityId and variableId.

By the way I have one question concerning VariableCoverageTable. Honestly, I don't know well about how the values are computed, but I wonder if we need to show all variables' info regardless of a variable is selectable via variable tree (input variable selector). For example, boxplot at alphadiv only offers input variable selector for X variable, Overlay, and facet and Y variable is collection/computed variable. In this case, do we still need to show Y variable info at the table or hide it from the table?

asizemore:
> It finds entity and variable and the variable contains metadata, like variable.displayName etc

Wow perfect! Thanks so much!


> I wonder if we need to show all variables' info regardless of a variable is selectable via variable tree

Great question! Yes i think that would be ideal, as long as it doesn't cause too much of a problem. I'd guess we currently restrict to only showing those vars that are in the input var selector as a sanity check to prevent us from showing a bunch of random data? Simply showing whatever is in the response for the variable coverage table (`completeCasesTable`) would be a big win for the mbio apps, but it does mean we couldn't cross check these vars against much. What do you think?

Perhaps in the future after the reorganization the apps could tell the vizs if there are any vars that should be in the variable coverage table that also would not show up as an input variable?

moontrip:
> > It finds entity and variable and the variable contains metadata, like variable.displayName etc
> 
> Wow perfect! Thanks so much!
> 
> > I wonder if we need to show all variables' info regardless of a variable is selectable via variable tree
> 
> Great question! Yes i think that would be ideal, as long as it doesn't cause too much of a problem. I'd guess we currently restrict to only showing those vars that are in the input var selector as a sanity check to prevent us from showing a bunch of random data? Simply showing whatever is in the response for the variable coverage table (`completeCasesTable`) would be a big win for the mbio apps, but it does mean we couldn't cross check these vars against much. What do you think?
> 
> Perhaps in the future after the reorganization the apps could tell the vizs if there are any vars that should be in the variable coverage table that also would not show up as an input variable?

@asizemore I think that I make you misunderstand my question. Your response is much more than I wanted, which is great to know by the way 😉 And honestly again, I have no idea about variable table haha, so I don't know the answer, sorry.

My question was much simpler than that. As you know, current VariableCoverageTable in this branch only shows limited information. For example, alphadiv's boxplot where yAxisVariable is not selectable (computed from backend) is not shown in the table (no labe/text, no numbers). I wonder whether we hide the computed yAxisVariable from the table or we need to find a way to fill the yAxisVarible info.

moontrip:
@asizemore I will take a look at the VariableCoverageTable and see if I can make it: though I first need to understand the table better 😄

asizemore:
Thanks @moontrip !!

I'm curious @dmfalke and @moontrip if either of you have had thoughts on how we want to continue developing, especially over the next 1.5 weeks? 

I know I have a few tasks to work on, so I could branch from and merge into this `initial-apps-run-update` branch. However, that makes an extra-large PR to review and merge into main in a few weeks. Alternatively, we could try to wrap this PR up asap so that we can spread new work into more discrete PRs. Is one of these options preferred for either of you?

moontrip:
My two cents: Although this PR is currently supposed to be merged into its mother branch, `initial-apps-run`, but technically this can be directly merged into `main` too: we may need to let other developers know to use a parameter at env.local for ClinEpiDB, 
`REACT_APP_SINGLE_APP_MODE=pass` For that purpose, perhaps it is better to ask someone to test ClinEpiDB with this PR and see if it works fine in there: based on my many tests, they are fine but just in case that I missed something.

moontrip:
@asizemore here are what I have learned about VariableCoverageTable on the computation apps. Since I will have to do lineplot works, I won't do anything for the table for the time being but just for your information.

The table actually relies on the backend response, CompleteCasesTable, and manipulating it with each Viz's props on the VariableCoverageTable. If variable is undefined, like yAxisVariable at boxplot, then it did nothing in the corresponding component.

From my check, it appears to me that alphadiv's VariableCoverageTable can be implemented with some changes at both Viz and VariableCoverageTable component(s) (as it also refers to a hook) because the format of CompleteCasesTable response is quite similar to pass app. However, abundance returns different format, including all category (boxplot) or overlay (scatter plot) in the table: pass app always returns max. of four arrays such as X, Y, overlay, and facet. For example, below is a boxplot's CompleteCasesTable contents

- abundance boxplot: Host body habitat as overlayVariable (as it does not start automatically now)
```
[
    {
        "variableDetails": {
            "variableId": "EUPATH_0009253_Bacteria_Bacteroidota_Bacteroidia",
            "entityId": "EUPATH_0000808"
        },
        "completeCases": 1016
    },
    {
        "variableDetails": {
            "variableId": "EUPATH_0009253_Bacteria_Firmicutes_Bacilli",
            "entityId": "EUPATH_0000808"
        },
        "completeCases": 1016
    },
    {
        "variableDetails": {
            "variableId": "EUPATH_0009253_Bacteria_Firmicutes_Clostridia",
            "entityId": "EUPATH_0000808"
        },
        "completeCases": 1016
    },
    {
        "variableDetails": {
            "variableId": "EUPATH_0009253_Bacteria_Firmicutes_Negativicutes",
            "entityId": "EUPATH_0000808"
        },
        "completeCases": 1016
    },
    {
        "variableDetails": {
            "variableId": "EUPATH_0009253_Bacteria_Proteobacteria_Gammaproteobacteria",
            "entityId": "EUPATH_0000808"
        },
        "completeCases": 1016
    },
    {
        "variableDetails": {
            "variableId": "EUPATH_0009253_Bacteria_Actinobacteriota_Coriobacteriia",
            "entityId": "EUPATH_0000808"
        },
        "completeCases": 1016
    },
    {
        "variableDetails": {
            "variableId": "UBERON_0000466",
            "entityId": "EUPATH_0000096"
        },
        "completeCases": 1016
    }
]
```

As you can see, except the last array element that is user-defined overlayVariable, all others show similar variableIds: seemingly `variableId + name` format. Two question arise in this regard as one is for the table and the other is for the Vizs. For the table, IDK if we need to show all those info at the table and if so what the format is (both X and Y are computed variables)

For the Viz, in this case, the X-axis label for each category starts with the common text, `EUPATH_0009253_`,  as shown in the following screenshot (because backend response's boxplot.data.label has the same texts). Preferably that suffix needs to be removed in the plot but not quite sure if it should be done in the backend or frontend. Similar thing happens at legend of scatter plot like the 2nd screenshot that you may have already known. Certainly since we are using ellipsis with mouseover tooltip at both X-axis categorical label and legend label, it may be okay for the time being though 😄 

boxplot-abundance-Host body habitat

scatter-abundance-shannon

dmfalke:
> I'm curious @dmfalke and @moontrip if either of you have had thoughts on how we want to continue developing, especially over the next 1.5 weeks?
> 
> I know I have a few tasks to work on, so I could branch from and merge into this `initial-apps-run-update` branch. However, that makes an extra-large PR to review and merge into main in a few weeks. Alternatively, we could try to wrap this PR up asap so that we can spread new work into more discrete PRs. Is one of these options preferred for either of you?

I think we should get this all merged into `main` asap. The longer we wait, the more likely we will have to wait until after b57 is released to merge. Is there anything (feature-wise) preventing us from merging into `main`, at this point? This will have to be a major version bump, fwiw.

asizemore:
@dmfalke from my perspective there's nothing feature-wise that's preventing us from merging. I think the variable coverage table fixes can wait for another PR, if that's okay with you @moontrip ?

dmfalke:
Thanks @asizemore! To be clear, the rubric for merge-ability is that nothing is broken for ClinEpiDB users.

asizemore:
Also Re: variable coverage table, @moontrip thank you for all of your detailed information!! Your questions made me realize I missed scoping some parts of the variable coverage table with the apps and i need to get that sorted out! You have given me a lot to think about and digest :) Thank you!

asizemore:
@dmfalke that's perfect 🙂 

Outside of the three of us testing clinepi backend + single_app_mode = pass, are there any other checks we can do to ensure we haven't broken anything? Or would it be helpful to have more people as testers?

moontrip:
In a Viz-wise, basically most of my works were focused on box and scatter plot Vizs (and corresponding utils, types, etc.). I have been trying to make sure of not breaking ClinEpiDB Vizs while working on mbio, but there is a possibility that I might have missed something: this was the reason why I recommended another person (like @dmfalke if you did not test this PR for ClinEpiDB Vizs? 😄) to do some tests, just in case.

@asizemore sounds great! 👍 Glad my TL;DR comments were useful for you.

asizemore:
Ohh! I found one. Public analyses aren't working. I'm guessing this is a singleAppsMode issue that i missed. I'll look into it!

moontrip:
> Ohh! I found one. Public analyses aren't working. The error starts off like "Invalid value `undefined` supplied to 0 (type: { name: string, collectionVariable: { entityId: string, variableId: string } })"

@asizemore based on the type, it looks like it is related to this at vizualization.ts

```
export const ComputationConfiguration = intersection([
  type({
    name: string,
    collectionVariable: VariableDescriptor,
  }),
  partial({
    alphaDivMethod: string,
    rankingMethod: string,
  }),
]);
```

So, maybe previously made public analysis did not have it or perhaps `collectionVariable` should be placed at partial that allows undefined?

asizemore:
Ahh cool thanks @moontrip !

Okay so embarrassing but the 'alphadiv' issue in 5. was just because that analysis had used the alphadiv computation 🤦‍♀️ I thought for sure we haven't been working on this since Feb 1! But nope i was wrong haha.

I went _further_ back to November when i'm _sure_ we didn't have any computations 🙂 . Now im getting a similar error to that seen with the public analyses. Same text just only appears in the analysisState.error instead of printed on the page.

asizemore:
Okay i've got a funky observation. I was hoping to get those public analyses working and i noticed that the `analysisCache` is empty. If I log `analysisCache` just after [this line](https://github.com/VEuPathDB/web-eda/blob/26043fcc765ca391ed2603852927542eb5e8d7eb/src/lib/core/hooks/analysis.ts#L102), then I see it's empty {} when I try to load an analysis from December (gives the same error as the public analysis). However, if i switch back to an old branch `initial-apps-run`, I can see it is not empty.

Is it possible that we've introduced a change that alters the analysisCache? Or is that something that I shouldn't worry about because it's handled elsewhere?

The result is the analysisState.analysis is undefined.

asizemore:
Thanks for keeping this up to date with main @moontrip !

Discussed in the UX meeting on 04.08.22 - After this gets merged, we'll want to allow ourselves a large chunk of QA time so that the clinepi team can find any issues before the release on the 20th. Because of this fast approaching date, we decided to give ourselves a cutoff of **Tuesday night** for merging this branch before the release. If we're not ready to merge by then, we'll hold off until after the clinepi release.

moontrip:
Hi @asizemore Sounds great! I think that this PR needs to be merged into main branch, not its parent, initial-apps-run: I am not saying it should be done right now haha. It is currently a bit behind of main branch and there are several other PRs to be merge in to main, so I will update this PR early next week (probably Monday).

asizemore:
@moontrip yes you're right, this PR should go into `main` 👍 
That's the part (merging into main) that has the Tuesday cutoff 😄 

Sounds great thank you so much!

asizemore:
Got some feedback from Jamie:
"The “computation” should memoized with a useMemo ; also, the useMemo dependencies for defaultAnalysis should be changed to computation and studyId."

Addressed in the most recent commit! He also suggested cleaning up any “missing dependencies” eslint warnings that may be around. I'll work on that a bit, too.

Note this doesn't fix the analysisCache issue :(

moontrip:
@asizemore just be careful when addressing the eslint warnings for missing dependencies. Most of them are fine to be addressed but some may break something unexpectedly, especially if a missing dependency object has many keys/values embedded. One example case is vizConfig at visualization has many keys like vizConfig.xAxisVariable, .yAxisVariable, etc. In my experience, it is safer to separately add them instead of using vizConfig only because of various functionalities involved in a viz. 

So, whenever fixing a warning, please check if it works fine - usually fall to infinite loop if incorrectly assigned.

asizemore:
Thank you for the advice @moontrip !! Definitely good to know that the infinite loop is a likely way for fixing these warnings to break something. I appreciate it!

dmfalke:
@asizemore @moontrip Let me know if I can help with getting this merged, in any way.

asizemore:
@dmfalke and @moontrip sorry this took me sooo long - but thanks to a well-placed console.log (thank you Dave!) and some typescript advice (thank you DK!) we have a working solution to the issue i was having with old/public analyses.

@moontrip please feel free to add more detail, but from what i understand the old analyses' configurations don't fit in nicely to our `ComputationConfiguration`. Hence the switch back to `configuration: unknown`.

What do you both think of the solution? Also that was the last issue I saw with merging this PR. @moontrip and @dmfalke did you see any other behaviors that differed from the clinepi live site?

dmfalke:
`configuration: unknown` is correct. A way to avoid all of the casting to `any` is to use a [type guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), which `io-ts` is really nice for. You can see how they are used with the [BoxPlotVisualization configuration](https://github.com/VEuPathDB/web-eda/blob/87080ff5e4b9bf65d964eb42fc034cf95df54707/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx#L124-L134). We can do something similar with the computation configuration.

I will test out the branch more thoroughly tonight.

moontrip:
Yes as we discussed at Slack, such a revert somehow works. As I left a message, perhaps this kind of things will happen again and I think that it is eventually inevitable to be always compatible with previous public analysis ones. 

I have briefly tracked down some changes on types and a relationship with public analysis data. I could see at least one case that a public analysis data failed to load (made around Oct. 2021) and it is probably because some relevant types were changed after the data made. So perhaps we may need to determine how far back we should support pre-existing public analysis to be compatible with. This may also be complicated considering the number of users in the Live site and their publications to public analysis.

moontrip:
> `configuration: unknown` is correct. A way to avoid all of the casting to `any` is to use a [type guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), which `io-ts` is really nice for. You can see how they are used with the [BoxPlotVisualization configuration](https://github.com/VEuPathDB/web-eda/blob/87080ff5e4b9bf65d964eb42fc034cf95df54707/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx#L124-L134). We can do something similar with the computation configuration.
> 
> I will test out the branch more thoroughly tonight.

@dmfalke Actually I could type the configuration in detail without using unknown in there. It worked fine but one issue that the change (to be more strictly typed) is not compatible with pre-existing public analysis as Ann reported. So, only way is to revert back to unknown - I just made a comment regarding this.

asizemore:
Also - i don't think either of you were at the meeting last Friday so i know neither of you were able to give your input on the decision about Tuesday (tomorrow) being the last day to merge this PR. Do either of you have concerns about this plan? Is tomorrow still too close to the b57 deadline for comfort in your opinion? The goal cutoff for other b57 features is Friday, but since this one is so big I wanted to give it a few extra days.

Let me know what you think!

dmfalke:
> @dmfalke Actually I could type the configuration in detail without using unknown in there. It worked fine but one issue that the change (to be more strictly typed) is not compatible with pre-existing public analysis as Ann reported. So, only way is to revert back to unknown - I just made a comment regarding this.

@moontrip I don't think my comment was clear. What I'm suggesting is to keep the type as `unknown` in the `Analysis` type, but to use an `io-ts` decoder to see of the configuration conforms to the known type, similar to what we are doing with visualization configurations.

dmfalke:
> Also - i don't think either of you were at the meeting last Friday so i know neither of you were able to give your input on the decision about Tuesday (tomorrow) being the last day to merge this PR. Do either of you have concerns about this plan? Is tomorrow still too close to the b57 deadline for comfort in your opinion? The goal cutoff for other b57 features is Friday, but since this one is so big I wanted to give it a few extra days.
> 
> Let me know what you think!

I'm optimistic that we can deal with any potential bugs before release :slightly_smiling_face:

moontrip:
@asizemore I think that it depends on @dmfalke 's review/test tonight!

> > @dmfalke Actually I could type the configuration in detail without using unknown in there. It worked fine but one issue that the change (to be more strictly typed) is not compatible with pre-existing public analysis as Ann reported. So, only way is to revert back to unknown - I just made a comment regarding this.
> 
> @moontrip I don't think my comment was clear. What I'm suggesting is to keep the type as `unknown` in the `Analysis` type, but to use an `io-ts` decoder to see of the configuration conforms to the known type, similar to what we are doing with visualization configurations.

Thank you for your explanation @dmfalke 👍 That configuration: unknown is at visualization.ts, not analysis.ts, so perhaps you are saying that "Computation" is a kind of Analysis-related type 😃

dmfalke:
> Thank you for your explanation @dmfalke +1 That configuration: unknown is at visualization.ts, not analysis.ts, so perhaps you are saying that "Computation" is a kind of Analysis-related type smiley

That's true, @moontrip. I said `Analysis` because its descriptor contains a list of visualizations. Either way, the current approach is that those configurations are `unknown` at the `Analysis` level. It is up to individual `Visualization` and `Computation` (aka, `App`) components to "decode" those values into a known type, which is what we are currently doing with the visualization configuration. I'm happy to discuss more, if needed.

dmfalke:
FYI, I have tested the branch out with the latest ClinEpiDB backend services, and I didn't see any issues. In terms of not breaking ClinEpiDB, I think we are in good shape.

asizemore:
Wow thanks @dmfalke ! Just to clarify, is there anything else we should do before merging? I didn't quite follow all the typing discussion, but i can put those guards in place if that would be helpful?

moontrip:
There is at least one child PR, InputVariable work by Connor (https://github.com/VEuPathDB/web-eda/pull/1009, which will be merged into this PR. It is already approved so I will ask him about merging.

asizemore:
Woot sounds great! Thanks @moontrip and @chowington !!

Now that the last branch has been merged into this one, @moontrip and @dmfalke any other necessary changes we need to make before merging into main? (Please pardon my excitement 😄 )

moontrip:
I cannot think of others. Not quite sure how we will handle variable coverage table for computation app. Should we disable it for this release? (conditionally)

moontrip:
@asizemore thank you for your information! I can gladly take your offer 👍

dmfalke (approved):

dmfalke:
This should be a major version increase:
```
yarn version --major
```

PR #982, opened 2022-03-18
update study-data-access and move to dev and peer dependency

asizemore:
Related to [PR #8](https://github.com/VEuPathDB/web-study-data-access/pull/8) in web-study-data-access.

Upgrades the `study-data-access` package to the new version, and moves this package to the peer and dev dependencies (removed from regular dependencies).

**Testing**
Tested with dfalke clinepi backend. Tried making a visualization, subsetting, and downloading on a public study (all successful), same for protected study (downloading gave me a nice warning that i should request access, but other tabs successful), and on the private studies i saw only the study details page.

Recall plan is to patch update web-eda with this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "@veupathdb/components": "^0.11.27",
     "@veupathdb/coreui": "^0.13.0",
     "@veupathdb/http-utils": "^1.1.0",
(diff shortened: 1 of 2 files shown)


jtlong3rd (approved):
Works great!

PR #983, opened 2022-03-21
resolve r2 = null and change truncation message for axis range control

moontrip:
This resolves https://github.com/VEuPathDB/EdaNewIssues/issues/199 and https://github.com/VEuPathDB/EdaNewIssues/issues/198

A screenshot is attached to show a correct behavior as compared to the ones at each ticket above.

r2=null

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "1.8.2",
+  "version": "1.8.3",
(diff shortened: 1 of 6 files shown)


moontrip:
@asizemore @jtlong3rd it would be great if anyone can review this. It was simpler than I initially thought so a review would not take too much time I suppose 😄

Issue #958, opened 2022-03-09
Trivial: change order of visualization icons in new viz picker


bobular:
Not for next week's demo, but needs to be done for b57

The icons are currently in this order (below).  Geo map needs to go higher, before the faded out other icons.

![image](https://user-images.githubusercontent.com/308639/157548177-3bb77ee1-52d7-4dd5-ad92-7b4bb06d7481.png)

dmfalke:
The lodash function [orderBy](https://lodash.com/docs/4.17.15#orderBy) is well suited for this task.

SheenaTomko:
@aaronwlsong, when qc'ing, use UMSP or some of the SCORE studies like SCORE seasonal, which have geolocation data, as well as some studies like PRISM, which doesn't.

aaronwlsong:
Checked different type studies, it looks good now. Closing ticket.

Eg1. Study includes geolocation data

<img width="1996" alt="Screen Shot 2022-03-31 at 3 13 03 PM" src="https://user-images.githubusercontent.com/61077845/161132446-5616f21d-49d9-458a-8311-ee3f2f91eab1.png">

Eg2. Study does not include geolocation data

<img width="1931" alt="Screen Shot 2022-03-31 at 3 18 19 PM" src="https://user-images.githubusercontent.com/61077845/161132678-a08f7504-2864-4513-9bbc-a1c894e5e351.png">

PR #989, opened 2022-03-22
Sort viz icons to render visualizations in development last

jernestmyers:
Relates to issue #958

Imported `orderBy` from `lodash` to sort `visualizationsOverview` according to whether or not it is included in the `visualizationTypes` prop. Those `undefined`, like density plot, heatmap, and pie plot, will thus render behind the defined visualizations.

I tested with several studies wherein Geolocation Map is "Not applicable to this study" and with one study, USMP, wherein Geolocation Map is available. All tested studies rendered the viz icons in the desired order.


src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -242,7 +243,12 @@ function NewVisualizationPicker(props: Props) {
       </div>
       <h3>Select a visualization</h3>
       <Grid>
-        {visualizationsOverview.map((vizOverview, index) => {
+        {/* orderBy ensures that available visualizations render ahead of those in development */}
+        {orderBy(visualizationsOverview, [
+          function (viz) {
+            return viz.name ? visualizationTypes[viz.name] : undefined;
+          },
```
bobular, line 250:
```suggestion
          (viz) => viz.name ? visualizationTypes[viz.name] : undefined
```

This has fewer keywords and more hieroglyphics but I still think it's readable.

bobular (commented):
This is a great approach. I hadn't realised the on-screen icon order came from the back end `/apps` response order in `visualisationsOverview`.

In addition to the style suggestion comment I would suggest having the orderBy "iteratee" function return `true` and `false` for implemented and non-implemented respectively (and sort descending, because false comes before true). Currently it is `visualizationTypes[viz.name]` and `undefined` respectively.

Presumably you want to sort on "has implementation", and then retain the existing back end order where there are ties on "has implementation". This is currently what's happening, but under the hood I'm not exactly sure how.  lodash is currently sorting on VisualiationType objects and undefined values.

lodash may just be sorting on "truthiness" vs "undefined" but I think the code would be clearer if we made that explicit.

e.g. `orderBy(blah, viz.name != null && visualizationTypes[viz.name] != null, 'desc')`

(Seems that the array is not needed for sorting on a singe criterion in orderBy.)

dmfalke (approved):
This is great! Nice and simple.

dmfalke:
@bobular lodash performs a stable sort, so the backend order is preserved. I personally think the current approach is ok, but I'm also fine with your suggestion of the iteratee always returning the same type.

dmfalke:
> lodash may just be sorting on "truthiness" vs "undefined" but I think the code would be clearer if we made that explicit.
> 
> e.g. orderBy(blah, viz.name != null && visualizationTypes[viz.name] != null, 'desc')
> 
> 

I suppose it would be even clearer to return a `number`.

bobular:
> > lodash may just be sorting on "truthiness" vs "undefined" but I think the code would be clearer if we made that explicit.
> > e.g. orderBy(blah, viz.name != null && visualizationTypes[viz.name] != null, 'desc')
> 
> I suppose it would be even clearer to return a `number`.

yes, then the `'desc'` wouldn't be needed.

bobular:
> @bobular lodash performs a stable sort, so the backend order is preserved. I personally think the current approach is ok, but I'm also fine with your suggestion of the iteratee always returning the same type.

yes, the bit I wasn't clear about wasn't the stability of the underlying order, but the object vs. undefined sort behaviour.  Seemed that the order was 

`any object ; undefined`

and no sorting of objects themselves.

jernestmyers:
I agree @bobular that it was unclear exactly how lodash was ordering by truthiness vs undefined. How does the following look in response to your and @dmfalke's ideas/feedback?
```js
{orderBy(
    visualizationsOverview, [ (viz) => viz.name && visualizationTypes[viz.name] ? 1 : 0 ], ['desc']
    ).map( // remaining existing code
```
I could switch the `1` and `0` to get rid of the `'desc'` argument, but I feel like from a code readability perspective it makes more sense to return 1 for the truthy scenario. And Bob, you mentioned comparing against `null` specifically like `viz.name != null`, but I'm not sure why checking truthiness alone isn't sufficient. Am I missing something?

bobular:
That's nice - keep the 1,0 and 'desc' I think.

Much cleaner without the null checks!

bobular:
I used `!= null` to ensure true/false return values. You don't need to do this now with the `? 1 : 0`

bobular (approved):

PR #991, opened 2022-03-22
move "dependencies" to "{peer,dev}Dependencies"

dmfalke:
refs veupathdb/web-dev#11


package.json
```diff
@@ -2,33 +2,32 @@
...
+    "@material-ui/icons": "^4.11.2",
+    "@material-ui/lab": "^4.0.0-alpha.58",
+    "@veupathdb/components": "^0.11.27",
+    "@veupathdb/coreui": "^0.13.0",
+    "@veupathdb/http-utils": "^1.1.0",
     "@veupathdb/study-data-access": "^0.2.0",
-    "@veupathdb/wdk-client": ">=0.5.11"
+    "@veupathdb/wdk-client": ">=0.5.11",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
```
jtlong3rd, line 28: Perhaps we should allow React 16 as well?

jtlong3rd (approved):

package.json
dmfalke: Yes, that's a good idea. I will work on making the version ranges for peerDependencies more forgiving soon.

PR #999, opened 2022-03-25
set viewport for histogram viz with independent axis range control

### LINKED ISSUE #996: Visualisations: x-zoomed histogram is not including config.viewport.xMin/xMax in back end request
e.g. GEMS1 
histogram visualization
main var: Age
binWidth: 3 (default)
set the x-axis range [20,40]

You can see a tiny sliver of the (15,18] bar on the left hand side and its mouse-over is active.

Solution
`config.viewport.xMin` and `config.viewport.xMax` should be sent to the back end.

moontrip:
This resolves #996 

As suggested by Bob, I have set the viewport at config additionally, which the values follow the inputs of independent axis range. Here are screenshots before and after applying the viewport: same example is used at the issue

- before
histogram without viewport

- after  
histogram with viewport


bobular (approved):
Looks good, thanks.

Note that I have added similar functionality to your lineplot PR, @moontrip in [this commit](https://github.com/VEuPathDB/web-eda/pull/1007/commits/b3c698015c2248688883fefb8dec55b05349b06a)

Issue #988, opened 2022-03-04
Download modal slow for large studies, need indication that download is starting


SheenaTomko:
Tested on MAL-ED on the live site. Went to the downloads tab without subsetting and opened the modal for Participant Repeated Measures. Added all columns to the table, which took ~20 sec to show up. Hit "Download. Saw a tiny “waiting for clinepi” message in the bottom left corner, but no other indication that anything was happening for a few minutes. Is there some way to speed up the initiation of the downloads? Or show a more obvious message that a download file is being created?

dmfalke:
I labelled this as **low cost**, referring just to the loading indicator for the table data, and for reducing the redundant requests.

SheenaTomko:
Now the subset modal table loads in <5 sec and has loading indicators both when you open the modal and when you add columns. Downloading still takes a long time, >1 min to see the file start downloading, but we can address that in a separate ticket

dmfalke:
> Now the subset modal table loads in <5 sec and has loading indicators both when you open the modal and when you add columns. Downloading still takes a long time, >1 min to see the file start downloading, but we can address that in a separate ticket

See https://github.com/VEuPathDB/EdaSubsettingService/issues/58

PR #1001, opened 2022-03-25
Download modal performance issue

jernestmyers:
Relates to issue #988

This PR does the following:
1. Adds a Loading component that gives the user feedback that the modal's tabular data is being processed
2. Fixes duplicate fetch calls when `onModalOpen` is first opened with a new study

A new issue will be created to further examine ways in which to address the duplicate fetch calls when `handleSelectedVariablesChange` is called for the first time in a new study. Currently, the fetch call is made 3 times on its first click event. Subsequent click events only trigger one fetch call, as desired.


dmfalke:
This looks pretty good, but I think it might be a good idea to have the loading indicator when columns are added, as well.

jernestmyers:
@dmfalke Good point. I added a `LoadingOverlay` to the `DataGrid`. I also increased the `zIndex` of the variable selection panel to ensure the user isn't "blocked" by the loading indicator.

Since `setDataLoading` is called in `fetchPaginatedData`, we may get some flickering of the `LoadingOverlay` until we address the extra fetch calls.

jernestmyers:
@dmfalke Good point. I added a `LoadingOverlay` to the `DataGrid`. I also increased the `zIndex` of the variable selection panel to ensure the user isn't "blocked" by the loading indicator.

Since `setDataLoading` is called in `fetchPaginatedData`, we may get some flickering of the `LoadingOverlay` until we address the extra fetch calls.

dmfalke (approved):

Issue #926, opened 2022-02-25
styling: remove one of the scroll bars in viz tab variable selector


danicahelb:
The variable selector for the viz tab has 2 scroll bars and which sometimes makes it difficult to scroll down to variables at the bottom of the tree. please remove one of the scroll bars

<img width="344" alt="image" src="https://user-images.githubusercontent.com/25450900/155805837-5aaf1748-0b02-4bfd-ab2a-6a90bde3cc3c.png">

PR #1002, opened 2022-03-25
Adjust CSS to remove extra scrollbar in VariableList dropdown container

jernestmyers:
Addresses issue #926 to remove the extra scrollbar in viz tab variable selector container.

These CSS changes are supplemented by [a change in WDKClient](https://github.com/VEuPathDB/WDKClient/pull/194).

In summary, the CSS changes are as follows:
1. Remove `height: 60vh` from the `.EDAWorkspace-VariableList` and `.wdk-CheckboxTree` classes
2. Add `height: 100%` to the `.EDAWorkspace-VariableList` class
3. Add `flex-grow: 1` to the `.wdk-CheckboxTree` class (in [wdk-CheckboxTree.scss](https://github.com/VEuPathDB/WDKClient/blob/master/Client/src/Components/CheckboxTree/wdk-CheckboxTree.scss))


dmfalke:
I don't think this is quite ready. There are two issues that should be addressed.

First, I think we want just the checkbox tree items to be scrollable, not the entire container:
image

I was able to make this work with a simple addition to the css:
image
image

Second, in **Browse and Subset**, we want to put a max-height on the `.Variables` container. I think `70vh` is a good starting point.
image

dmfalke (approved):
I think this is good to go!

moontrip:
@jernestmyers I am curious if this works without updating wdk-client version at package.json? For peerDependencies it would be fine due to its condition, but what about devDependencies? Still 0.6.0 in there, not 0.6.6 in [this merged PR](https://github.com/VEuPathDB/WDKClient/pull/194)

jernestmyers:
ACK! I just reflexively fired off merges without thinking about the revamp to dependencies. Thanks for pointing this out.

moontrip:
haha no problem. As I said, I was just curious if there was a hidden magic to deal with it 😉

PR #976, opened 2022-03-16
Lineplot categorical WIP

bobular:

PR #1004, opened 2022-03-30
fixing bug for scatter plot viz - switch number to date variable etc

### LINKED ISSUE #1003: Visualizations: switching from number to date variable on scatterplot and (forthcoming) lineplot y-axis causes an exception
This may also affect other visualizations, but for sure scatter and line.

Reproduce with any scatter plot with two number variables on x and y. Then change the y variable to "observation date" (GEMS1, or whatever date variables are available in the study you are using).

The problem is that, when the user changes the `yAxisVariable`, the `defaultDependentRangeMargin` calculation uses the new value of `yAxisVariable.type` (which is now `"date"`) but the old values of `data.value.yMin` and `data.value.yMax` are still hanging around and contain numbers, not dates. This causes the error in `axisRangeMargin()`.

I can't see how to fix this right now, or I would have fixed it for line plot!

moontrip:
This resolves #1003 in part - only scatter plot viz is handled here. This may not be the best solution but it works without throwing error.

Note that a similar issue at the line plot viz will be addressed when working for line plot viz works: will branch out [lineplot-categorical](https://github.com/VEuPathDB/web-eda/tree/lineplot-categorical) and work for several other items Bob noted at #976 PR, e.g., axis range control etc.


bobular (approved):
Approving - but is there a solution where all the checks are made inside `axisRangeMargin`?
Then the lineplot/scatterplot solutions would not be so repetitive.

bobular (approved):

Issue #970, opened 2022-03-15
InputVariables: Handle provided inputs


dmfalke:
Some apps may provide inputs that are used by visualizations. `InputVariables` needs to be updated to handle this in a graceful and generic way.

So far, there are two types of provided inputs: **computed variables** and **collection variables**.  **Computed variables** are not a part of the usual study variables, so are less straight-forward. We will likely need to augment the type of `selectedVariables` to add support for them. **Collection variables** are just study category variables, so should be relatively easy to support.

I believe the following changes are sufficient to address this issue:
- [ ] Allow an `input` to be readonly.
- [ ] Allow a `selectedVariable` to be a **computed variable**. This includes passing along the display value of the computed variable.
- [ ]  Handle computed variables when finding disabled variables for each input.

SheenaTomko:
@asizemore, does this need qa or can we move to "Complete"?

asizemore:
No QA needed. Moving to Complete!

PR #1009, opened 2022-04-04
Input variables computation handling

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/970
Blocked by https://github.com/VEuPathDB/web-eda/pull/974 and https://github.com/VEuPathDB/web-eda/pull/824

### DIFF
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -18,6 +18,10 @@ export interface InputSpec {
   name: string;
   label: string;
   role: 'primary' | 'stratification';
+  /** Provide a string here to indicate that the input is readonly.
+   * The string will be displayed instead of a variable selector.
+   */
+  readonlyValue?: string;
 }
 
 export interface Props {
@@ -219,6 +223,11 @@ export function InputVariables(props: Props) {
     ]
   );
 
+  const spanStyles = {
+    height: '32px',
+    lineHeight: '32px',
+  };
+
   return (
     <div>
       <div className={classes.inputs}>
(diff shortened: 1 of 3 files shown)


moontrip:
@chowington thank you for your update and I will take a look at this tomorrow. By the way, I think that this PR should be merged into it's parent, initial-apps-run-update, not main branch? It is because the parent brach should be firstly merged with major updates before this PR.

chowington:
Thanks DK! I certainly was planning on waiting for the parent branched to be merge before merging this one---I thought that might happen soon, but maybe not! I changed the base accordingly.

moontrip:
@chowington I have tested this PR and overall it works fine. It is understandable that overlayVariable for scatter plot viz in the abundance app was not implemented due to the lack of the information we can get for now: it can probably be added later. I think that Dave needs to review this PR which means we need to wait for his comeback after vacation.

One minor comment is that the texts for computed variable (e.g., yAxisVariable for alphadiv app) seem to be smaller than others (I mean, other variable selector's title like X-axis and its selected variable, etc.).

chowington:
Thanks for the comments @moontrip!

Can you give me a photo of the font size difference you're seeing? I'm not seeing it on my screen. They all have the same font size to me:

image

moontrip:
@chowington oh! I thought Y texts were smaller than others, but was wrong probably due to: a) saw it at 4k screen without zooming; b) that gray pull-down menu at X-axis seemed to make me misread the size of Y-axis one. Sorry about that!

By the way, I am not quite sure whether `(fixed)` text is appropriate. I think it may be `(computed)` as well. Probably @asizemore can give a better idea of the word :)

inputselector

asizemore:
@moontrip @chowington I think "(fixed)" is great! I do like "(computed)" as @moontrip suggested, but since not all of our fixed variables are computed variables (sometimes they come from collections without any computations being run), i think "(fixed)" is preferable because it can be generally applied.

dmfalke (approved):
This looks good! We might want to adjust the styling in the future, but we should get general feedback before spending too much time on it.

moontrip (approved):

PR #1018, opened 2022-04-05
Variable coverage table for alphadiv app

moontrip:
This is an implementation of VariableCoverageTable for alphadiv app.

Made as a draft PR because this is not a generic way, even for alphadiv app. The idea of this is to utilize backend's response.(boxplot or scatterplot).config.yVariableDetails, which the yAxisVariable is computed one for alphadiv. The backend only returns xVariableDetails and yVariableDetails, so this approach would not work for the case where overlayVariable is a computed variable (e.g., scatter plot for abundance app). Perhaps better way to handle this is to get the info of computed variables at, e.g., computedVariableMetadata, response.(boxplot or scatterplot).config.computedVariableMetadata

Just leave this as a record for @asizemore 😃

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
@@ -75,6 +75,8 @@ type BoxplotData = { series: BoxplotSeries };
 // type of computedVariableMetadata for computation apps such as alphadiv and abundance
 type BoxplotComputedVariableMetadata = {
   computedVariableMetadata?: ComputedVariableMetadata;
+  // for alphadiv VariableCoverageTable
+  yVariableDetails?: VariableDescriptor;
 };
 
 // add type of computedVariableMetadata for computation apps such as alphadiv and abundance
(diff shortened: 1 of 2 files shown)


moontrip:
some screenshots to show VariableCoverageTable are attached in the following (used GEMS1):

boxplot-alphadiv-VCT

scatterplot-alphadiv-VCT

asizemore:
ahh this makes sense. Thanks @moontrip ! It's exciting to see the screenshots at least!

Do you think this will be easier to generalize when the app tells the viz what variable is on the y-axis, for example? 

> Perhaps better way to handle this is to get the info of computed variables at, e.g., computedVariableMetadata, response.

Yes that makes sense to me!It's still complicated by abundance using the computedVariableMetadata.collectionVariable in the overlay instead of just the regular computed variable, but maybe for now that's not too bad? What do you think?

moontrip:
@asizemore Thank you for sharing your thoughts! Below are my thoughts

>Do you think this will be easier to generalize when the app tells the viz what variable is on the y-axis, for example?

It may simplify something but it probably needs refactoring for better representation in that case. I remember that Dave has a plan to refactor Vizs so perhaps he can offer more details about this.

>Yes that makes sense to me!It's still complicated by abundance using the computedVariableMetadata.collectionVariable in the overlay instead of just the regular computed variable, but maybe for now that's not too bad? What do you think?

I don't think it must be computedVariableMetadata that was just an example location :) Unless a new response info causes a conflict with pre-existing backend response for ClinEpiDB, then anything would be fine, I suppose.

PR #1022, opened 2022-04-07
no data points at scatter plot viz

### LINKED ISSUE #1006: Scatter: Errs when no data points are returned
Screen Shot 2022-03-30 at 5 43 27 PM

Took this screenshot _right_ before the error covered the screen 🙂  Looks like the scatterplot response has `scatterplot.data` as an empty array. `sampleSizeTable` is also an empty array. 

Would the front end prefer an empty array be returned in this case? Or something else like null?

moontrip:
This would resolve #1006 

I have checked that the error occurs when calculating dependent axis range. If setting a condition to check no data, then it shows a message like the below screenshot owing to PluginError component. Thus, I think this fix would be good enough. What do you think? @asizemore 

scatter-PluginError

### DIFF
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -719,9 +719,6 @@ function ScatterplotViz(props: VisualizationProps) {
       };
   }, [xAxisVariable, defaultIndependentRange]);
 
-  //DKDK
-  console.log('legendItems =', legendItems);
-
   const plotNode = (
     <ScatterplotWithControls
       // data.value
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -719,6 +719,9 @@ function ScatterplotViz(props: VisualizationProps) {
       };
   }, [xAxisVariable, defaultIndependentRange]);
 
+  //DKDK
+  console.log('legendItems =', legendItems);
+
   const plotNode = (
     <ScatterplotWithControls
       // data.value
@@ -788,8 +791,10 @@ function ScatterplotViz(props: VisualizationProps) {
       checkedLegendItems={checkedLegendItems}
       legendTitle={variableDisplayWithUnit(overlayVariable)}
       onCheckedLegendItemsChange={onCheckedLegendItemsChange}
-      // add a condition to show legend even for single overlay data
-      showOverlayLegend={vizConfig.overlayVariable != null}
+      // add a condition to show legend even for single overlay data and check legendItems exist
+      showOverlayLegend={
+        vizConfig.overlayVariable != null && legendItems.length > 0
+      }
     />
   );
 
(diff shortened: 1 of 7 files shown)


src/lib/core/hooks/computeNumberDateDefaultDependentAxisRange.ts
```diff
@@ -22,6 +24,17 @@ export function useDefaultDependentAxisRange(
 ): NumberOrDateRange | undefined {
   // find max of stacked array, especially with overlayVariable
   const defaultDependentAxisRange = useMemo(() => {
+    // explicitly check empty data
+    if (
+      (!isFaceted(data?.value?.dataSetProcess) &&
+        data?.value?.dataSetProcess.series == null) ||
+      (isFaceted(data?.value?.dataSetProcess) &&
+        data?.value?.dataSetProcess?.facets != null &&
```
asizemore, line 32: For my education, what is the value of this check that the facets != null? Is it simply to ensure we're getting a good answer about the length === 0 check in the next line?

asizemore (commented):
Thanks @moontrip this looks great!

I noticed that the legend still appears but just has the title. 

Screen Shot 2022-04-07 at 11 17 51 AM

Do you think the legend should totally disappear? Or have the labels show up but all be grayed out like when there's no data for each? 

I think the legend appearing may be worth a second issue to prevent scope creep - does that sound good to you?

moontrip:
@asizemore Thank you for your review and comments. First of all, yes I also noticed the legend stuff that only shows legend title. It is because the legend items rely on data, so if no data then legend markers/labels are not shown: so we can't show legend markers/labels in this case. Perhaps I can hide the legend in that case not to be confused. What do you think? 

As for the facets != null condition, I don't know exactly why I added that condition: I thought it was due to type error but it was not. I might have seen an exception that does not return facets somewhere else, but can't remember. But I will remove that line for now as it works without that.

asizemore:
> Perhaps I can hide the legend in that case not to be confused. What do you think?

Sounds great to me! Is that easy enough to do with this PR or would you prefer I make a new issue?

This is great thanks @moontrip!

moontrip:
@asizemore I can add that hiding one in this PR sooner or later, then I will let you know 😃

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -719,6 +719,9 @@ function ScatterplotViz(props: VisualizationProps) {
       };
   }, [xAxisVariable, defaultIndependentRange]);
 
+  //DKDK
+  console.log('legendItems =', legendItems);
```
asizemore, line 723: Extra console.log?

src/lib/core/hooks/computeNumberDateDefaultDependentAxisRange.ts
```diff
@@ -20,8 +22,20 @@ export function useDefaultDependentAxisRange(
   updateVizConfig: (newConfig: Partial<ScatterplotConfig>) => void,
   yAxisVariable?: Variable
 ): NumberOrDateRange | undefined {
+  //DKDK
+  console.log('data =', data);
```
asizemore, line 26: Extra console.log?

asizemore (approved):
Looks lovely! Just had two tiny comments.

Thanks for fixing everything @moontrip !

moontrip:
@asizemore thank you for your review! 👍 Haha I forgot to remove those console.log - thank you for catching them. By the way, I will make the same approach to hide legend logic for all other Vizs, just for consistency: I don't think you would need to further review 😄

PR #1023, opened 2022-04-07
changing axis range control UI

### LINKED ISSUE #1011: Improve UI for axis controls on visualizations
The UX for x-axis and y-axis controls is confusing, it is not clear that X-axis and Y-axis are titles for these sections

Suggested improvements:
Underline the text for x-axis and y-axis
Put a line between the controls for x-axis and y-axis

Similar styling of the axis controls section should be implemented for all visualizations

moontrip:
This resolves #1011 

I am waiting for Danica's review and response for my screenshots, which is available at https://github.com/VEuPathDB/web-eda/issues/1011#issuecomment-1090643293. 

It is worth nothing that I cannot modify `LabelledGroup` component at web-components repo because it also affects other texts at other controls, e.g., Log scale, Range, Bin width at histogram viz, etc. Thus, individual viz is changed instead.

### DIFF
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -609,7 +609,7 @@ function HistogramPlotWithControls({
       />
 
       <div style={{ display: 'flex', flexDirection: 'row' }}>
-        <LabelledGroup label="Y-axis">
+        <LabelledGroup label="Y-axis controls">
           <Switch
             label="Log scale"
             state={uiState.dependentAxisLogScale}
@@ -654,8 +654,20 @@ function HistogramPlotWithControls({
             }}
           />
         </LabelledGroup>
-
-        <LabelledGroup label="X-axis">
+        {/* add vertical line in btw Y- and X- controls */}
+        <div
+          style={{
+            display: 'inline-flex',
+            borderLeft: '2px solid lightgray',
+            height: '14em',
+            position: 'relative',
+            marginLeft: '-1.2em',
+            top: '1.5em',
+          }}
+        >
+          {' '}
+        </div>
+        <LabelledGroup label="X-axis controls">
           <BinWidthControl
             binWidth={data?.binWidthSlider?.binWidth}
             binWidthStep={data?.binWidthSlider?.binWidthStep}
--- a/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/src/lib/core/components/filter/HistogramFilter.tsx
@@ -659,7 +659,7 @@ function HistogramPlotWithControls({
           style={{
             display: 'inline-flex',
             borderLeft: '2px solid lightgray',
-            height: '14em',
+            height: '13.6em',
             position: 'relative',
             marginLeft: '-1.2em',
             top: '1.5em',
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "2.1.4",
+  "version": "2.1.5",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
@@ -19,7 +19,7 @@
     "@material-ui/core": "^4.12.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
-    "@veupathdb/components": "^0.14.0",
+    "@veupathdb/components": "^0.14.3",
     "@veupathdb/coreui": "^0.16.0",
     "@veupathdb/http-utils": "^1.1.0",
     "@veupathdb/study-data-access": "^0.3.0",
@@ -78,7 +78,7 @@
     "@types/react-table": "^7.7.9",
     "@types/uuid": "^8.3.0",
     "@veupathdb/browserslist-config": "^1.0.0",
-    "@veupathdb/components": "^0.14.0",
+    "@veupathdb/components": "^0.14.3",
     "@veupathdb/coreui": "^0.16.0",
     "@veupathdb/eda": "^1.8.5",
     "@veupathdb/eslint-config": "^1.0.0",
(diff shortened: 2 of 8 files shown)


bobular (commented):
Hi @moontrip 

I don't see why we should remove the label feature from LabelledGroup and then re-engineer a new label with underline styling in every usage of LabelledGroup?

All the uses of LabelledGroup are in EDA as far as I know - so why don't we change the styling of LabelledGroup and then no changes are need in all the visualisations (and two uses in HistogramFilter) ?

PS. I don't like underlining personally - maybe others won't. We may have to change this back at some point - easier if we only have to do it in one place?

moontrip:
@bobular thank you for your comments. Certainly I could change LabelledGroup's style, like `textDecoration: 'underline'`, but as I mentioned, it affects all texts inside LabelledGroup component (see below screenshot as an example), because typically it is used to wrap all sub-controls. Possibly one way is to add a condition to check if the label is either 'X-axis' or 'Y-axis' at LabelledGroup's style but not quite sure if it is a good approach with possible exceptions.

histogram-control1

moontrip:
updated after discussing with Danica via slack: summary and a screenshot is found here, https://github.com/VEuPathDB/web-eda/issues/1011#issuecomment-1092083587

bobular:
As initial author of `LabelledGroup` I know it's not themeable/styleable and future-proofed like the newer CoreUI components but we could add a `labelStyles` prop, or just hardcode the underlining like this

```
      <div>
        {label && (
          <Typography
            variant="button"
            style={{ color: '#000000', fontWeight: 'bold', textDecoration: 'underline' }}
          >
            {label}
          </Typography>
        )}
        {children}
      </div>
```

moontrip:
@bobular thanks! 👍 I tried to add it to upper div, not inside typography which was the reason why I got all texts were underlined - I missed that part. By the way we will probably not use underline for X-/Y-axis label per discussions in the ticket, so I am reverting my work to the original. Also, there is a discussion whether we will change the text to be X-axis controls instead of X-axis in the ticket - it would be great if you can opine it too.

bobular:
Thanks for the heads up, I'll comment on the ticket.

I guess with hindsight one way to do it would be to add some customisation props to `LabelledGroup` (e.g. `labelStyles`).
Then in EDA create an `PlotControlsGroup` component that is a simple wrapper for `LabelledGroup` with appropriate styling, and use that everywhere. Makes moving `LabelledGroup` to CoreUI easier later on too.

moontrip:
changes are made like https://github.com/VEuPathDB/web-eda/issues/1011#issuecomment-1092855862. Also histogram filter is changed to be consistent with viz like below

histogramfilter-control

bobular:
I'm liking it.

I've noticed (and this is an issue on the `main` branch, not just here) that the "Bin width" label is wrapping in the HistogramVisualization but not HistogramFilter

I guess space is tight here.

I'd be happy with narrowing something globally so it fits without wrapping in both use cases. (Or maybe a `&nbsp;` will work too)

image

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -654,8 +654,20 @@ function HistogramPlotWithControls({
             }}
           />
         </LabelledGroup>
-
-        <LabelledGroup label="X-axis">
+        {/* add vertical line in btw Y- and X- controls */}
```
bobular, line 657: Maybe make the vertical separator a component?

moontrip:
@bobular yes I know that bin width are wrapped in the viz while it is not at histogram filter. As you guessed, it is because histogram filter has enough space to have (400px) while histogram viz is limited (350px) as I wanted not to go over the plot size (something like 720px or so). There seems no easy solution except allowing larger space for X-axis control: actually bin control also has another control for date variable (pull down menu to select month etc), which make things more complicated. To adjust size, each component/widgets like label (bin width text and it's padding/margin), pull down menu (month etc), bin slider and it's values' padding/margin/width but not quite sure if it is the way we have to do as it may affect other UI that uses those widgets.

src/lib/core/components/filter/HistogramFilter.tsx
moontrip: I think that perhaps it can be a technical debt or wait for viz refactoring that Dave mentioned before

bobular:
> @bobular yes I know that bin width are wrapped in the viz while it is not at histogram filter. As you guessed, it is because histogram filter has enough space to have (400px) while histogram viz is limited (350px) as I wanted not to go over the plot size (something like 720px or so). There seems no easy solution except allowing larger space for X-axis control: actually bin control also has another control for date variable (pull down menu to select month etc), which make things more complicated. To adjust size, each component/widgets like label (bin width text and it's padding/margin), pull down menu (month etc), bin slider and it's values' padding/margin/width but not quite sure if it is the way we have to do as it may affect other UI that uses those widgets.

Yes, @moontrip, it's tricky.

I have played around and I think `whiteSpace: 'nowrap'` on BinWidthSlider's label `Typography` and instead of the bizarre `max(...)` calculated width of Slider's `showTextInput` `NumberInput` (which I implemented* - my bad!) - a simple `maxWidth: 100` seems to work.  With dates, the right hand number label of the slider slightly overflows the parent div but it doesn't look too bad.

image


\* this was in the days of many decimal places of back-end suggested bin widths

moontrip:
@bobular Thank you for taking care of the changes in your `Slider` component/widget! 👍 I saw that max() part at the widget but I was hesitant to mess up your calculation in there without in-depth knowledge. I made changes following your suggestions at web-components, https://github.com/VEuPathDB/web-components/pull/337/commits/7a916be60f6abaa3e8d584fbf8fc4a7e41573c3a, and it seems to work great! Here is my screenshot on a lineplot case and now bin width is in the single line!

binWidth

bobular:
That's great @moontrip - did you push it? I can't see it on this page.

Thank you for treading lightly on my `max(...)` but it was trash 😉 

I will approve anyway 👍

bobular (approved):

moontrip:
@bobular yes I made changes at web-components, https://github.com/VEuPathDB/web-components/commit/7a916be60f6abaa3e8d584fbf8fc4a7e41573c3a, Thank you for your review! 👍

PR #1026, opened 2022-04-08
Fixed client to updated `map-markers` and `pieplot` API

bobular:
Just getting the maps working again. There is no issue/ticket for this, but this is essential groundwork to finishing off the maps for b57.

Only dfalke-b back-end seems to have the changes so far. Will need to coordinate roll out to qa back end along with this PR being deployed to qa.

### DIFF
--- a/src/lib/core/api/DataClient/types.ts
+++ b/src/lib/core/api/DataClient/types.ts
@@ -533,6 +533,16 @@ export interface MapMarkersRequestParams {
     geoAggregateVariable: VariableDescriptor;
     latitudeVariable: VariableDescriptor;
     longitudeVariable: VariableDescriptor;
+    viewport: {
+      latitude: {
+        xMin: number;
+        xMax: number;
+      };
+      longitude: {
+        left: number;
+        right: number;
+      };
+    };
   };
 }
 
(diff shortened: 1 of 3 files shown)


src/lib/core/api/DataClient/types.ts
```diff
@@ -557,16 +567,21 @@ export interface PieplotRequestParams {
   filters: Filter[];
   config: {
     outputEntityId: string;
-    showMissingness: 'TRUE' | 'FALSE';
+    showMissingness:
+      | 'TRUE'
+      | 'FALSE'
+      | 'noVariables'
+      | 'allVariables'
+      | 'strataVariables';
```
chowington, line 575: Interested in the meaning of true/false vs noVariables/allVariables. Do they mean different things?

d-callan: True/false are essentially deprecated values now, but in the interest of not breaking loads of things.... 
The new values are the other three, where the old false has the same behavior as the new 'noVariables' and the old true has the same behavior as 'strataVariables'

d-callan (approved):

chowington (approved):

src/lib/core/api/DataClient/types.ts
chowington: That makes sense! :)

bobular: You got it from two sources, must be true!

Issue #924, opened 2022-02-25
styling fixes for layout of birds eye view plot with overlayed vs faceted plots


danicahelb:
Noticed this in the LLINEUP study:

when plot is not stratified, birds eye shows on right side of plot:
<img width="1310" alt="image" src="https://user-images.githubusercontent.com/25450900/155808014-4ff6baae-75aa-47c5-ac2b-ba8f5b8a970f.png">

when an overlay is added, birds eye moves to below the plot (presumably because the birds eye view is less important than the featured plot):
<img width="1063" alt="image" src="https://user-images.githubusercontent.com/25450900/155807840-3ca5d99d-6e67-4ca6-834b-c9aa1fc9e8b2.png">

when a facet is added, birds eye now moves to the top of the plot!
<img width="1252" alt="image" src="https://user-images.githubusercontent.com/25450900/155807969-c1b9e81e-0e87-4f36-8786-fb82a42d4385.png">

we need to be consistent with the placement of the birds eye view so it doesn't jump around as the user builds their visualizations

bobular:
This is mainly an issue for small screens.  The moving of the supplementary plots between faceted and non-faceted is intentional and I think has already been through the committee.

There is a related issue with the new map viz which causes wrapping of the birds eye plot below the map, which @chowington and I have discussed (but I don't remember where). The map's main viz panel is wider than the other plot types, so it makes things wrap more readily.

When the birds eye goes below the plot or map, it doesn't look great. It needs some padding/margin above it.

jernestmyers:
@bobular Obviously, I don't know the history of these UX conversations. I played with this Friday afternoon and thought it made sense to fix the table, birds eye, and legend to be above the visualizations. I've attached pictures to demonstrate how it behaves when building a visualization. My screenshots are in Firefox on a 27" monitor.

1. New scatterplot
![image](https://user-images.githubusercontent.com/69446567/162744345-cea89f6d-96ea-440c-8782-ca11b9ecfd76.png)
2. Set axis variables
![image](https://user-images.githubusercontent.com/69446567/162744478-4f9a0d2e-476c-4776-9ff2-3fba0fd1d546.png)
3. Add an overlay
![image](https://user-images.githubusercontent.com/69446567/162744805-d2c1e767-44ee-412c-8419-f81ec548407b.png)
4. Add a facet
[...]

PR #1031, opened 2022-04-11
Adjust layout for consistency when building visualizations

jernestmyers:
Addresses #924

The layout takes into account these [notes from the 04/11/2022 UX meeting](https://github.com/VEuPathDB/web-eda/issues/924#issuecomment-1095605698). The build progression looks as follows on a `1280x900` resolution, simulated via Firefox's dev tools:

1. New visualization with no selections. Of note, the `VariableCoverageTable` is not yet rendered.
image
2. Select x and y variables
image
3. Select an overlay
image
4. Select a facet
image


src/lib/core/components/VariableCoverageTable.tsx
```diff
@@ -59,7 +59,10 @@ export function VariableCoverageTable({
       ? 'VariableCoverageTable'
       : `${containerClassName} VariableCoverageTable`;
 
-  return (
+  return variableSpecs[0].display ||
+    variableSpecs[1].display ||
+    variableSpecs[2].display ||
+    variableSpecs[3].display ? (
```
bobular, line 65: I was getting a runtime error about variableSpecs[3] being undefined.  One solution is below

```suggestion
  return variableSpecs[0]?.display ||
    variableSpecs[1]?.display ||
    variableSpecs[2]?.display ||
    variableSpecs[3]?.display ? (
```

Another might be something like `variableSpecs.some( (spec) => spec.display != null )` (if the test for not null is appropriate)

bobular (commented):
Just letting you have this comment for now. Will continue more testing

src/lib/core/components/VariableCoverageTable.tsx
bobular: I have continued testing with the `some(...)` option and that seems good.

Could you talk with @moontrip if he's available and talk about widening the legend to the standard 400px width that the variable coverage table and birds eye are using?

jernestmyers: I noticed with this approach that histograms are assigned `Main` as the `spec.display` and that bar plots are not. This resulted in the table rendering for histograms before selecting any variables.

I'm changing to `variableSpecs.some( (spec) => spec.variable)` to account for this.

dmfalke (commented):
The code looks good. I will look at the UI shortly.

dmfalke:
The UI looks pretty good, but I would suggest making the legend use the width available from its container, and remove the top/bottom margin from the right side.

**Before**
image

**After**
image

jernestmyers:
> Could you talk with @moontrip if he's available and talk about widening the legend to the standard 400px width that the variable coverage table and birds eye are using?

We were having this discussion earlier with @bobular and he spoke with @danicahelb about it. She suggested we not deal with the legend width until b58. Is the final verdict to wait until b58 or to sneak it in now?

bobular:
We could sneak it in if it's not likely to break anything.  The 400px wide legend with labels cut off less than half-way across will look sillier than the narrow legend IMO.  I'm just looking to see where it's done.  I don't think web-component's PlotlyPlot.tsx handles this any more.

dmfalke:
I think we should wait. It's a minor bother to me, but it sounds like there are a host of other related issues that might arise.

bobular:
I had a look at web-components' PlotLegend.tsx and I don't think we should rush this.
Ideally that component should have a width prop, from which the ellipsis threshold is calculated (I know this is not an exact science).  So let's punt on this now and save it for https://github.com/VEuPathDB/web-components/issues/344 in 58

bobular (approved):

dmfalke (approved):

dmfalke:
> I had a look at web-components' PlotLegend.tsx and I don't think we should rush this. Ideally that component should have a width prop, from which the ellipsis threshold is calculated (I know this is not an exact science). So let's punt on this now and save it for [VEuPathDB/web-components#344](https://github.com/VEuPathDB/web-components/issues/344) in 58

Now that this is being rendered as HTML, rather than SVG, we should be able to use the CSS `text-overflow: ellipsis` rule to automatically truncate based on available space, rather than truncating the text in JavaScript. This will allow things to be more responsive. I will update the issue linked above.

PR #1032, opened 2022-04-12
Scatter legend bug and message

### LINKED ISSUE #1016: Scatterplot: Give feedback when smoothed mean returns no points
To reproduce the error:

Filter: Cluster study arm: Handwashing

Screen Shot 2022-03-30 at 5 56 31 PM
Screen Shot 2022-03-30 at 5 56 41 PM

The scatterplot response is fine, but the smoothedMeanX, smoothedMeanY, etc are just empty arrays. 

Should we offer some sort of info/warning about why the user isn't seeing any smoothed mean lines, even after clicking the Smoothed mean with raw option?

moontrip:
This resolves #1016 

This commit addresses a scatter legend bug, as mentioned in https://github.com/VEuPathDB/web-eda/issues/1016#issuecomment-1095247051.


moontrip:
For one who is interested in:
The bug occurred a rare but plausible case, like the case in #1016. Scatter plot legend has the most complicated logics among other vizs due to its plot options to show raw, smoothed mean with raw, and best fit line with raw. Basically the fundamental logic is to find basic legend label and then make iterations to find corresponding marker color for raw, smoothed mean and confidence interval, and best fit line. 
The test case has the following base labels, e.g.:
label = [
    "Control",
    "Water",
    "Sanitation",
    "Handwashing",
    "Nutrition",
    "Water, sanitation, and handwashing",
    "Nutrition, water, sanitation, and handwashing"
]

When mapping color, iteration was made to find whether the data.name _included_ the base label using .includes(label). The problem in this case is somehow there were duplicate words, e.g., Water, Sanitation, Handwashing, Nutrition vs the last two labels, "Water, sanitation, and handwashing" and "Nutrition, water, sanitation, and handwashing". So, if using, e.g.,  .includes('Water'), then the logic detect two cases, "Water" and "Water, sanitation, and handwashing" so assigning two colors at "Water, sanitation, and handwashing" data which caused error. Same thing happened to "Nutrition, water, sanitation, and handwashing" data. The following screenshot shows before and after fix for the case of plot option of 'raw': the former shows gray like colors for the last two labels, while the latter (fixed version) shows correct colors

- before 
scatter-legend-bug1

- after fix
scatter-legend-fixed

When selecting 'Smoothed mean with raw' option, this threw errors because the option has three kinds of plots per label, e.g., raw label, raw label + smoothed mean, and raw label + confidence interval. Tried a couple of logics and finally found the simplest one to resolve: strictly checking string (using ===) rather than .includes(). 

Now it works for smoothed mean with raw case as well (certainly for other option, best fit with raw too)
scatter-legend-bug-fix-overlay

moontrip:
@asizemore This should use yarn link with web-component and use the PR, https://github.com/VEuPathDB/web-components/pull/341 for testing. 

I have made a **conditional message** to show only if plot option = smoothed mean with raw and the presence of no smoothed mean. Here is a screenshot: certainly the sentence can be changed so it would be great if you can suggest one. It is currently, "One or more smoothing for some data are not available: Possibly the sample is too small or too highly skewed. Relevant smoothed mean and confidence interval items in the legend are disabled and marked as lightly grayed checkboxes." Background color suggestion is also welcome :) The style of the box was taken from our error message box with red and I just changed the color to be greenish.

scatter-no-smooth-message

asizemore:
Thank you for the detailed explanation that was really helpful! This is on my To Do list for tomorrow!

asizemore:
Looks really awesome @moontrip !!

Also tested with the one point situation (seen often in mbio) and it works as expected 👍 

Screen Shot 2022-04-13 at 4 36 13 PM

Note that with one datapoint, the line of best fit has one point, while smoothed mean has none. Based on this new code the smoothed mean throws a banner while the best fit line does not. I think plot.data should handle not returning a best fit line for one data point, so i'll make an issue there.

All in all - code looks great!


Re: wording
How about "A smoothed mean could not be calculated for one or more data series. Likely the sample is too small or the data too highly skewed. <br />
Smoothed mean and confidence interval items for these traces have been disabled in the legend and marked with light gray checkboxes"? May be worth @d-callan or @danicahelb 's feedback as well.

Re: banner color. 
What i'm struggling with is if this message is a warning or info? I guess info. @jernestmyers remind me - what did you end up doing with the banner colors? Does the banner component take any color? Or is it set based on banner type or something?

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -1025,6 +1049,31 @@ function ScatterplotViz(props: VisualizationProps) {
             ) : undefined,
         ]}
       />
+
+      {/* show message if no smoothed mean exists */}
+      {!data.pending &&
+        vizConfig.valueSpecConfig === 'Smoothed mean with raw' &&
+        dataWithoutSmoothedMean != null &&
+        dataWithoutSmoothedMean?.length > 0 && (
+          <div
```
asizemore, line 1058: Can we use Jeremy's banner component from CoreUI here?

moontrip:
@asizemore Thank you for your detailed review! 👍 Certainly a suggestion of the sentence would be helpful for me to finalize me. I think it is just an info style?

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: I didn't carefully check the CoreUI's banner before so will see whether it fits here.

jernestmyers:
> What i'm struggling with is if this message is a warning or info? I guess info. @jernestmyers remind me - what did you end up doing with the banner colors? Does the banner component take any color? Or is it set based on banner type or something?

The color of the `Banner` component in `CoreUI` is indeed based on its `type` prop, much like it was before. Here are the lines that dictate the colors:
https://github.com/VEuPathDB/CoreUI/blob/d88c07e591c1e07a0c73c30af816e1aa455d1a0d/src/components/banners/Banner.tsx#L42-L59

moontrip:
> A smoothed mean could not be calculated for one or more data series. Likely the sample is too small or the data too highly skewed.
> 
> Smoothed mean and confidence interval items for these traces have been disabled in the legend and marked with light gray checkboxes

Hi @jernestmyers Thank you for your information! May I ask one quick question? I am looking at the use of Banner at VisualizationContainer component you did work before, and I wonder what the `SHOULD_SHOW_WARNING_KEY` means. Seems like empty string for that prop just work fine?

jernestmyers:
The `SHOULD_SHOW_WARNING_KEY` is specific to the `useLocalBackedState` hook, it's not affecting the `Banner` component. We wanted a user to be able to dismiss the performance warning banner, so we had to leverage `localStorage`.

Here's the type definitions for the `Banner` component:
```
export type BannerProps = {
  type: 'warning' | 'danger' | 'error' | 'success' | 'info' | 'normal';
  message: ReactNode;
  pinned?: boolean;
  intense?: boolean;
}
```

moontrip:
> The `SHOULD_SHOW_WARNING_KEY` is specific to the `useLocalBackedState` hook, it's not affecting the `Banner` component. We wanted a user to be able to dismiss the performance warning banner, so we had to leverage `localStorage`.
> 
> Here's the type definitions for the `Banner` component:
> 
> ```
> export type BannerProps = {
>   type: 'warning' | 'danger' | 'error' | 'success' | 'info' | 'normal';
>   message: ReactNode;
>   pinned?: boolean;
>   intense?: boolean;
> }
> ```

@jernestmyers Thank you for your information! 👍 Actually I just figured it out and was updating my question above 😄

moontrip:
@asizemore The Banner is quite easy to use so I could make it happen just now. As Jeremy showed the type, we can set 'info' or 'warning' which have different colors. Here are screenshots with your suggested sentence.

- info 
scatter-info-banner

- warning
scatter-warning-banner

asizemore:
Love it! Let's go with the info version 😄

asizemore:
Oof I missed a period at the end of that last sentence. Mind adding the period at the end of the "Smoothed mean and confidence interval items for these traces have been disabled in the legend and marked with light gray checkboxes" please? 

Otherwise looks amazing! Love the banner!

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -1025,6 +1051,23 @@ function ScatterplotViz(props: VisualizationProps) {
...
+      {/* show Banner message if no smoothed mean exists */}
+      {!data.pending &&
+        vizConfig.valueSpecConfig === 'Smoothed mean with raw' &&
+        dataWithoutSmoothedMean != null &&
+        dataWithoutSmoothedMean?.length > 0 && (
+          <Banner
+            banner={{
+              type: 'info',
+              message:
+                'A smoothed mean could not be calculated for one or more data series. Likely the sample is too small or the data too highly skewed. Smoothed mean and confidence interval items for these traces have been disabled in the legend and marked with light gray checkboxes',
```
asizemore, line 1064: + period at the end of this text please! My mistake sorry!

asizemore (approved):

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: no problem~ I just made the change 👍

Issue #1061, opened 2022-04-13
Viz: Error message in line plots when switching x-axis variable


danicahelb:
I got an error message with line plots using LLINEUP data. I had set up a line plot of "Meals per day categorization" vs "Household study timepoint", which looked good. Then I switched the x-axis variable to "Household data collection date" and got the error message below. I reloaded the page and it loaded just fine

<img width="1421" alt="image" src="https://user-images.githubusercontent.com/25450900/163241668-eca96f71-beb1-4d35-85e4-dcc5cf5cb22c.png">

danicahelb:
its fixed! thanks DK!

Issue #1050, opened 2022-04-12
Update Y-axis labels for line plots 


danicahelb:
**For mean and median modes, add "Mean: " or "Median: "**
For example, the label should look like "Mean: Age (years)"

**For proportion mode, label should just be "Proportion"**

Y-axis for line plots in proportion mode is confusing. Currently users will have to be careful about interpretation. Here I chose to plot the "No" proportion for Undernutrition, but the Y axis title is just Undernutrition, and the user has to be remember that in fact they plotted the proportion of kids that don't have undernutrition.

<img width="664" alt="image" src="https://user-images.githubusercontent.com/25450900/163027473-58cb245b-ad42-4de3-a91c-4af428773778.png">

It would be better if the Y-axis label just said "Proportion" for line plots in proportion mode.

In the future, we can offer users the ability to edit labels (see placeholder ticket VEuPathDB/web-eda#1053)

moontrip:
I will take a look at this

danicahelb:
looks great, thanks DK!

moontrip:
My pleasure 😃 Happy Easter!

PR #1062, opened 2022-04-14
conditional yLabel as Proportion, switching string to date variable bug

moontrip:
This addresses two issues: a) [Update Y-axis labels for line plots #1050](https://github.com/VEuPathDB/web-eda/issues/1050) and [Viz: Error message in line plots when switching x-axis variable #1061](https://github.com/VEuPathDB/web-eda/issues/1061).

- For a): the Y-axis label is conditionally set to be Proportion in the line plot viz when Proportion mode is used. A example screenshot is shown in the following (SCORE S. mansoni Cluster Randomized Trial)

- For b): this is similar situation to [switching from number to date variable on scatterplot and lineplot y-axis causes an exception #1003](https://github.com/VEuPathDB/web-eda/issues/1003). For the line plot viz, categorical and/or ordinal have string type and its value is string. For a date variable, its type is date but its value is also string: more exactly, date string. So when switching string type to date type, old values of sting from string type is used as a date string, which causes the error when computing axis range margin. To resolve this, a condition is set to check whether it is date type as well as date string value using Date.parse(). 

- screenshot of Y-axis label as Proportion
lineplot-yaxis-label

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.2.2",
+  "version": "3.2.3",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -592,7 +592,13 @@ function LineplotViz(props: VisualizationProps) {
                 : defaultIndependentRange.max,
           } as NumberOrDateRange)
         : defaultIndependentRange;
-    return axisRangeMargin(extendedIndependentRange, xAxisVariable?.type);
+    // need to check date type for the case switching from string to date variable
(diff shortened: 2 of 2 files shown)


bobular:
I haven't reviewed yet, but I am wondering if the continuous y-axis variables should have y axis labels: "Mean Age", "Median Age" ? @danicahelb

bobular:
Hi @moontrip - I don't see this bit implemented

```
For mean and median modes, add "Mean: " or "Median: "
For example, the label should look like "Mean: Age (years)"
```

moontrip:
@bobular Thank you for your suggestion. I will add them 👍

bobular (approved):
This is great - I tested and reproduced the error and then your branch fixes it.

Unfortunately I found another bug while testing (unrelated to this) that I now have to write up...

moontrip:
@bobular Yep as I replied at slack, I added the additional prefix for Mean and Median options. Here are screenshots. Are they alright for you?

lineplot-mean

lineplot-median

moontrip:
Bob liked and approved my changes at slack so I will merge this PR

Issue #994, opened 2022-03-22
Standardise birds eye plot mouse-over with (?) icon and it should not be active when plot is hidden


bobular:
Make a new barplot, but don't pick a variable. Hover over the area above the almost empty table and you see a small rectangle appear.

The Lineplot viz has an example of the (?) font awesome icon we want to use.

moontrip:
I have know that issue too. Perhaps the simplest solution is to only show the BirdsEyeView component's return with a condition (e.g., birdsEyeData != null)

bobular:
I have just added a new help tooltip in Lineplot

![image](https://user-images.githubusercontent.com/308639/160250934-6f5cb6e4-a21f-4c81-b4a5-b7578f41fd37.png)

as per the mockups [here](https://docs.google.com/presentation/d/16xCzGUs744KD3IQ91ITMt30pbAF9SJ7yvOkWZoE-AeE/edit#slide=id.g1097c687406_0_96)

I wonder if we should standardise and put a font-awesome `fa-question-circle` next to the birds-eye too (instead of the whole-mouse-over we have now) ?

bobular:
> I have know that issue too. Perhaps the simplest solution is to only show the BirdsEyeView component's return with a condition (e.g., birdsEyeData != null)

We don't want the variable data/no data table below to jump around.  But maybe that could disappear too.  What we want is nice behaviour for the first plot the user makes, and then on every tweak they make to it.  Perhaps the BirdsEye should stay on the screen during an update (like the main plot does).

The empty variable table looks a bit bad IMO.  Tidying up all this behaviour would be nice to have!

jernestmyers:
@bobular I'm curious where you'd place the icon in the `BirdsEyeView`. I played around a bit and reworked it as shown, but not sure how I feel about it.

![Image](https://user-images.githubusercontent.com/69446567/177401473-b6259e5b-184c-46d9-b9a6-f56401c03ccc.png)

danicahelb:
thanks, I've added to the agenda for 7/11/2022 EDA UX meeting

nkittur-uga:
Looks good.
<img width="1335" alt="Screen Shot 2022-08-01 at 12 07 07 PM" src="https://user-images.githubusercontent.com/67710160/182192032-e67f0246-3476-49b6-9b28-beae9d2ce1fc.png">

PR #1063, opened 2022-04-14
Condition the render of BirdsEyePlot to prevent phantom tooltip

jernestmyers:
Partly addresses #994 

@bobular @moontrip My solution to what I'm calling the "phantom tooltip" is, like DK suggested, to conditionally render the `BirdsEyePlot` based on `birdsEyeData`. To prevent any undesired jumping, the `falsy` condition renders an empty container with the same `height` and `weight` as the `containerStyles` props passed into the `BirdsEyePlot` component.

Nothing urgent here, but this was bothering me!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.2.4",
+  "version": "3.2.5",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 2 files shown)


bobular:
Hi Jeremy,
Thanks for taking a look. I'm going to approve this simply for the purpose of removing the phantom tooltip (great name, btw) for b57 but I don't that's the end of the story.

There's a related task here: https://github.com/VEuPathDB/web-eda/issues/994 scheduled for b58 to add the :grey_question: icon.

It would be cleaner to avoid the duplication of the empty div if possible (`visibility:` CSS?), or put the `minHeight` on an enclosing div.  I am not really the person to ask!

bobular (approved):

jernestmyers:
@bobular Sounds like a good plan for this to be a temporary fix until we revisit the entirety of issue #994 in `b58`.

PR #1067, opened 2022-04-14
Fix client-side decision on proportional vs mean/median

### LINKED ISSUE #1066: Lineplot visualization - erroneous logic to determine proportion vs mean/median mode
e.g. LLINEUP
x-axis: household coll. date
y-axis: Meals per day

This is being treated as a categorical variable.  There are no server-side constraints provided for yAxisVariable so it is up to the client to decide what to do.

The current logic sees the vocabulary for this variable and distinctValuesCount > 1 and decides it's suitable for proportion mode.
The back end doesn't like this and throws a 500.

I think I'll add a `dataShape !== 'continuous'` and see how we get on with that.

bobular:
Fixes #1066 

Please could someone merge this for me if we don't get this turned around tonight?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.2.3",
+  "version": "3.2.4",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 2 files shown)


d-callan (approved):

dmfalke (approved):
I'm approving this, but I would like to know why we don't have a constraint in place for this.

PR #1068, opened 2022-04-15
Add basic collection list

asizemore:
The goal of this PR is to provide basic functionality of collection variables - that is to be able to select them from the app configuration in the dropdown.

My current thinking is that it'd be great to have a basic implementation early on, so that @dpbisme can QA the app results and collection data while we implement a better solution (i'm guessing it'll involve the reorganization and everything!).

Leaving as a draft PR for a bit since this is not at all b57 related :) And i have a few more features i'd like to add before reviews.

### DIFF
--- a/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/src/lib/core/components/computations/plugins/abundance.tsx
@@ -37,10 +37,13 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
   const [rankingMethod, setRankingMethod] = useState(ABUNDANCE_METHODS[0]);
   const { computationAppOverview, addNewComputation } = props;
   const studyMetadata = useStudyMetadata();
+  console.log(studyMetadata);
   const entities = useStudyEntities(studyMetadata.rootEntity);
   let collections: Array<any> = [];
   function findCollections(entity: StudyEntity) {
+    console.log(entity.displayName);
     if (entity.collections?.length) {
+      console.log(entity.displayName);
       collections.push(
         entity.collections.map((collection) => {
           collection.entityId = entity.id;
@@ -54,17 +57,12 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
     }
   }
   findCollections(studyMetadata.rootEntity);
-  collections = collections[0];
-  console.log(collections);
+  console.log(collections.flat());
+  collections = collections.flat();
 
-  const ABUNDANCE_COLLECTION_VARIABLES = [
-    { entityId: 'EUPATH_0000808', variableId: 'EUPATH_0009253' },
-    { entityId: 'EUPATH_0000808', variableId: 'EUPATH_0009251' },
-  ];
   const [collectionVariable, setCollectionVariable] = useState(
     variableDescriptorToString(collections[0])
   );
-  console.log(collectionVariable);
 
   return (
     <div style={{ padding: '1em 0' }}>
--- a/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/src/lib/core/components/computations/plugins/abundance.tsx
@@ -6,6 +6,10 @@ import { findEntityAndVariable } from '../../../utils/study-metadata';
 import { boxplotVisualization } from '../../visualizations/implementations/BoxplotVisualization';
 import { scatterplotVisualization } from '../../visualizations/implementations/ScatterplotVisualization';
 import { ComputationConfigProps, ComputationPlugin } from '../Types';
+import {
+  StudyMetadata,
+  CollectionVariableTreeNode,
+} from '../../../types/study';
 
 export const plugin: ComputationPlugin = {
   configurationComponent: AbundanceConfiguration,
@@ -19,9 +23,7 @@ export const plugin: ComputationPlugin = {
 const ABUNDANCE_METHODS = ['median', 'q3', 'variance', 'max'];
 
 // Include known collection variables in this array.
-const ABUNDANCE_COLLECTION_VARIABLES = [
-  { entityId: 'EUPATH_0000808', variableId: 'EUPATH_0009253' },
-];
+// const studyMetadata = useStudyMetadata();
 
 function variableDescriptorToString(
   variableDescriptor: VariableDescriptor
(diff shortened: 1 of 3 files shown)


src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -18,26 +21,24 @@ export const plugin: ComputationPlugin = {
...
-  );
+  const [name, setName] = useState('New ranked abundance tool');
   const [rankingMethod, setRankingMethod] = useState(ABUNDANCE_METHODS[0]);
   const { computationAppOverview, addNewComputation } = props;
-  const study = useStudyMetadata();
-  const entities = useStudyEntities(study.rootEntity);
+  const studyMetadata = useStudyMetadata();
+  const entities = useStudyEntities(studyMetadata.rootEntity);
+  // Include known collection variables in this array.
+  const collections = findCollections(studyMetadata.rootEntity).flat();
```
jtlong3rd, line 37: Because `findCollections` could be computationally expensive, I would memoize this with a `useMemo`, and also put this memoization logic into a reusable [custom hook](https://reactjs.org/docs/hooks-custom.html):

```typescript
function useFlattenedCollectionVariables(entity: StudyEntity) {
  return useMemo(
    () => findCollections(entity).flat(),
    [entity]
  );
}
```

src/lib/core/utils/study-metadata.ts
```diff
@@ -31,3 +31,25 @@ export function makeEntityDisplayName(entity: StudyEntity, isPlural: boolean) {
...
+    );
+  }
+  if (entity.children?.length) {
+    entity.children.forEach((childEntity) =>
+      findCollections(childEntity, collections)
+    );
+  }
+
+  return collections;
+}
```
jtlong3rd, line 55: The overall logic looks sound - looks like you're doing a pre-order traversal of the study entities and "transforming" each  `collection` into a form the plugin needs.

Instead of "rolling your own" preorder traversal of the study entities, I would use our existing `preorder` utility from WDKClient to collect all the study entities into an array. (Example use [here](https://github.com/VEuPathDB/web-eda/blob/413b0ef2bcbe7e5fd15277670cb875570564c1f9/src/lib/mapveu/MapVeuAnalysis.tsx#L26-L28).) You could then use array methods like `map` / `flatMap` to transform each `collection` into whatever form the plugins expect.

jtlong3rd (commented):
This is almost good to merge - exciting stuff!

asizemore:
Thanks so much for the help @jtlong3rd ! The `preorder` function was really interesting to learn about and i appreciate the help with the custom hook!

Two other things i updated in the recent commits:
1. I messed with the output of `findCollections` a bit so that we could remove the `findEntityAndVariable` function later in the file, in hopes of avoiding duplicate work.
2. I realized line plot was disabled still in the XY Relationships app so i re-enabled it. This is just lazy on my part for not making a new PR so i apologize it was just so few lines to change and so tempting 😄 !!

asizemore:
Courtesy tag: @dmfalke you may be interested in this PR

src/lib/core/utils/study-metadata.ts
```diff
@@ -31,3 +32,23 @@ export function makeEntityDisplayName(entity: StudyEntity, isPlural: boolean) {
...
+    preorder(entity, (e) => e.children ?? [])
+  ).flatMap((e) => {
+    const collectionWithEntity = e.collections?.map((collection) => {
+      return {
+        ...collection,
+        entityId: e.id,
+        entityDisplayName: e.displayName,
+      };
+    });
+    return collectionWithEntity;
```
dmfalke, line 50: You should avoid returning `undefined` here. Otherwise, you will end up with an array that has `undefined` elements. Returning an empty array ensures that the resulting array (`collections`) does not have `undefined` elements. This will allow you to remove some defensive logic where the return value of this function is used.
```suggestion
    return collectionWithEntity ?? [];
```

src/lib/core/hooks/study.ts
```diff
@@ -224,6 +227,10 @@ export function useFindEntityAndVariable(entities: StudyEntity[]) {
   );
 }
 
+export function useFlattenedCollectionVariables(entity: StudyEntity) {
```
dmfalke, line 230: Can we rename this? I think "Flattened" is a little confusing.
```suggestion
export function useCollectionVariables(entity: StudyEntity) {
```

src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -18,26 +17,28 @@ export const plugin: ComputationPlugin = {
...
-  );
+  const [name, setName] = useState('New ranked abundance tool');
   const [rankingMethod, setRankingMethod] = useState(ABUNDANCE_METHODS[0]);
   const { computationAppOverview, addNewComputation } = props;
-  const study = useStudyMetadata();
-  const entities = useStudyEntities(study.rootEntity);
+  const studyMetadata = useStudyMetadata();
+  // Include known collection variables in this array.
+  const collections = useFlattenedCollectionVariables(studyMetadata.rootEntity);
+  if (collections == null || !collections[0])
```
dmfalke, line 33: I think this can be changed to the following, once the return-value of the callback passed to `flatMap` in `findCollectionVariables` is updated as I suggested.
```suggestion
  if (collections.length === 0)
```

src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -63,12 +64,16 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
...
+              collectionVar && (
+                <option
+                  value={variableDescriptorToString({
+                    variableId: collectionVar.id,
+                    entityId: collectionVar.entityId,
+                  })}
+                >
+                  {collectionVar.entityDisplayName}: {collectionVar.displayName}
                 </option>
               )
```
dmfalke, line 78: I think this can be changed to the following, once the return-value of the callback passed to `flatMap` in `findCollectionVariables` is updated as I suggested.
```suggestion
                <option
                  value={variableDescriptorToString({
                    variableId: collectionVar.id,
                    entityId: collectionVar.entityId,
                  })}
                >
                  {collectionVar.entityDisplayName}: {collectionVar.displayName}
                </option>
```

dmfalke (commented):
Thanks for tagging me, @asizemore. This is great. I added some suggestions. My comments in `abundance` also apply to `alphaDiv`. Let me know what you think. Thanks!

jtlong3rd (approved):

dmfalke (approved):

asizemore:
Thank you both so much for your help! 

@dmfalke i can see now why avoiding arrays with undefined elements can make the downstream logic much cleaner in this case. Seems like this is generalizable, too? So that more often than not when making arrays, i should prefer empty over undefined elements?

dmfalke:
> Thank you both so much for your help!
> 
> @dmfalke i can see now why avoiding arrays with undefined elements can make the downstream logic much cleaner in this case. Seems like this is generalizable, too? So that more often than not when making arrays, i should prefer empty over undefined elements?

Yes, anytime you can make an array contain only non-`undefined` values, you're going to have a much easier time!

PR #1070, opened 2022-04-15
Remove nulls-hack, since we use a custom legend, and so data displays correctly

### LINKED ISSUE #1069: Line plot is empty if overlay variable only has data for one value
If I add a filter to only include data where lost earnings due to care = yes, then the plot is empty. I would expect the plot to still have the same blue line

image

_Originally posted by @danicahelb in VEuPathDB/plot.data#157

dmfalke:
Fixes #1069

The previous hack to fill x and y series data with `null` values appears to break plots with only one non-empty data series. This PR removes that hack, since we now use a custom legend.

cc @moontrip @bobular

ETA Note that I'm not completely satisfied with this solution, as it seems to rely on Plotly.js's behavior to treat such elements as "empty". I think a better solution would be to remove the entry from the data array, but this requires some reworking of how [legendItems is constructed](https://github.com/VEuPathDB/web-eda/blob/9cbf9b2656206e32fbfe64557732e9fab45d599d/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx#L633-L676). I will probably make an issue for this as a follow-up, since I don't want to mess with the current logic right now :slightly_smiling_face:.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.2.5",
+  "version": "3.2.6",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -1819,8 +1819,8 @@ function processInputData(
           : (el.seriesY as string[]);
 
       dataSetProcess.push({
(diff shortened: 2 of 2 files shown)


jtlong3rd (approved):

dmfalke:
https://github.com/VEuPathDB/web-eda/issues/1071

bobular:
Yeah - I think `legendItems` could always be constructed from the variable's vocabulary. At least to determine which legend items to show. The `hasData` checks will need to be made on the data still.

bobular:
Haha, you said exactly the same in the new ticket.

PR #1083, opened 2022-04-19
Ranked abundance app uses display names

### LINKED ISSUE #1084: Apps: Abundance vizs use display names and respect ordering
From mbio feature site, GEMS, ranked abundance app.

The ranked abundance app returns a list of variable ids as labels (see box plot example below, also see scatter -> overlay). We want to 
1. Replace these var ids with their display labels
2. Have the labels show in order, based on the order defined in `computedVariableMetadata`.

Screen Shot 2022-04-19 at 3 36 15 PM

asizemore:
Resolves #1084 

The ranked abundance app response has box labels (or scatter overlay labels) that are variable ids instead of display names. Additionally the computation metadata lists these variable ids in a specific order which is not yet respected by the box or scatter.

This PR replaces the var ids with display names, and then also ensures they are displayed in the correct order.

To Dos
- [x] Box
- [x] Scatter

### DIFF
--- a/src/lib/core/utils/visualization.ts
+++ b/src/lib/core/utils/visualization.ts
@@ -12,6 +12,7 @@ import { CompleteCasesTable } from '../api/DataClient';
 import { Bounds } from '@veupathdb/components/lib/map/Types';
 import { Filter } from '../types/filter';
 import { VariableDescriptor } from '../types/variable';
+import { findEntityAndVariable } from './study-metadata';
 
 // was: BarplotData | HistogramData | { series: BoxplotData };
 type SeriesWithStatistics<T> = T & CoverageStatistics;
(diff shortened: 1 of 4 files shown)


asizemore:
Reminder - not to be merged until after b57 prod tags created.

src/lib/core/utils/visualization.ts
```diff
@@ -130,6 +134,55 @@ export function fixLabelForNumberVariables(
...
+      (varItem) => varItem.variableId === label
+    );
+    const variableDescriptor = variableDescriptors && variableDescriptors[0];
+    const retrievedVariable = findEntityAndVariable(
+      entities,
+      variableDescriptor
+    );
+    return retrievedVariable?.variable.displayName || label;
+  });
+  return displayNames;
```
dmfalke, line 162: You ought to be able to delegate to `fixVarIdLabel`:
```suggestion
  return labels.map((label) => fixVarIdLabel(label, variableList, entities);
```

dmfalke (commented):
Overall, this looks pretty good. I did suggest a change in an inline comment.

I don't really like all of the functions and variables that start with the word `fix` (e.g., `fixVarIdLabels`, `fixLabelsForNumberVariables`, `fixedOverlayLabel`, etc). It implies that _something_ is being modified, whereas I think it is really just an issue of formatting. I think I will make another issue addressing that, since that has become something of a convention.

Issue #1086, opened 2022-04-19
Apps: Allow computed vars to appear in variable coverage table


asizemore:
Variable coverage table in ranked abundance and alpha diversity vizs does not add the computed variables. Note the computed variables are returned in the complete cases table response data, but this info doesn't make it onto the screen.

<img width="1758" alt="Screen Shot 2022-04-19 at 4 19 09 PM" src="https://user-images.githubusercontent.com/11710234/164088716-a28610a6-50f0-4b0e-ba05-8cae9aa41cea.png">

Possibly related to VEuPathDB/plot.data#164

dmfalke:
Does the backend provide this data, and the frontend is simply not rendering it? Regardless, I just want to mention that we want to approach this in a generic way.

asizemore:
Good call. The backend does provide the information (see screenshot: feature mbio, GEMS, ranked abundance, no subset), but i'm guessing there is a check or filter somewhere on the front end to use only those vars in the table that match one of the input variables? I haven't looked into it deeply yet, that's just a guess. 

<img width="1766" alt="Screen Shot 2022-04-20 at 9 41 36 AM" src="https://user-images.githubusercontent.com/11710234/164243785-b0f84af6-d9b7-42a2-8377-e2ce922c041e.png">

Note that in the response, there is one row for each box in the boxplot. This is a consequence of using a collection-type variable on the x axis.

moontrip:
Please refer to what I temporarily did for alphadiv's variable coverage table before, but it is not a generic way to do
https://github.com/VEuPathDB/web-eda/pull/1018

I forgot a detailed info myself but there are several issues for abundance case, e.g., multiple values in completeCases, no specific info to search for X and Y variable details, etc. I may take a look at it later to remind myself :)

dmfalke:
I want to point you both (@asizemore and @moontrip) to https://github.com/VEuPathDB/web-eda/issues/580. Ultimately, I envision a component (currently referred to as `StandardVisualization`) that would handle this sort of thing. It would help to think about how one would make a generic component that will do what we're discussing here.

I'm working on a prototype of `StandardVisualization` right now. Hopefully we can look at it soon, and think more about this issue.

asizemore:
Thank you both!
@moontrip thank you for the reminder - I will remind myself what we discussed and think about how we might want to change the response. You brought up some good points and you're right i haven't addressed them yet.

@dmfalke very cool! Really looking forward to the prototype!

[...]

PR #1092, opened 2022-04-20
Issue 580/standard visualization

### LINKED ISSUE #580: Refactor visualization implementations
All of our visualizations have the same basic layout and data flow. This should be encapsulated in a component that handles the layout and data flow.

### Overview

All visualizations have the same basic layout, as illustrated in the image below:
image

The **Heading** section is provided by the `VisualizationsContainer` component, so we can ignore that. The other two sections are handled by individual visualization components. The goal is to create a component that can render these two sections with a minimal amount of configuration.

The **Inputs** section is typically rendered using the `InputVariables` component. In most visualizations, the specification for the inputs can be statically declared. However, there are increasingly more cases where this can be dynamic, either based on inputs provided by the App (computed variables), or based on the selection of other variables (see Y-aggregation options for line plots).

The **Plot and controls** section generally has the same structure across all visualizations: a plot and controls on the left, and legend and summary elements on the right. 

### Proposal

Implement a `StandardVisualization` component. This component will be responsible for rendering the sections **Inputs** and **Plot and controls** sections detailed above. Each visualization will specify how to render inputs, and how to render the plot and controls. `StandardVisualization` will wire the pieces together, handling data flow from inputs to plots, and will render the layout, included the summary elements.

We will likely want to incorporate some changes desirable for better handling integration with Apps. For example, an App instance can provide one or more variables to the visualization. This should be handled in a generalized way.

### TODOs

- [x] Prototype `StandardVisualization` to determine what the props should be.
- [ ] Cost what is involved with keeping things backward compatible. This has to do with how the viz config is specified.
- [ ] Specify changes for Apps integration.

### Additional details (copied from [confluence doc](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/12027390/App+viz+interactions))

*A collection of thoughts about the viz refactor and how apps may want to interact with visualizations in web-eda*

Apps may bring their own computed variables, opinions about plot options, and axes titles into a visualization. Below is a compiled list of all the overrides I currently know about or expect based on our planned apps. Note that the following is a union of all the app declarations, and each may not be applicable to all apps.

**Known overrides**
Variables:
- xAxisVariable. Example: Ranked abundance boxplot sets its own xAxisVariable `boxplot.config.computedVariableMetadata.collectionVariable.collectionVariableDetails`
- yAxisVariable. Example: Alpha diversity box specifies the yAxisVariable `boxplot.config.yVariableDetails`
- overlayVariable. Example: Ranked abundance scatter sets its own overlay variable
- facetVariable. Example: Beta diversity scatter does not allow any faceting

Plot metadata:
- x axis title. Example: Ranked abundance boxplot has info that should appear as x axis title
- x axis display range. Example: Differential abundance (b60 app) could set the x axis display range
- x axis vocabulary (used for reordering the x axis). Example: Ranked abundance boxplot should use the order of the `boxplot.config.computedVariableMetadata.collectionVariable.collectionVariableDetails`
- y axis title. Example: Alpha diversity box sets the y axis title based on the computed variable’s display name `boxplot.config.computedVariableMetadata.displayName[0]`
- y axis display range. Example: Alpha diversity box gets display range info `boxplot.config.computedVariableMetadata.displayRangeMin`
- overlay (legend) title. Example: Ranked abundance scatter. Derived from `computation.configuration.collectionVariable`
- overlay vocabulary (used for reordering the legend labels). Example: Ranked abundance scatter. Should use the order found in `boxplot.config.computedVariableMetadata.collectionVariable.collectionVariableDetails`
- plot title. Example: Ranked abundance boxplot should say if it cutoff the data (see live site for examples). Open to other suggestions of where this info could appear! 

Plot modes:
- Hide any scatterplot trend line options. Example: Beta diversity scatterplot should not allow trend lines
- Hide show missingness toggle. Example: Ranked abundance scatter (under discussion, see [Apps: allow no data toggle with computed variables? · Issue #1089 · VEuPathDB/web-eda](https://github.com/VEuPathDB/web-eda/issues/1089) )
- Allow statistics table to be toggled “on” in boxplot. Example: Ranked abundance boxplot

Others:
- Apps set a “base entity” for vizs [Apps: Enable app instance to set a "base" entity · Issue #952 · VEuPathDB/web-eda](https://github.com/VEuPathDB/web-eda/issues/952) . Example: Ranked abundance the base entity should be the collection variable’s entity.

dmfalke:
Fixes #580 

Changes in this pull request address the need to DRY up visualization implementations. This is a work-in-progress, so things are likely to change as requirements evolve.


asizemore:
This is really cool!
One thing im not quite following yet is what is the vision for how this gets used? Something like
```
<StandardVisualization>
    <Scatterplot />
</StandardVisualization>
```
or would it call the Scatterplot component inside based on the VisualizationProps part of the props?

asizemore:
This is really cool!
One thing im not quite following yet is what is the vision for how this gets used? Something like
```
<StandardVisualization>
    <Scatterplot />
</StandardVisualization>
```
or would it call the Scatterplot component inside based on the VisualizationProps part of the props?

dmfalke:
> This is really cool! One thing im not quite following yet is what is the vision for how this gets used? Something like
> 
> ```
> <StandardVisualization>
>     <Scatterplot />
> </StandardVisualization>
> ```
> 
> or would it call the Scatterplot component inside based on the VisualizationProps part of the props?

@asizemore that's a good question. This isn't really finalized, and a lot of details are missing, but I was thinking it would be used more like this:

```javascript
function ScatterPlotVisualization(props) {
  return (
    <StandardVisualization
      {...props}
      renderPlot={(data) => (
        <ScatterPlot data={data}/>
      )}
  );
}
```

dmfalke:
I have made some more progress on this PR. @asizemore see https://github.com/VEuPathDB/web-eda/blob/9b5d72d0feb310863683c069277cf141a4762a63/src/lib/core/components/visualizations/implementations/TestVisualization.tsx#L52-L95 for an example of how to use `StandardVisualization`.

At this point, I would love feedback about the current approach. Specifically, is the interface clear enough to tell a story, or is it utterly confusing? Note that I decided to use a factory function[^1] pattern for this interface, which returns a React component. The motivation for using a factory function, rather than using the `StandardVisualization` component directly, is that it removes some moving pieces, and it removes the need to use memoization hooks for props being passed to `StandardVisualization`. 

CC: @asizemore @bobular @chowington @jtlong3rd @moontrip 

[^1]: A _factory function_ is a function that returns an `object`. In this case, the `object` being returned is a React `Component`. Technically, a `function` is being returned, so this could also be called a _higher order function_. However, there are several different kinds of React `Components` (`function`, `class`, and `object`). Thus, conceptually, this is closer to a _factory function_.

src/lib/core/components/visualizations/StandardVisualization.tsx
```diff
@@ -0,0 +1,221 @@
...
+   */
+  mapShowMissingnessToConfig?: (showMissingness: boolean, config: U) => U;
+  /**
+   * Map configuration object to selected variables
+   */
+  mapConfigToShowMissingnes?: (config: U) => boolean;
+  /**
+   * A function that take a vizConfig object and returns a Promise for data.
+   */
+  getData: (vizConfig: U) => Promise<T>;
```
jtlong3rd, line 76: I'm guessing you'll want `getData` to receive the data client as an argument?

jtlong3rd (commented):
I think the "factory function" approach lends itself well to making "global" updates to all visualizations - it seems like you'd just alter the interface of `StandardVisualizationProps` and update each visualization's invocation of `makeStandardVisualizationComponent` to accommodate the new "shape" of `StandardVsualizationProps`.

src/lib/core/components/visualizations/StandardVisualization.tsx
dmfalke: That's a good possibility. I don't really see a downside. Thanks!

asizemore:
I was just looking at the discussion in #1086 and realized we had discussed the possibility of handling computed variables in the variable coverage table. Is the idea that if these variables/config are appropriately mapped via the inputs to `StandardVisualization`, then the variable coverage table would look there for all of the vars?

dmfalke:
> I was just looking at the discussion in #1086 and realized we had discussed the possibility of handling computed variables in the variable coverage table. Is the idea that if these variables/config are appropriately mapped via the inputs to `StandardVisualization`, then the variable coverage table would look there for all of the vars?

Yes! I'm taking an iterative approach to this. First, I'm going to get the existing visualizations working with this, including making the summary views (table, birds eye, etc) in `StandardVisualization`. Then, I will start adding the features we've been talking about for apps. We can revisit this plan of action if our timeline is too short.

dmfalke:
I'm definitely on board! I think we can ramp up development on this pretty soon. I think once we have an existing visualizations ported over, it won't be difficult to start adding the new features, and port the other visualizations in parallel. Maybe I should start with box or scatter...

Issue #1076, opened 2022-04-15
Change verbiage to "Plot mode" on scatter (& other places where plot mode exists)


danicahelb:
<img width="1215" alt="image" src="https://user-images.githubusercontent.com/25450900/163551659-27ca7d14-6595-43ca-b498-b4a5b00b930c.png">

Current verbiage says "Plot modes" but the singular "Plot mode" is more appropriate. Please update, thanks!

danicahelb:
looks good, thanks DK!

Issue #1015, opened 2022-04-04
Plot padding should not affect data ranges


danicahelb:
Discussed in 4/4/2022 EDA meeting

Padding in plots prevents data points from appearing on axes
Currently the padding looks like data (ie, shows up in the range selector in axis zoom controls); if the user truncates the padding, a yellow bar blocks the data even though only the padding was truncated out
Padding should instead be a few pixels of space around the edges of the plot; the range selector should indicate the actual data range, not the range of padding

danicahelb:
Plot padding looks much better, but there are now too many digits in the range box

<img width="985" alt="image" src="https://user-images.githubusercontent.com/25450900/166513230-49d6a3b0-31b7-41ae-aab6-045f17cffb0e.png">

I'm not sure what rule we were previously using... (4 decimal places? 3 sig figs?) but please implement that here also.

On the live site, the range has a more appropriate number of decimal places:
<img width="333" alt="image" src="https://user-images.githubusercontent.com/25450900/166513663-2e098046-1c36-44df-aa01-3bce5fb49a3a.png">

moontrip:
@danicahelb thank you for your review and check. I will take a look at it 👍

moontrip:
@danicahelb Can you please let me know what study you have used for making the screenshot?

moontrip:
@danicahelb FYI, the X-axis range (for all Vizs that use such a range control) utilizes backend response, either displayMin/Max or rangeMin/Max. Thus usually I did not have to fix decimal points. Y-axis range is data-based value, i.e., computed from data to find min/max, thus it certainly has a setting to fix decimal points, 4.

I could not find the study but from my guess, it is an exception case to have such a lengthy decimal points. The variable is Age in year unit, so for example a new-born baby has more likely day unit, which will have decimal points when converting to year. The value shown in your screenshot, 0.041666... is about 15 days 😃 Anyway I will try to set the decimal points for X-axis range too

danicahelb:
@moontrip it was LLINEUP

moontrip:
@danicahelb Thank you for your information! 👍 Will check it!

moontrip:
@danicahelb @dmfalke 

[...]

Issue #1098, opened 2022-04-21
Switch position of X-axis controls and Y-axis controls on all viz tools


danicahelb:
Axes controls for X- and Y-axes should always be shown in that order (meaning X-axis is listed first, or on the left).

moontrip:
Hi @danicahelb I wonder if we should always place the Y-axis controls in the right side for the case that there is no X-axis controls such as barplot and boxplot Vizs: currently it is in the left side (location of the X-axis controls).

danicahelb:
Discussed in 5/16/2022 EDA UX meeting

@moontrip
When there are no x-axis controls, the y-axis controls should be left justified (this is the way things currently are on the live site)

<img width="977" alt="image" src="https://user-images.githubusercontent.com/25450900/168656265-a80f7025-a2f5-4b4b-b5f8-5494a155784b.png">

moontrip:
@danicahelb Thank you for your information. I will comply with it 👍

aaronwlsong:
Looks good, closing ticket. 

Randomly chose some studies and explored their line plots, mosaic plot and scatter plots (which contains both x-axis and y-axis), seems that they are all in expected order (x-axis is on left and y-axis is on right). 
<img width="1715" alt="Screen Shot 2022-06-02 at 9 42 30 AM" src="https://user-images.githubusercontent.com/61077845/171644404-7d2670ab-23f7-4b2e-a42a-665a9f10a7ef.png">
<img width="947" alt="Screen Shot 2022-06-02 at 9 42 49 AM" src="https://user-images.githubusercontent.com/61077845/171644407-da1661b8-3750-4e7c-8e2b-475f71b32726.png">
<img width="1018" alt="Screen Shot 2022-06-02 at 9 43 36 AM" src="https://user-images.githubusercontent.com/61077845/171644409-bb1dc179-f60d-489c-bc0e-43ff167e5d37.png">
<img width="1419" alt="Screen Shot 2022-06-02 at 9 45 07 AM" src="https://user-images.githubusercontent.com/61077845/171644412-76df0bab-eb50-435b-b648-4f56144f98f7.png">
<img width="1053" alt="Screen Shot 2022-06-02 at 9 45 24 AM" src="https://user-images.githubusercontent.com/61077845/171644416-5a769dbd-7c22-48cb-a68e-7de44b536e16.png">
[...]

Issue #589, opened 2021-10-13
Subsetting: set y-axis range for histogram filter for log scale


moontrip:
This is also related to Phase 1 issue for histogram visualization, https://github.com/VEuPathDB/web-eda/issues/559

We want to set range min of 0 (count) so that y-axis tick labels are better shown in log scale. range max need to be computed from data as there is no variable metadata nor backend response: as of now backend only returns x-axis stats

danicahelb:
tested with LLINEUP, everything looks good to me!
<img width="858" alt="image" src="https://user-images.githubusercontent.com/25450900/171283147-2ba3180a-d421-45a2-9736-2774f561063f.png">

moontrip:
Thank you for your tests @danicahelb 👍

PR #1102, opened 2022-04-24
changing padding truncation logic

moontrip:
This will address #1015  with corresponding web-components PR, https://github.com/VEuPathDB/web-components/pull/346. Now, axis ranges shown in the input fields reflect data or metadata min/max, not including padding/margin value.

- changes
a) Tried to minimally change web-eda's Vizs, so key changes are made at web-components.
b) Padding/margins are decreased as requested
c) Also fixed a bug at boxplot concerning dependent axis range.
d) Also addressed #1076 in this PR. (see the second screenshot)
e) For histogram and barplot, I did not make padding/margin for dependent axis's min as it is typical in those plots. Not quite sure if I should have set this to boxplot as well.
f) truncations are also considered
g) and more stuff which I might have forgotten to say :)

For a comparison purpose and an example, below are two screenshots with GEMS1 case control and same X/Y variables. First one is taken from qa site and the other one is taken with this PR. One can see that both independent and dependent axis ranges show data's (or metadata's) min/max, not including padding/margin values.

## qa site
scatter-padding-qa

## this PR
scatter-padding

## this PR and truncations for demo purpose
scatter-padding-truncation


jtlong3rd (approved):

moontrip:
@jtlong3rd Thank you for your review and approval PRs at both web-eda and web-components.
I have addressed another issue in this PR, #1098. Some CSS values, etc. are changed but basically not much changes are made from the programming sense: this also includes histogram filter at Subsetting. Not clearly hear the answer for the case of Y-axis range controls only (asking Danica), but I presumed that it required to have spaces so that it is located in the right side (screenshot). If we don't need the space, I will change it accordingly.

Here are screenshots. 
scatter-XY-position

boxplot-XY-position

moontrip:
Additional ticket is addressed in this PR, #589 at https://github.com/VEuPathDB/web-eda/pull/1102/commits/65f53ad6460726d42694016506ed1f0d1200ab41. It is basically to set the dependent axis range for histogram filter at Subsetting so that its log scale shows from 0. 

While doing this, I also found a bug for the original padding/margin work so will address it as well.

Here are two screenshots taken from qa site and this PR for a comparison purpose.

### qa site
qa-histogram-filter-logscale

### current PR
dev-histogram-filter-logscale

dmfalke:
Hi @moontrip, nice work with this. I think it's great that you are addressing so many issues, but I think it might make sense to make new PRs for the other two issues, rather than reusing this one. @jtlong3rd has already approved the initial issue being addressed, so that work could be merged. If there are issues with the subsequent issues being addressed.

Is this something you would feel comfortable doing? This is what I would do:

- Make a new branch from this one, and make a PR for the other issues being addressed.
- In this branch, drop the commits, starting with https://github.com/VEuPathDB/web-eda/pull/1102/commits/827e3d8d8c8cbc6a278a0ff09b815b09f54541ed: `git reset --hard 75add6f` should do it.
- Once this PR is merged, update the new branch from `main`.

moontrip:
> Hi @moontrip, nice work with this. I think it's great that you are addressing so many issues, but I think it might make sense to make new PRs for the other two issues, rather than reusing this one. @jtlong3rd has already approved the initial issue being addressed, so that work could be merged. If there are issues with the subsequent issues being addressed.
> 
> Is this something you would feel comfortable doing? This is what I would do:
> 
> * Make a new branch from this one, and make a PR for the other issues being addressed.
> * In this branch, drop the commits, starting with [827e3d8](https://github.com/VEuPathDB/web-eda/commit/827e3d8d8c8cbc6a278a0ff09b815b09f54541ed): `git reset --hard 75add6f` should do it. You will need to do `git push --force-with-lease` after this.
> * Once this PR is merged, update the new branch from `main`.

@dmfalke Thank you for your advising. Yes normally I won't do like this, but somehow those issues are similar to each other and I was waiting for Bob's response: certainly I can merge with Jamie's approval only for sure 😄

One minor issue is that I found a bug at the originally approved PR and it would probably fixed at the corresponding web-components (WIP) so I will follow your suggestion after fixing it in there: the other works are done in the web-eda so it would be more convenient to fix it and then merge the PR.

bobular (commented):
Hi - sorry for taking a while to get to this.

I haven't tracked down the lines of code yet, but I noticed that lineplot with a date-based x-axis variable, when binned (e.g. 1 year), no longer has any padding at the left (and very tiny amount on the right).

GEMS1: diarrhoeal episode duration vs birth date
Old/main:
image

New/this branch:
image

bobular:
> Hi - sorry for taking a while to get to this.
> 
> I haven't tracked down the lines of code yet, but I noticed that lineplot with a date-based x-axis variable, when binned (e.g. 1 year), no longer has any padding at the left (and very tiny amount on the right).

Looks like Lineplot.tsx is using `extendAxisRangeForTruncations` - so maybe the Date Math is rounding to zero or something?

moontrip:
@bobular Thank you for your tests. Although I think you already did, but Just in case please note that this PR relies on web-components's PR, https://github.com/VEuPathDB/web-components/pull/346, so you should use yarn link. 

The reason why I am checking above is that I can actually see correct behavior with your example at my local dev. Below is my screenshot.

lineplot-birthDate

bobular:
Duh I am sorry @moontrip - I turned off yarn link due to the troubles earlier.  And obviously forgot to put it back on.

Works fine then.  Have already looked through the code - will approve.
image

bobular (approved):
Looks good. Glad it wasn't as tricky as we feared.

moontrip:
> Looks good. Glad it wasn't as tricky as we feared.

@bobular Glad it was not an issue! 😄 To reach that solution, I had to check several other logics 😄 Thank you for your reviews/test/approval! 👍

Issue #1019, opened 2022-04-05
Apps: Add app title div - basic function


asizemore:
Once we enter an app, currently there is no indication on the page of what app we're in. To solve this problem, let's add the app title + name to the top of each page. 

*NOTE:* This ticket is only for the basic implementation, and styling will be covered in a later ticket (styling not yet approved and we won't be doing mbio styling tickets for another month at least).

For example, the title would eventually look something like the below and would reside in that same location for each page within the app (select a visualization, making visualizations).

<img width="961" alt="Screen Shot 2022-04-05 at 6 48 24 PM" src="https://user-images.githubusercontent.com/11710234/161863473-a1a7b18d-188f-4a15-93ff-a8582a026a8d.png">

For this first implementation, the following information in the title is sufficient:
1. User-given app name. Ex. My first app.
2. App name. Ex. Ranked abundance.

asizemore:
Addressed in #1104 

After this is deployed, will assign to @dpbisme for QA. Please check that you can see all the app info you need, on any page you need it!

asizemore:
I deployed this to mbio feature, so moving to Ready for Testing!

asizemore:
Moving to complete

PR #1104, opened 2022-04-25
working app thumbnail page

### LINKED ISSUE #950: Apps: add thumbnail page that has all apps together
Currently we have apps and their vizs being saved, and links back to them. We'd like a page that holds all of the apps and has all of their vizs in thumbnail view. 

Mockup
Screen Shot 2022-03-08 at 4 29 27 PM
s

The above mockups assume we will have a compute service which would supply info like "Status", "Started at", etc, so we don't have to worry about those yet.

Major pieces to get functional:

- [ ] Table of contents/List of apps on the left.
- [ ] New app/module button. Goes to add-new-app page
- [ ] App instance section that includes app name, icons for copy/configure/delete, a New Visualization button, and app instance's thumbnails.

Not worried about styling in this ticket, just function :)

asizemore:
Resolves #950 and also #1019

The goal of this PR is to have a functional app thumbnail page. 

Done:

- [x] Check that nothing i've done so far messes up clinepi
- [x] Consider making different component for the list of configured apps? This component would basically just add the the computation title and such above the `ConfiguredVisualizations` on `/visualizations`. Or maybe make special title component for the configured visualizations view and a different one that renders at the top of any other computation instance route.
- [x] Make it look a _little_ nicer
- [x] Check again that nothing i've done so far messes up clinepi

Left for another issue

- [ ] Functional app copy/delete/modify configuration.
- [ ] All styling

Notes:
- Very curious if the approach i took is reasonable. Felt easy to do but possibly overloading `ComputationInstance` too much? 
- I put in a little bit of support for having two different app titles: one for the `/visualization` page (a small version that can go just above the app's vizs) and one to live above pages within the app (adding a new viz, configuring a viz, etc). In short, eventually these two titles should look pretty different from each other, but they would both have the same information. Should they be two separate components?
- the `web-common` upgrade was to combat an error when the site was building. I can take it out if this is not the right place to upgrade!

### DIFF
--- a/src/lib/core/components/computations/ComputationInstance.tsx
+++ b/src/lib/core/components/computations/ComputationInstance.tsx
@@ -9,7 +9,7 @@ import { ComputationProps } from './Types';
 export interface Props extends ComputationProps {
   computationId: string;
   visualizationTypes: Record<string, VisualizationType>;
-  baseUrl?: string;
+  baseUrl?: string; // right now only defined when *not* using single app mode
 }
 
 export function ComputationInstance(props: Props) {
@@ -63,7 +63,7 @@ export function ComputationInstance(props: Props) {
 
   return (
     <div>
-      <h3>{computation.displayName}</h3>
+      {baseUrl && <h3>{computation.displayName}</h3>}
       <VisualizationsContainer
         geoConfigs={geoConfigs}
         computation={computation}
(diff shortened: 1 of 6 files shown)


src/lib/core/components/computations/ComputationInstance.tsx
```diff
@@ -1,13 +1,20 @@
+import { string } from 'fp-ts';
```
dmfalke, line 1: Unused import
```suggestion
```

dmfalke (approved):
I think this is a great start! I noted one unused import, but otherwise I think this is good to merge

PR #1105, opened 2022-04-26
change X-Y control order and set Y range for histgram filter

moontrip:
Following @dmfalke's [suggestion](https://github.com/VEuPathDB/web-eda/pull/1102#issuecomment-1110010528), these works are branched out from axis-padding branch.  

This PR addresses two issues, #1098 and #589. My comments and screenshots can be found in https://github.com/VEuPathDB/web-eda/pull/1102#issuecomment-1108848018 and https://github.com/VEuPathDB/web-eda/pull/1102#issuecomment-1109909577, respectively


jtlong3rd (approved):

moontrip:
@jtlong3rd thank you for taking time to review this PR! 👍 It would be great if you can review this PR as well when you are available: you already reviewed it but I added a fix for histogram case that should not have padding in the X-axis range. https://github.com/VEuPathDB/web-components/pull/346

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -386,6 +405,8 @@ type HistogramPlotWithControlsProps = HistogramProps & {
   updateUIState: (uiState: Partial<UIState>) => void;
   filter?: DateRangeFilter | NumberRangeFilter;
   variable?: HistogramVariable;
+  // typing defaultDependentAxisRange
```
bobular, line 408: The comment above doesn't add any useful info

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -373,6 +390,8 @@ export function HistogramFilter(props: Props) {
           updateUIState={updateUIState}
           showSpinner={data.pending}
           variable={variable}
+          // pass defaultDependentAxisRange
```
bobular, line 393: Comment not really informative.

bobular (approved):
Took me a while to figure out what we were previously doing wrong in HistogramFilter (using Plotly's default y axis range) - and this solution looks good.

A tiny issue with the switched x/y controls.
We used to let the right-hand bin slider value go outside its div (I think).
Now it looks a little bit tight against the divider (the '22' below)

image

I'm not 100% convinced about putting the box/barplot y-axis controls in a consistent location. I feel that maybe it should have a greyed out x-axis control heading?  But this is not something for this ticket/PR. It might grow on me.  It looked a bit strange under a faceted plot, though.

Sorry for the delay getting to this!

moontrip:
@bobular Thank you for your reviews and comments. 👍 Here are my responses

- for the vertical bar and its spacing, yes it is quite tricky to deal with because the X-axis controls with bin/bin slider do not have consistent width for number and date variables. Will see if I can alleviate it a bit more.

- I was not quite sure how to locate locate of y-axis control for bar/box case, but for consistency I have placed it at the same location with other Vizs. Perhaps your suggestion also works: with empty X-axis range control title. Anyway it can be easily addressed later. 

- Will remove those unnecessary comments: usually they come with my initial (DKDK haha) not to forget about my changes, especially because I often worked with a couple of issues at the same time 😃

moontrip:
@bobular I have tested the case of Y-axis controls only as you suggested. Below is an example with barplot Viz. Is this better for you than empty space at X-axis controls? If you think so, then I will also work for boxplot Viz.

barplot-viz-gray-x-axis

bobular:
> @bobular I have tested the case of Y-axis controls only as you suggested. Below is an example with barplot Viz. Is this better for you than empty space at X-axis controls? If you think so, then I will also work for boxplot Viz.

Hi! I think that looks really clean/nice, but let's run it past @dmfalke and @danicahelb before going ahead.

moontrip:
@dmfalke Thank you for your input! Actually it was my initial work, but @bobular opined that it was a bit strange. That's the reason why I added the screenshot addressing Bob's suggestion.

danicahelb:
@moontrip 
When there are no x-axis controls, the y-axis controls should be left justified (this is the way things currently are on the live site)

image

moontrip:
@bobular @jtlong3rd @dmfalke 
Danica said we are better place the Y-axis controls at left side if there is no X-axis one: barplot and boxplot Vizs. So I have reflected the decision and a screenshot is attached

boxplot-y-axis control

PR #1108, opened 2022-04-29
force computedVar display ranges to be numbers

### LINKED ISSUE #1107: Apps: ensure display range min and max are numbers
See https://github.com/VEuPathDB/microbiomeComputations/issues/17 for more details.

In short, displayRangeMin and max within computedVariableMetadata are currently strings. They should be numbers (they are numbers in the study metadata for continuous vars). This issue was blissfully hiding until a recent commit exposed it.

asizemore:
Resolves #1107 

Goal: force the computed variable display ranges to be always numbers in the boxplot. 

At first i thought we should change the api to send along numbers instead of strings, but i realize we want to keep sending as strings so that we can handle dates in the future. So instead, the fix is to just force the values to be numbers here!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.5.0",
+  "version": "3.5.1",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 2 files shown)


moontrip (approved):
@asizemore yes it would work as, from my understanding, we do not have date variable for boxplot 👍 Briefly tested and it seems to work fine

Issue #1075, opened 2022-04-15
map viz needs "Plot mode" heading


danicahelb:
we should have a "Plot modes" heading for the map viz (for b58...), between the map and the marker selection radio buttons (bar count, bar proportion, pie). See styling of "Plot modes" in scatter plots

<img width="821" alt="image" src="https://user-images.githubusercontent.com/25450900/163552763-aadcf566-5d4d-4e06-bbf5-b7a70b73da6b.png">
 
Note that on scatter and other plots, we are converting "Plot modes" to the singular "Plot mode". See VEuPathDB/web-eda#1076

SAK22:
Looks good!

Issue #1037, opened 2022-04-11
Map viz tool need variable completeness table


danicahelb:
Please add a variable completeness table to the map visualization. The table should include information on the main variable as well as for lat/long coordinates

chowington:
Here's what I have currently:

![image](https://user-images.githubusercontent.com/12377649/166307873-9e2252d7-d2e4-4a5c-9080-b86e94917965.png)

The backend doesn't give separate values for lat & lon, only the completeness for the geohash variable, which has a nonsensical name to users. I assume the geohash value exists iff lat & lon both exist, so logically it's pretty much equivalent. I don't see a huge need to distinguish between latitude and longitude for variable completeness purposes, so I'm personally fine with just the geohash variable.

The question is what to call it in the table. "Latitude & longitude"? Just leave the geohash name? Another name?

@danicahelb @d-callan

d-callan:
I'd be fine w 'geolocation', it's less wordy than saying 'latitude and longitude' but I think/hope equally understandable to most

SAK22:
Map tool currently down on QA.

SAK22:
Looks good!

PR #1110, opened 2022-05-02
Add variable coverage table to map viz

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1075
Fixes https://github.com/VEuPathDB/web-eda/issues/1037

Just waiting on potential verbiage change for variable name in table (instead of "GEOHASH_X"). See second ticket for further discussion. EDIT: Resolved.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
@@ -35,6 +35,7 @@ import { preorder } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
 import DataClient, {
   MapMarkersRequestParams,
   PieplotRequestParams,
+  PieplotResponse,
 } from '../../../api/DataClient';
 import { useVizConfig } from '../../../hooks/visualizations';
 import { usePromise } from '../../../hooks/promise';
@@ -57,6 +58,7 @@ import { useCheckedLegendItemsStatus } from '../../../hooks/checkedLegendItemsSt
 import { variableDisplayWithUnit } from '../../../utils/variable-display';
 import { BirdsEyeView } from '../../BirdsEyeView';
 import RadioButtonGroup from '@veupathdb/components/lib/components/widgets/RadioButtonGroup';
+import { VariableCoverageTable } from '../../VariableCoverageTable';
 
 export const mapVisualization: VisualizationType = {
   selectorComponent: SelectorComponent,
@@ -374,7 +376,7 @@ function MapViz(props: VisualizationProps) {
    * Now we deal with the optional second request to pieplot
    */
   const proportionMode = vizConfig.markerType === 'proportion';
-  const pieplotData = usePromise<PieplotData | undefined>(
+  const piePlotResponse = usePromise<PieplotResponse | undefined>(
     useCallback(async () => {
       // check all required vizConfigs are provided
       if (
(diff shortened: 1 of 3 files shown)


bobular:
Hi @chowington - I've had a look at this and there are a few things to sort out still.

Actually it might just need one thing sorting!

The `pieplot` request now accepts a `viewport` parameter (this didn't exist until recently, I think).  `map-markers` already has this, so we can follow its lead. Note that there will no longer be any need for `filtersPlusBoundsFilter` and the pieplot request can take the regular `filters || []` 

This should mean that zooming/panning no longer affects the completeness table, which is what we want here.

I'll have a think about what the appropriate numbers for the birds eye plot are...

bobular:
BirdsEyeNumbers

`completeCasesAxesVars` should take the result from map-marker's `response.config.completeCasesGeoVar`
`completeCasesAllVars` could be pieplot's `response.config.completeCasesAllVars`

HOWEVER - the birds eye labels are hardcoded in the component (my bad) and aren't quite right for the map viz. They aren't horribly wrong either, so for now let's leave them, and I'll make a new ticket to sort those out.

I'll also need to double check that these numbers I've suggested are correct.

Thanks!

bobular:
Update - awaiting clarification on `viewport` parameter to `pieplot` endpoint which is giving some trouble.

d-callan:
@asizemore was looking into the reported viewport issue, and based of conversation w her it seems like this can be resolved by including the latitude and longitude variables explicitly in the request.

bobular:
> @asizemore was looking into the reported viewport issue, and based of conversation w her it seems like this can be resolved by including the latitude and longitude variables explicitly in the request.

Oh yeah, of course - it needs to know which lat/long variables to use!

bobular:
Testing this, it seems that when zoomed in (e.g. after the fly-to with SCORE Seasonal using Participant->Sex) - the back end response `pieplot.data` is an empty array.
image


If you zoom out enough, the single marker response comes back OK:
image

chowington:
Thanks for testing, Bob! I discovered this too and sent @d-callan and @asizemore a message on Slack....neglected to mention it here though!

bobular:
Any news, @chowington ?

chowington:
Waiting on @d-callan to look into the above issue

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -374,63 +363,89 @@ function MapViz(props: VisualizationProps) {
    * Now we deal with the optional second request to pieplot
    */
   const proportionMode = vizConfig.markerType === 'proportion';
-  const pieplotData = usePromise<PieplotData | undefined>(
+  const piePlotResponse = usePromise<PieplotResponse | undefined>(
```
bobular, line 366: Sorry to be pedantic, but could `pieplotResponse` capitalisation be consistent with `pieplotData` below?

Further to my previous comment (before this review) that Dave :+1:ed - don't change the general structure (separating `pieplotResponse`) but use a useMemo and no usePromise.

You'll have to do something with `<MapVEuMap showSpinner ../>` value.  Presumably `pieplotResponse.pending` instead of `pieplotData.pending`.  And obvs check other uses of `.pending` and `.value`.  It's hard to test if it's working at present because the second request comes back just as fast as the first.

bobular (changes requested):
Looks good if we just make the small change suggested.

bobular (approved):
Cool, thanks!

PR #1123, opened 2022-05-04
set four decimal points for independent axis range

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/1015#issuecomment-1116389339

The same study and variables shown in the comment's screenshot is reproduced with this PR as follows. Now the decimal point is limited up to 4 digits: see X-axis range's min value. I also changed existing function to check whether the number value is integer or not because there is no need to have additional decimal point for the integer case.

scatter-LLINEUP-X-range-decimal-points

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.5.2",
+  "version": "3.5.3",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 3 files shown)


jtlong3rd (approved):

moontrip:
@jtlong3rd ! Are you watching my works in real-time??? 😄 I was even modifying my initial comment when you just approved! 👍

jtlong3rd:
> @jtlong3rd ! Are you watching my works in real-time??? 😄 I was even modifying my initial comment when you just approved! 👍

😆 We recently added a #github-pull-requests channel to the EPVB slack, so I got a push notification when you added me as a reviewer.

moontrip:
@jtlong3rd Aha I see! I had an experience to make it at another slack myself. But it is soooo fast anyway 👍👍

Issue #1103, opened 2022-03-24
Visualizations: Stronger indicators for zoomed visualizations


d-callan:

dmfalke:
I added the `question` label because it seemed like there wasn't a conclusion on how to address this (just a suggestion), so UX discussion is necessary. I think that is what we use that label for.

bobular:
Thanks. There's also discussion needed about how the warning triangle will be triggered:

- With server-side viewport/zooming (e.g. x-axis of histograms and XXX) we know for sure if there are points not shown due to viewport/zooming because we can compare various numbers from the response.
- WIth client-size zooming (e.g most dependent axis range zooming) we don't currently make a data-based decision on whether or not to show the yellow warning **bars**. The bars are shown if the zooming **could** truncate data (e.g. when zooming into a subset of the annotated range, but not when zooming out beyond it) but not if they actually *do* truncate data.  I don't quite remember the decision process behind this, but it does prevent possibly yet-another-pass-through-the-data on the client side. 

The logic for the yellow triangle could simply be `any yellow bars are shown` and then there is no real debate needed.

I'll add the `question` label back again!

moontrip:
@danicahelb for your information, I have tried to make yellow warning icon (hazard symbol) and its tooltip above the plot. Unfortunately, yellow icon is not clearly visible under white background so tried a couple of different colors and round icon: also shows a tooltip implementation when mouseover to the icon.

![barplot-truncated-warning](https://user-images.githubusercontent.com/12802305/166982596-c93fc420-1359-4409-af66-bca49e2f0b13.png)

danicahelb:
Thanks @moontrip. We discussed this ticket in today's UX meeting and still have a lot of debate about what should be done, so I'm going to add a paused label & move out of this cycle for now.


**Discussion from today (these are ideas not decisions):**

1. Range truncation messages should also be in yellow (to match the color of the color of the zoom indicator(s) on the plot
2. Just have the warning icon next to the “Range” label, the rest of the information in the warning message can be moved to mouseover information
3. Move the warning message to the top next to the plot title (ie, after XX Communities), and simplify (we don’t need separate text for y-axis and x-axis truncations
4. For the map, when the user has zoomed in and removed some data from view, the title of the plot should change to “Viewing 12 out of 16 communities”
[...]

PR #1124, opened 2022-05-05
testing warning icons for axis range truncation

moontrip:
This is a draft PR to test warning icon above any zoomed plot, discussed #1103: not cleaned up as testing several icons. 

Yellow icon is not clearly visible under white background so tried a couple of different colors and round icon as follows: also shows a tooltip implementation  when mouseover to the icon.

barplot-truncated-warning

### DIFF
--- a/src/lib/core/components/visualizations/OutputEntityTitle.tsx
+++ b/src/lib/core/components/visualizations/OutputEntityTitle.tsx
@@ -2,15 +2,34 @@ import { makeClassNameHelper } from '@veupathdb/wdk-client/lib/Utils/ComponentUt
 
 import { StudyEntity } from '../../types/study';
 import { makeEntityDisplayName } from '../../utils/study-metadata';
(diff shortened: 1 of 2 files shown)


dmfalke:
@asizemore do you have any thoughts about the color to use here? It seems like it should be similar to what we decided for the `Banner` component.

asizemore:
Ooof hard question. It's a warning, so i want to say yes it should be the same as for `Banner`, but it's also referencing yellow bars on the page, so i am also inclined to color it the same color as those yellow bars (but that gives the readability issue). 

There are at least two options forward. One is to use the same color as we will for `Banner` (needs discussion with styling group still) for the bars and the icon. A second option is to outline the icon with a thin line or to place it on a dark gray background.

What do you both think?

dmfalke:
I think we should first try using a darker background or border. Could be a chip-type thing? I did this quick and the padding is off, but you get the idea: 
image

Issue #1059, opened 2022-04-12
Apps: Show apps as disabled if they do not have a client plugin implemented


dmfalke:
Similar to what we do with visualizations, we should show apps as disabled (grayed out) if a client plugin does not exist for it.

asizemore:
Looks good! Moving to complete

PR #1125, opened 2022-05-06
Show apps as disabled when client plugins are not implemented

jernestmyers:
Addresses #1059 

 If `!singleAppMode`, the disabling logic is handled in the `StartPage` component by using the CSS property `pointer-events: none` for apps missing client plugins. As in `ClinEpi`, disabled implementations are rendered after enabled implementations.

image

### DIFF
--- a/src/lib/core/components/computations/StartPage.tsx
+++ b/src/lib/core/components/computations/StartPage.tsx
@@ -1,14 +1,16 @@
-import React from 'react';
 import { Link } from 'react-router-dom';
 import { ComputationAppOverview } from '../../types/visualization';
+import { ComputationPlugin } from './Types';
+import { orderBy } from 'lodash';
 
 interface Props {
   baseUrl: string;
   apps: ComputationAppOverview[];
+  plugins: Record<string, ComputationPlugin>;
 }
 
 export function StartPage(props: Props) {
-  const { apps, baseUrl } = props;
+  const { apps, baseUrl, plugins } = props;
   return (
     <div>
       <h3>To get started, select an app.</h3>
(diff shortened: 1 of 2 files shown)


src/lib/core/components/computations/StartPage.tsx
```diff
@@ -18,22 +20,35 @@ export function StartPage(props: Props) {
...
-              borderRadius: '.5em',
-              margin: '1em 0',
-            }}
-          >
-            <div style={{ fontWeight: 'bold', fontSize: '1.2em' }}>
-              {app.displayName}
+        {/* orderBy renders available apps ahead of those in development */}
+        {orderBy(apps, [(app) => (plugins[app.name] ? 1 : 0)], ['desc']).map(
+          (app) => (
+            <div
```
asizemore, line 26: Is there a reason it's beneficial to have the outer element be a div instead of a Link? Noting that the hierarchy is swapped.

We'll eventually change these buttons to real buttons somehow (that's on me to figure out still!) so for now I think it's good either way, just curious for my own education.

asizemore (commented):
Just left one quick question. Otherwise looking good!

src/lib/core/components/computations/StartPage.tsx
jernestmyers: Great question! I altered the hierarchy in order to use `cursor: not-allowed` when the link is disabled via `pointer-events: none`. Setting pointerEvents to none essentially disables the mouse event and thus the `cursor: not-allowed` wasn't showing properly. 

[But in re-reading the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#values), it looks like I could explicitly set the children's `pointer-events` property to auto and maintain the previous hierarchy? We'd need to make a few tweaks to allow the `divs` to occupy the full space of the `Link` such that hovering over any part of `Link` will show the disabled hover effect.

asizemore: Ahhh interesting! Thank you for the explanation!

I think it's okay to leave as is. Since we're going to redo the button styling in a few weeks anyways, it's probably not worth the time now getting them to work a certain way when we may change it/use a different component/etc.

asizemore (approved):

dmfalke (approved):

Issue #992, opened 2022-03-22
Apps: apps make 2 instances when the analysis gets saved


asizemore:
When creating a new app (any app), sometimes a second app instance is also created accidentally.

Steps to reproduce: 
1. Click to add a new app (alpha div, for example)
2. Click Add Computation
3. Go back to the Visualize page to check the list of apps
4. See if there were two copies initialized.

asizemore:
Ahh i think it happens only on the first app that gets created. After clicking on add app the first time, it saves the analysis. Then we get back to that page that has the add app button.

asizemore:
another observation: if the analysis has already been saved (for example because we starred some vars in the subset tab), this does not happen

asizemore:
Looks great! Moving to complete

PR #1126, opened 2022-05-09
Fix redirect when creating a new app in a new analysis

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/992

### DIFF
--- a/src/lib/core/hooks/analysis.ts
+++ b/src/lib/core/hooks/analysis.ts
@@ -15,7 +15,7 @@ import { isNewAnalysis, isSavedAnalysis } from '../utils/analysis';
 import { useAnalysisClient } from './workspace';
 
 /** Type definition for function that will set an attribute of an Analysis. */
-type Setter<T> = (value: T | ((value: T) => T)) => void;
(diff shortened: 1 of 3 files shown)


asizemore (approved):
Very cool! Confirmed I no longer see the bug!

So it seems like the big change here is that the createAnalysis function and the Setter are returning a promise instead of nothing. For my education, can you please explain why that solves the problem?

src/lib/core/hooks/analysis.ts
```diff
@@ -15,7 +15,7 @@ import { isNewAnalysis, isSavedAnalysis } from '../utils/analysis';
 import { useAnalysisClient } from './workspace';
 
 /** Type definition for function that will set an attribute of an Analysis. */
```
jtlong3rd, line 17: I would update this comment to indicate what the returned promise represents.

jtlong3rd (approved):
Clever! I can see how updating the `Setter` signature gives you finer control over when the two competing URL redirects occur.

asizemore:
Ahhh that makes a lot of sense! Thank you so much for the detailed explanation @chowington that was really helpful!

Issue #935, opened 2022-02-14
Styling updates for subsetting results modal


danicahelb:
Post b56, consider: 

1. Should the variable tree (selector) always be shown open on the left side of the modal (to mimic the styling on the Browse & subset tab)?

2. If the variable tree is shown open on the left side of the modal by default, should the tree be collapsable to give more real estate to the table?

3. Should the variable tree be expanded by default to show all variables? or just all parents & sub-parents? or just entities?

Edit (Connor): See [Dave's comment below](https://github.com/VEuPathDB/web-eda/issues/935#issuecomment-1123141782) for the final reqs for this ticket

danicahelb:
discussed in 3/21/2022 EDA UX meeting. @danicahelb will make mockups based on feedback from the meeting:

**In the subsetting results download modal:**

1. Make it look more like a table that the user needs to configure: 4 ghost columns vs 1 single ghost column. Perhaps add a gear icon to the ghost column headers?
2. Change the “Add columns” text to “Build the table” 
3. Try out having the variable tree (with check boxes to select / deselect variables) open on the left by default, and putting text between the Entity numbers and the tree / table that says “Build a table to download” (in this option, the “Add columns” button would be removed)
4. If the variable tree is shown open on the left side of the modal by default, should the tree be collapsable to give more real estate to the table?
5. Should the variable tree be expanded by default to show all variables? or just all parents & sub-parents? or just entities?
6. “X” icons in column headers to delete those columns?

**On the download tab:**
1. Update the text under “My subset” to be more user friendly
2. Somehow indicate that rows come from subsetted data, columns are what you are selecting

danicahelb:
Discussed in 4/25/2022 EDA UX meeting.

We debated having 3 step vs 2 step process, and decided to stick to 2 step for now

Mockups were approved, and now require updates from styling team. Assigning to @asizemore

dmfalke:
I'm giving this issue a high cost, given the substantial number of changes. We should get someone working on this asap, so that it can be ready for b58.

As I understand the latest mock-ups, these are the necessary changes:
- [ ] New 2-column layout, with headings
- [ ] Create "cartoon" table graphic
- [ ] Show "cartoon" table when no non-required columns have been selected
- [ ] Add **Required columns** section above variable tree
- [ ] Add **lock** and **remove** icons to table column heading cells
- [ ] Allow table to be horizontally (and vertically?) scrolled, independent of controls that are below the table
[...]

PR #1135, opened 2022-05-13
Implement 2-column subsetting modal layout

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/935
Depends on https://github.com/VEuPathDB/CoreUI/pull/70


jernestmyers:
The download modal looks **_really_** good! I'm only playing with the UI currently and haven't yet dug into the code, but I do wonder about some stylistic things that stood out:

1. Should we add some more space around the "X" that removes rows, and possibly reduce its size?
2. Do we want any sort of hover effect on the "X"?
3. Should the lock icon be centered vertically in line with the column header text? It appears to be centered vertically [in the Figma mockup](https://www.figma.com/file/P4jtVBKMdq18fc4C9HqVGp/Subsetting-results-download-modal?node-id=351%3A20227).

image

chowington:
Thanks @jernestmyers for the comments! I'll respond here:

1. I can look into this for sure
2. Yeah, hover effects are helpful! I sometimes skimp on them when doing styling with React's `style` prop because it doesn't support them, but that's not a good excuse 😅 
3. I choose to put the lock icon in the corner like the X button is mainly because having it vertically centered looked odd when the headings had multiple lines of text (which in practice happens a lot), but also I thought it gave a nice visual link between the two icons.

chowington:
Here's the updated styling:

image

On hover:

image

chowington:
If you want to see the lock icons centered with multiple lines of text, I can reimplement that and take a screenshot of it too

jernestmyers:
> If you want to see the lock icons centered with multiple lines of text, I can reimplement that and take a screenshot of it too

I agree with your point about how it would look strange when the column headers occupy multiple lines.

I wonder if we would prefer that the `VariableList` or `MultiSelectVariableTree` containers be scrollable when the variables are expanded, as opposed to the modal itself becoming scrollable? When there are a lot of required columns, real estate is limited so scrolling within the variable container would be awkward until the user collapses the required and featured variable sections. This is likely not urgent for the demo purposes but probably worth clarifying the desired behavior.
image

chowington:
Done! Thanks Jeremy!

Issue #931, opened 2022-02-23
Include mergeKey vars in the subsetting results table by default


danicahelb:
in the subsetting results modal the table is currently just showing columns for the primary keys (PID, HHID, etc) and the user has to configure the table by selecting other variables of interest. 

we may want to also include “mergeKey” variables in the default, un-configured table (ie, date or study timepoint variables in repeated measures entities). If we did this, the default table would look like this:

<img width="1262" alt="image" src="https://user-images.githubusercontent.com/25450900/155403214-d414f895-3357-49d5-abd6-0a40888cf3a5.png">

If we imlemented this, how would we indicate that "mergeKey" variables are somehow special & warrant default inclusion in the table?

danicahelb:
Discussed in 4/11/2022 EDA UX meeting. Decision is to always include the mergeKey variables in the subsetting results tables for their appropriate entity

danicahelb:
We will still want the same primary and foreign keys as is currently in each of the subsetting tables for each entity.

The only additional column(s) will always be annotated in the owl file. the annotation property is "mergeKey" = yes

danicahelb:
Looks great!


![Image](https://user-images.githubusercontent.com/25450900/169186090-8714327e-29e3-4054-9429-9fbd319000de.png)

PR #1136, opened 2022-05-16
Default to include mergeKey column in download modal results table

jernestmyers:
Addresses #931 

Added logic to the `fetchPaginatedData` callback that concatenates the `mergeKey` ids of the `currentEntity` with the `selectedVariableDescriptors` state.

### DIFF
--- a/src/lib/workspace/Subsetting/SubsettingDataGridModal.tsx
+++ b/src/lib/workspace/Subsetting/SubsettingDataGridModal.tsx
@@ -148,12 +148,20 @@ export default function SubsettingDataGridModal({
   const fetchPaginatedData = useCallback(
     ({ pageSize, pageIndex }) => {
       setDataLoading(true);
+      if (!currentEntity) return;
+      const mergeKeys = () => {
+        return currentEntity.variables
+          .filter((variable) => 'isMergeKey' in variable && variable.isMergeKey)
+          .map((mergeKey) => mergeKey.id);
+      };
 
       subsettingClient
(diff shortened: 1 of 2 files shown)


dmfalke (commented):
I think this looks pretty good. To address the issue with a user selecting the merge key in the variable tree, you can prepend the merge keys here:

https://github.com/VEuPathDB/web-eda/blob/9da231c5f82a43159c7bebb6981e7dd0072f5515/src/lib/workspace/Subsetting/SubsettingDataGridModal.tsx#L350

Eventually, we want to also be able to make that item `readonly`, but that will likely require some additional work, and can be addressed as a part of https://github.com/VEuPathDB/web-eda/issues/935.

src/lib/workspace/Subsetting/SubsettingDataGridModal.tsx
```diff
@@ -145,23 +145,36 @@ export default function SubsettingDataGridModal({
...
   const fetchPaginatedData = useCallback(
     ({ pageSize, pageIndex }) => {
-      setDataLoading(true);
       if (!currentEntity) return;
-      const mergeKeys = () => {
-        return currentEntity.variables
-          .filter((variable) => 'isMergeKey' in variable && variable.isMergeKey)
-          .map((mergeKey) => mergeKey.id);
-      };
+      setDataLoading(true);
```
dmfalke, line 167: Good catch!

dmfalke (approved):

PR #1137, opened 2022-05-16
Make proportional chart markers show entity count, not sum of proportions.

### LINKED ISSUE #1114: Bar plot: proportion markers on the Map Viz Tool should display counts
Discussed in 5/2/2022 EDA UX meeting.

Bar plot: proportion markers have a proportion ranging from 0-1 under the x-axis, which looks odd and is confusing (especially when all values are included, “1” is displayed on every marker). For example: "age group" in SCORE seasonal.

We decided that -- instead of displaying the proportion for the bar plot: proportion markers -- we will display counts making up each marker graphic. This is in line with what is being displayed for both bar plot: count markers and also pie markers.

Previous discussion can be found in this closed ticket: https://github.com/VEuPathDB/web-eda/issues/1025

NOTE FOR TESTERS: for the first time the back end's `sampleSizeTable` responses are parsed "in anger" by the front end. The parser needed some work and may fail (with pink errors) for other visualizations in edge cases (main cases have been tested). Please report any issues!

bobular:
Resolves #1114 
Depends on https://github.com/VEuPathDB/web-components/pull/351

Note: It will show occasionally some blank marker plots with pink response parsing errors (for variables like Village Study Arm in particular) until https://github.com/VEuPathDB/EdaDataService/issues/187 is addressed.

Note that the number in the markers doesn't reset to zero any more during the two-step zoom process (first show markers, then show the mini plots).


src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -417,14 +417,12 @@ function MapViz(props: VisualizationProps) {
...
-                      item.facetVariableDetails[0].value === geoAggKey
-                  )?.size
-                )
+              entityCount: sum(
+                response.sampleSizeTable.find(
+                  (item) =>
+                    item.facetVariableDetails != null &&
+                    item.facetVariableDetails[0].value === geoAggKey
+                )?.size
               ),
```
dmfalke, line 426: Do you really need to use `sum` here? Maybe my eyes are deceiving me, but it looks like you are summing a single value.

bobular: However, it's correct.  the `size` property is an array.  IIRC correctly if the facet variable was Sex, then the values in `size` would be for male and female.

dmfalke: Ah, makes sense!

PR #1138, opened 2022-05-17
Use displayType annotations for geoConfig variable discovery

### LINKED ISSUE #979: Map visualization: make use of new `displayType` values in study variable metadata (frontend)
See #977 for prior work.

When all mappable studies are loaded with the new `displayType` values for geo-related variables can be used by the map visualization code in `entityToGeoConfig` in `geoVariables.ts`.

bobular:
Resolves #979 

TESTING: no new functionality here. All we can do is check that a) the map works and b) the map viz option is available in the appropriate projects (SCORE Seasonal, SCORE Mozambique, UMSP) and not in the others.


chowington (approved):
Looks good and works!

Issue #971, opened 2022-03-15
Map visualization: persist "mouse pointer mode" 


bobular:
The state of this little guy should persist in vizConfig

![image](https://user-images.githubusercontent.com/308639/158412541-af9270a5-85c2-4397-92f0-7f8440e28107.png)

It may need some web-components work on `MapVEuMap` because it doesn't have a `mouseMode` prop. Instead it has local state. While we are at it, we could remove the comments that start with "this is the React Map component" because I think they have long-lost the code they were referring to!

danicahelb:
This ticket is for situations where the user changes the mouse mode from the default, leaves the visualization, and then re-opens the viz. When this happens, we want to the users mouse mode selections to persist selections when they leave and come back

danicahelb:
@moontrip i moved this out of "this cycle" as it is only a nice to have and we are in crunch time for b58. thanks!

SAK22:
Looks good!

PR #1141, opened 2022-05-19
remembering the selection of mouse tool

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/971
Note that this replies on a web-components PR, https://github.com/VEuPathDB/web-components/pull/354


src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -106,6 +107,10 @@ const MapConfig = t.intersection([
       zoomLevel: t.number,
     }),
     baseLayer: t.keyof(baseLayers),
+    mouseModeConfig: t.keyof({
```
bobular, line 110: I would just call it `mouseMode`.

Is there any way to use the `MouseMode` type from web-components? To make it future proof (again)

bobular (approved):
Looks good - just made a couple of minor suggestions.

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
moontrip: @bobular I have changed it to mouseMode. But I couldn't figure out a clear way to use that union of string literals in the MouseMode type from web-components to a io-ts style. So, I did not change that part.

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -109,6 +111,10 @@ const MapConfig = t.intersection([
       zoomLevel: t.number,
     }),
     baseLayer: t.keyof(baseLayers),
+    mouseMode: t.keyof({
+      default: null,
+      magnification: null,
+    }),
```
dmfalke, line 117: I would move this to the `t.partial` section below so that existing map visualizations don't break (this makes the config setting optional).

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
dmfalke: You can do this
```typescript
    mouseMode: t.keyof<Record<MouseMode, null>>({
      default: null,
      magnification: null
    })
```

bobular (approved):
+1 for Dave's suggestion to make the config backwards compatible, but other than this I appprove :-)

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/971

Issue #1095, opened 2022-04-21
Apps: Organize vizs in thumbnail page by app type and parameters


asizemore:
05/13/22 Updates in a comment below.

For this ticket, let's apply the styling of app info and vizs seen in the below screenshot. 

<img width="714" alt="Screen Shot 2022-04-21 at 11 55 55 AM" src="https://user-images.githubusercontent.com/11710234/164501613-a5650304-379d-423c-ab81-79db8413f0b5.png">

In short, each app should have its user-given name, the app's display name, and icons+styling as a sort of heading that separates the apps. Each app's visualizations are in thumbnails right underneath the heading, and next to the + New Visualization button.

asizemore:
See v6.1 in figma file for more details.

Let's organize the visualizations on the thumbnail page by 1) app type, 2) parameter set. In web-eda, if each viz is added to the app instance that matches its app type and parameter set, then implementing this layout should just be going through each app instance and having the app display name in a header (for the first instance of that app type), and then the  parameters listed within a subheader. 

Let's try ordering app types and parameter sets by most recently made. So if i go make a new ranked abundance viz with a brand new configuration, when i go to the thumbnail page i see:
Ranked abundance
    The new viz i just made
    Second most recently used ranked abundance parameter set and associated vizs
    Third most recently used param set...
Second most recently opened app type
... 

<img width="1338" alt="Screen Shot 2022-05-13 at 11 45 11 AM" src="https://user-images.githubusercontent.com/11710234/168319727-25622606-b21e-47e2-9638-5c8e082e3146.png">

dmfalke:
The behavior of clicking the **New Visualization** button is not yet confirmed (see https://github.com/VEuPathDB/web-eda/issues/1130). Let's hold off on making that button to anything, until we have sign-off, which will hopefully be soon.

asizemore:
Confirmed we're good with the behavior in #1130

asizemore:
Closing and assigning to dan for QA. Deployed on feature.

Issue #1132, opened 2022-05-13
Apps: move configuration into the full screen visualization view


asizemore:
Move app configuration _into_ the fullscreen viz as shown below.

Notes: 
- The intent is to have app parameters look just like viz parameters, but be on their own line with the heading "{app display name} paramters". 
- For the foreseeable future, all app parameters are required for the visualizations (unlike strata)

<img width="1578" alt="Screen Shot 2022-05-13 at 2 16 59 PM" src="https://user-images.githubusercontent.com/11710234/168343501-9653448d-f59e-456d-ab82-ae6edcb2bfd9.png">

asizemore:
please note those buttons in the screenshot are my lazy attempt at remaking the standard mesa buttons that we currently use for visualization variable selections

asizemore:
Closing and assigning to dan for QA

Issue #1131, opened 2022-05-13
Apps: Select a visualization page organization


asizemore:
When _not_ in single app mode, let's organize the visualization selection page by app. 

For example, we'd first get a list of the apps associated with this project. For each app, create a row and display the app display name and description, then the visualization options. 

Notes:
1. I'll leave a non-alphabetical ordering of apps and/or vizs for another ticket since it's a nice to have feature for later.
2. For apps that are not yet implemented, what we'd like to do is still have a row for the app but have the planned visualization grayed out like on the clinepi live site. For beta diversity, i believe this means we need to make a plugin for beta div in web-eda. The [data service](https://github.com/VEuPathDB/EdaDataService/blob/master/src/main/java/org/veupathdb/service/eda/ds/metadata/AppsMetadata.java) already knows that beta div will eventually have a scatterplot, but we need to have (i believe?) a plugin and then just don't list any visualizations for it, correct? That way we can leverage the current way of saying "this viz is coming soon"?

<img width="1002" alt="Screen Shot 2022-05-13 at 1 45 10 PM" src="https://user-images.githubusercontent.com/11710234/168339032-15cda04e-f293-4d11-a9b7-0fece6265ed3.png">

asizemore:
Issue for making [beta div plugin](https://github.com/VEuPathDB/web-eda/issues/1148)

To be clear, i'm thinking that if we have a beta div plugin and then 
```
export const plugin: ComputationPlugin = {
  configurationComponent: AlphaDivConfiguration,
  visualizationTypes: {
    // scatterplot: scatterplotVisualization,
  },
};
```
in the beta div plugin, then we should be able to leverage the current clinepi method for setting visualizations as "Coming soon!". That way we'd see something like (but with the current clinepi Coming soon styling):

<img width="1382" alt="Screen Shot 2022-05-23 at 12 26 04 PM" src="https://user-images.githubusercontent.com/11710234/169891935-fa914c3e-9318-456e-a452-4a9742671af7.png">

PR #1149, opened 2022-05-23
Organize visualizations into app types and computations

jernestmyers:
Addresses #1095

A couple things of note in this iteration:
1. I've left the "New Visualization" button for each instance to assist with testing/visualizing the changes.
2. I've used `// @ts-ignore` liberally for ease of prototyping. I will add typeguards as I near approval for this stage (and subsequent stages), unless otherwise desired.

### DIFF
--- a/src/lib/core/components/computations/ComputationInstance.tsx
+++ b/src/lib/core/components/computations/ComputationInstance.tsx
@@ -14,6 +14,7 @@ export interface Props extends ComputationProps {
   computationId: string;
   visualizationTypes: Record<string, VisualizationType>;
   baseUrl?: string; // right now only defined when *not* using single app mode
+  singleAppMode: string | undefined;
 }
 
 export function ComputationInstance(props: Props) {
@@ -26,6 +27,7 @@ export function ComputationInstance(props: Props) {
     geoConfigs,
     visualizationTypes,
     baseUrl,
+    singleAppMode,
   } = props;
 
   const computation = useMemo(() => {
@@ -92,6 +94,7 @@ export function ComputationInstance(props: Props) {
         totalCounts={totalCounts}
         filteredCounts={filteredCounts}
         baseUrl={baseUrl}
+        showHeading={!singleAppMode}
       />
     </div>
   );
@@ -121,22 +124,18 @@ function AppTitle(props: AppTitleProps) {
     marginTop: 10,
   };
   return condensed ? (
-    <div>
-      <h3>
-        {computation.displayName} <i className="fa fa-cog"></i>{' '}
-        <i className="fa fa-clone"></i> <i className="fa fa-trash"></i>
-      </h3>
-      <h4>
-        <i>{computationAppOverview.displayName}</i>
+    <div style={{ marginTop: 10 }}>
+      <h4>{computationAppOverview.displayName}</h4>
+      <h4 style={{ marginLeft: 20 }}>
+        <em>{computation.displayName}</em>
       </h4>
     </div>
   ) : (
     <div style={expandedStyle}>
-      <h3>
-        {computation.displayName} <i className="fa fa-cog"></i>{' '}
-        <i className="fa fa-clone"></i> <i className="fa fa-trash"></i>
-      </h3>
       <h4>{computationAppOverview.displayName}</h4>
+      <h4>
+        <em>{computation.displayName}</em>
+      </h4>
     </div>
   );
 }
--- a/src/lib/core/components/computations/ComputationInstance.tsx
+++ b/src/lib/core/components/computations/ComputationInstance.tsx
@@ -36,6 +36,21 @@ export function ComputationInstance(props: Props) {
     );
   }, [computationId, analysis]);
 
+  // const groupingObject = useMemo(() => {
+  //   if (!analysis?.descriptor.computations.length) return {};
+  //   const test: any = {};
+  //   for (const computation of analysis?.descriptor.computations) {
+  //     const key  = computation.descriptor.type;
+  //     if (!(key in test)) {
+  //       test[key] = [computation]
+  //     } else {
+  //       test[key] = test[key].concat(computation)
+  //     }
+  //   }
+  //   return test;
+  // }, [analysis]);
+  // console.log(groupingObject);
+
   const toggleStarredVariable = useToggleStarredVariable(props.analysisState);
 
   const updateVisualizations = useCallback(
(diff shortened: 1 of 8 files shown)


src/lib/core/types/visualization.ts
```diff
@@ -61,8 +61,7 @@ export const ComputationConfiguration = intersection([
     collectionVariable: VariableDescriptor,
   }),
   partial({
-    alphaDivMethod: string,
-    rankingMethod: string,
+    method: string,
```
asizemore, line 64: I think this is causing 422s in the visualizations. For right now the two apps with computations need a type of method (though called different things) and a collection variable. In the future, we'll have an app with different data or even no data as input, and then 2 or more methods. So though i'd love to make it as general as a simple "method" here like you've done, unfortunately we won't be able to sustain the simplicity even for the next two apps

src/lib/core/components/computations/getAppDisplayProperties.tsx
```diff
@@ -0,0 +1,31 @@
+import { useMemo } from 'react';
+import { useStudyMetadata } from '../..';
+import { useCollectionVariables } from '../../hooks/study';
+import { Computation } from '../../types/visualization';
+
+const ABUNDANCE_METHODS = ['median', 'q3', 'variance', 'max'];
+const ALPHA_DIV_METHODS = ['shannon', 'simpson', 'evenness'];
+
+export function useAppPropertiesForDisplay(
```
asizemore, line 9: What do you think about having each plugin define it's own useAppPropertiesForDisplay or similar function in the exported plugin? The biggest obvious con to me is that would introduce duplicated code. But the big advantage i see is that as these configurations continue to diverge, we can keep any app-specific logic or verbiage within the plugin file.

src/lib/workspace/ComputationRoute.tsx
```diff
@@ -142,20 +191,36 @@ export function ComputationRoute(props: Props) {
...
-                    name,
-                    configuration,
-                    computations
+                  /* 
+                    Here we're checking for a matching configuration before instantiating a new one
+                    If there's a match, we'll store it as existingConfig and use that data in the else block
+                  */
+                  // @ts-ignore
+                  const existingConfig = computations.find((c) =>
+                    isEqual(c.descriptor.configuration, configuration)
```
asizemore, line 200: Whenever we're looking to see if an app instance already exists, we should also check to verify it has the same app type. I noticed that if i first make an XY-Relationships app instance (configuration=null) and then try to make a Counts and Proportions app instance (also configuration=null) i'm sent back to the original XY-Relationships app instance.

src/lib/core/components/computations/ComputationInstance.tsx
```diff
@@ -102,14 +108,20 @@ interface AppTitleProps {
   computation: Computation;
   computationAppOverview: ComputationAppOverview;
   condensed: boolean;
+  description:
```
asizemore, line 111: Since we expect the description to handle more than data and a single method, maybe `getAppDescription` or `useAppPropertiesForDisplay` could return a description string instead of a method and displayName?

asizemore (commented):
Looking good!

src/lib/core/components/computations/getAppDisplayProperties.tsx
jernestmyers: I didn't wire up the methods in this hook yet because I didn't see an obvious way to do so. But handling the display/description logic within the plugins themselves could make sense.

dmfalke (commented):
Based on a very cursory review, this is looking good. I replied to Ann's comment. I'm happy to discuss more, if clarification is needed.

src/lib/core/components/computations/getAppDisplayProperties.tsx
dmfalke: For instance, in [AlphaDivConfiguration](https://github.com/VEuPathDB/web-eda/blob/organize-vizs-1095/src/lib/core/components/computations/plugins/alphaDiv.tsx), we currently have an input box to allow the user to "name" their app instance. Instead, we can generate on based on the config.

jernestmyers:
Latest changes have updated the `StartPage` component to reflect the layout described in issue #1131. Screenshot below is how it looks as a result of these changes.

A couple things to note:
1. The buttons are not yet wired up to the correct config/route
2. In the absence of a `betadiv` plugin, I used a "helperPlugin" to allow me to render its scatterplot selector component.

image

jernestmyers:
Latest push has these notes/caveats:
1. The viz buttons are now wired up with the correct config/route without the logic that prevents config duplication
2. I'm pulling defaultConfigs from a newly-created hook for now, but not sure if this is a scalable approach
3. I haven't yet resolved the 422s thrown when interacting with the viz

src/lib/core/components/computations/getAppsDefaultConfigs.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return {
+      name: app.name,
+      configuration: {
+        name: defaultObject.name,
+        collectionVariable: {
+          variableId: collections[0].id,
+          entityId: collections[0].entityId,
+        },
+        [defaultObject.methodPropertyName]: defaultObject.method,
+        description: `Data: ${collections[0].entityDisplayName}: ${collections[0].displayName}; Method: ${defaultObject.method}`,
```
dmfalke, line 36: You might already know this, but I think this is the source of the 422. I think this should be moved to a `displayName` prop in the top-level object (sibling of `name` and `configuration`).

dmfalke (commented):
This is looking great. Very exciting!

src/lib/core/components/computations/getAppsDefaultConfigs.tsx
asizemore, line 36: @jernestmyers I agree with your comment about the hook not being scalable. For example one of the next two apps that we'll implement has at different parameters (data and variable for now, but will have a third one soon after).
Additionally, we'll eventually implement more complicated logic for the default collection var, and that logic would be app-specific.

asizemore (commented):
Very nice! I played around with it a bit and it works well!

jernestmyers:
I've now completed the first pass at [adding the plugin configuration to the full screen visualization](https://github.com/VEuPathDB/web-eda/issues/1132).

Note: logic in the `changeConfigHandler` will need to be extracted/refactored to be more DRY/extensible from plugin to plugin

src/lib/core/components/computations/Types.ts
```diff
@@ -20,6 +20,7 @@ export interface ComputationConfigProps extends ComputationProps {
   // alphadiv abundance
   addNewComputation: (
     name: string,
+    displayName: string,
```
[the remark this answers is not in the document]
dmfalke: Change this will require changes on the backend (I believe). I know it isn't the most clear property name--maybe we can revisit this later.

src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -24,25 +28,144 @@ function variableDescriptorToString(
...
   if (collections.length === 0)
     throw new Error('Could not find any collections for this app.');
 
-  const [collectionVariable, setCollectionVariable] = useState(
-    variableDescriptorToString({
-      variableId: collections[0].id,
-      entityId: collections[0].entityId,
-    })
-  );
+  const rankingMethod =
```
asizemore, line 45: I don't quite follow this part. From my naiive read it seems like both the getAppsDefaultConfigs hook and this section of the plugin are setting the default configuration?

asizemore (commented):
Works really well!! And the thumbnails all get sorted appropriately and everything! Really cool!

asizemore (approved):
🎉

PR #1171, opened 2022-06-03
let abundance boxplot appear on page load

### LINKED ISSUE #993: Apps: Ranked abundance box plot can appear on load
The ranked abundance boxplot gets both its x and y variables from the computation/collection variable. Consequently the boxplot could appear as soon as the visualization loads.

Pro: User gets to see un-stratified visualization. This is arguably the most important view, as it would show exactly the distribution of abundance values for the top N taxa in the data.
Con: No other visualizations work like this :/

asizemore:
Resolves #993 

An un-stratified ranked abundance viz is a very large part of what the mbio eda has to offer, and i'd love to show that part off at the demo on thursday. This PR implements a small hack to make it happen, with the intention that this will be able to be removed and all will look nicer after the refactoring :) 

Tested on the mbio backend. All three boxplots on the site work as they should. Note if i do ranked abundance boxplot, i get a plot, then add an overlay var, there's a brief time where it says "no plottable data". Looks like `outputSize` is getting set to 0, causing the error. Upon further investigation, i think when there is no stratification value set, we're getting that `data.value?.completeCasesAllVars = 0` because the boxplot response is not right. I will address this in plot.data.

Notes: Would definitely appreciate any advice on implementation! I felt squirmish about messing with `useFindOutputEntity` for this hack, but also didn't want to add a bunch of logic all over the place that would be hard to remove.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
@@ -261,23 +261,23 @@ function BoxplotViz(props: VisualizationProps) {
   );
 
   // outputEntity for OutputEntityTitle's outputEntity prop and outputEntityId at getRequestParams
-  let outputEntity = useFindOutputEntity(
-    dataElementDependencyOrder,
-    vizConfig,
-    'xAxisVariable',
-    entities
-  );
-
-  // Abundance boxplots already know their entity, x, and y vars. Set the output entity here so that
-  // the boxplot can appear on load.
-  if (computation.descriptor.type === 'abundance') {
-    outputEntity = entities.find(
-      (e) =>
-        e.id ===
-        (computation.descriptor.configuration as ComputationConfiguration)
-          ?.collectionVariable.entityId
-    );
-  }
+  // Abundance boxplots already know their entity, x, and y vars. If we're in the abundance app, set
+  // the output entity here so that the boxplot can appear on load.
+  const outputEntity =
+    useFindOutputEntity(
+      dataElementDependencyOrder,
+      vizConfig,
+      'xAxisVariable',
+      entities
+    ) ??
+    (computation.descriptor.type === 'abundance'
+      ? entities.find(
+          (e) =>
+            e.id ===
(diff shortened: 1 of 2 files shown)


dmfalke (approved):
This looks good! Generalizing this will be a fun challenge :smile:.

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -257,13 +261,24 @@ function BoxplotViz(props: VisualizationProps) {
   );
 
   // outputEntity for OutputEntityTitle's outputEntity prop and outputEntityId at getRequestParams
-  const outputEntity = useFindOutputEntity(
+  let outputEntity = useFindOutputEntity(
```
bobular, line 264: Would it be cleaner (but less readable!) with 

```
const outputEntity = computation.descriptor.type === 'abundance'
? entities.find(
      (e) =>
        e.id ===
        (computation.descriptor.configuration as ComputationConfiguration)
          ?.collectionVariable.entityId
    )
    : useFindOutputEntity(...
```

asizemore: Much cleaner! But it seems that react gets angry about violating the "rules of hooks" 😢 
```
React Hook "useFindOutputEntity" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks
```

bobular: Oh, I see...

bobular: What about
```
const outputEntity = useFindOutputEntity(...) ?? computation.descriptor.type === 'abundance'
? entities.find(
      (e) =>
        e.id ===
        (computation.descriptor.configuration as ComputationConfiguration)
          ?.collectionVariable.entityId
    ) : undefined;
```
?

asizemore: Works beautifully! Thanks @bobular ! I implemented your approach in the most recent commit

Issue #610, opened 2021-10-19
Apps: boxplot stats table


asizemore:
In mbio alpha div and abundance apps, the boxplot can return a statsTable (already in the response). Can we please 

1. Add this stats table to the display. Can be styled just like the other tables like variable coverage
2. Add a toggle to show/hide this table. By default, we let's set it to 'off', just so the user has to make a decision about if they want statistics. We want to be wary of just throwing stats at people when they're only here to explore :) 

Columns: 
- Parameter (numeric)
- Statistic (numeric)
- p-value (numeric, scientific supported here?)
- method (string)
- axis variables as needed. For example if `xVariableDetails` is in the response, we'd add a column called X-axis that displays the `xVariableDetails.value` in that row.

Example with facet:

<img width="863" alt="Screen Shot 2022-04-27 at 5 09 12 PM" src="https://user-images.githubusercontent.com/11710234/165631512-a5f4757c-00f6-40d3-9f64-cd28f6ce2531.png">

asizemore:
For reference, here's an example response for the boxplot in alpha diversity (1-row table with a column for "parameter", "statistic", "p-value", and "method") :

<img width="1779" alt="Screen Shot 2022-04-27 at 1 24 45 PM" src="https://user-images.githubusercontent.com/11710234/165584529-0c089ee5-69ea-4ca9-b552-c37920b0fd4c.png">

Example with an overlay (this would be a 2-row table with an additional column for the x variable now):
<img width="1768" alt="Screen Shot 2022-04-27 at 1 27 18 PM" src="https://user-images.githubusercontent.com/11710234/165584769-5381d4cd-728b-4990-abf8-601e3174eba5.png">

dmfalke:
Should we do this with tabs, like with mosaic tables?
![image](https://user-images.githubusercontent.com/365139/165626468-a2f7f52e-69be-45e7-b930-b2afdd326b2c.png)
![image](https://user-images.githubusercontent.com/365139/165626500-3f803b53-04b4-449f-a8dc-d633441dc999.png)

asizemore:
Oooh good question. 
Our intention was to make users ask for the stats explicitly, instead of just showing them by default. I suppose a tab supports this as well as an on/off toggle, because it's still one extra click for the user. @dpbisme and @d-callan probably have thoughts?

d-callan:
I'm fine w it, think I like the consistency across vizs.

asizemore:
[...]

PR #1175, opened 2022-06-06
mbio - boxplot stats table for alphadiv and abundance

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/610. Several tests need to be made for finding out what I might have missed.

Tab style like Mosaic is implemented. Screenshots will be shown in the following comments.  
Currently, N/A is assigned to be shown if table values are null or empty array.

### DIFF
--- a/src/lib/core/components/visualizations/Visualizations.scss
+++ b/src/lib/core/components/visualizations/Visualizations.scss
@@ -298,3 +298,21 @@ $data-table-inner-border: 1px solid #888;
     border-bottom: $data-table-inner-border;
   }
 }
+
+// boxplot stat table for alphadiv and abundance
+.BoxStatsTable {
+  @include data-table;
+  width: 750px;
+  min-width: 750px;
+
+  .percentage {
+    color: #333;
+    font-size: 0.85em;
+  }
+  th {
+    vertical-align: baseline;
+  }
+  th:first-of-type {
+    border-bottom: $data-table-inner-border;
+  }
+}
--- a/src/lib/core/api/DataClient/types.ts
+++ b/src/lib/core/api/DataClient/types.ts
@@ -521,32 +521,61 @@ const BoxplotResponseData = array(
   ])
 );
 
-export type BoxplotResponse = TypeOf<typeof BoxplotResponse>;
-export const BoxplotResponse = type({
-  boxplot: type({
-    data: BoxplotResponseData,
-    // typing computedVariableMetadata for computation apps such as alphadiv and abundance
-    config: intersection([
+// boxplot stats table
+export type BoxplotStatsTable = TypeOf<typeof BoxplotStatsTable>;
+export const BoxplotStatsTable = array(
+  partial({
+    xVariableDetails: type({
+      variableId: string,
+      entityId: string,
+      value: string,
+    }),
+    facetVariableDetails: array(
       type({
-        completeCasesAllVars: number,
-        completeCasesAxesVars: number,
-        xVariableDetails: type({
-          variableId: string,
-          entityId: string,
+        variableId: string,
+        entityId: string,
+        value: string,
+      })
+    ),
+    parameter: union([number, array(number), nullType]),
+    pvalue: union([number, array(number), nullType]),
+    statistic: union([number, array(number), nullType]),
+    method: union([string, array(string), nullType]),
+    statsError: string,
+  })
+);
+
+export type BoxplotResponse = TypeOf<typeof BoxplotResponse>;
+export const BoxplotResponse = intersection([
+  type({
+    boxplot: type({
+      data: BoxplotResponseData,
+      // typing computedVariableMetadata for computation apps such as alphadiv and abundance
+      config: intersection([
+        type({
+          completeCasesAllVars: number,
+          completeCasesAxesVars: number,
+          xVariableDetails: type({
+            variableId: string,
+            entityId: string,
+          }),
+          yVariableDetails: type({
+            variableId: string,
+            entityId: string,
+          }),
         }),
-        yVariableDetails: type({
-          variableId: string,
-          entityId: string,
+        partial({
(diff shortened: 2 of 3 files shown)


moontrip:
# Screenshots: used GEMS1 Case Control

- alphaDiv: xAxisVariable only
boxplot-alpha-X

- alphaDiv: X and Overlay variables
boxplot-alpha-X-overlay

- alphaDiv: X, Overlay, and Facet variables
boxplot-alpha-X-overlay-facet

- abundance: overlay and facet (lengthy table is cut when capturing)
boxplot-abundance-overlay-facet

moontrip:
- potential issues: example with abundance with overlay variable
boxplot-abundance-overlay-issue1

a) variable name is long with unnecessary entityID or variableID prefix ("EUPATH_0009252_"). Since corresponding plot's X label is already shortened, this may also need to be shortened.
b) question on how to handle very small number of p-value that can be treated as zero, like e-25. Do we need to change the value to be zero or just leave it as is for now?

asizemore:
@moontrip this is looking great!! I haven't had time to dive into the code yet but two quick things i can see based on the screenshots:

1. Check out the `fixVarIdLabel` function [here](https://github.com/VEuPathDB/web-eda/blob/d35e1b03b46acd4c0733d91523797d38bb3e5481/src/lib/core/utils/visualization.ts#L135-L148). In that example with the unnecessarily long var name, we should instead go find that variable's display name. The same issue was happening in the abundance boxplot, and was addressed [here](https://github.com/VEuPathDB/web-eda/pull/1083) if that is helpful!
2. Can we add a vertical scrollbar to the table for when it gets really long?

I'll hopefully be able to check out the code tomorrow!! So exciting!

moontrip:
@asizemore I have addressed your two suggestions: took some time for modifying types etc. The height of the table is set to be the same with the one in plot, 450px for now. Here is an screenshot for abundance case.

boxplot-abundance-facet

moontrip:
@asizemore I have made some changes to address bugs, especially for abundance case that now supports to display plot without selecting overlay or facet variable. Here are screenshots concerning that.

boxplot-abundance-no

boxplot-abundance-facet-only

src/lib/core/components/BoxStatsTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+                  ? overlayVariable == null
+                    ? independentAxisLabel
+                    : collectionVariableDetails && entities
+                    ? fixVarIdLabel(
+                        data.xVariableDetails?.value ?? '',
+                        collectionVariableDetails,
+                        entities
+                      )
+                    : data.xVariableDetails?.value
+                  : overlayVariable != null
```
asizemore, line 62: I'm not quite following this totally, it seems like we have
```
descriptor.type === 'abundance'
    ? overlayVar == null
        ? (...)
        : (...)
    : overlayVar != null
        ? (...)
        : (...)
```
where we are asking if overlayVar is/isn't null twice? What am i missing?

src/lib/core/components/visualizations/Visualizations.scss
```diff
@@ -298,3 +298,23 @@ $data-table-inner-border: 1px solid #888;
     border-bottom: $data-table-inner-border;
   }
 }
+
+// boxplot stat table for alphadiv and abundance
```
asizemore, line 302: We should ask clinepi if they want the stats table for their boxplot, too. Would you mind demoing at the front end meeting tomorrow morning? I'll reach out to the clinepi folks

asizemore (commented):
Looking great! Let's add this to tomorrow's meeting discussion, if you don't mind! I'd be interested to know if the team is okay merging this now where it's only going into the apps, or if they'd prefer to wait until we know more about if clinepi would also want the regular boxplot to have these stats.

asizemore:
Note: right now the regular boxplot isn't returning the statisticsTable. It's a boolean flag on the back end that we can turn on/off pretty easily if we decide to always send a statisticsTable

asizemore:
> question on how to handle very small number of p-value that can be treated as zero, like e-25. Do we need to change the value to be zero or just leave it as is for now?

Whoops sorry i never addressed this question! I think we should leave it as is for now. Sometimes researchers want that dramatic precision, so i'd rather wait and get feedback from users before altering the format or value.

src/lib/core/components/BoxStatsTable.tsx
moontrip: @asizemore thank you for your reviews! 👍 
This part is made this morning to address bug for abundance where it is now displaying plot when selecting boxplot: because its X and Y are computed variables. So, the condition starts from abundance or not. so the first condition, ? overlayVar == null is when type is abundance case, and the second condition, : overlayVar != null is general case, just not abundance. The reason why I introduced the first condition, ? overlayVar == null, is to display X axis contents for the initial plot or facet variable is selected (while no overlayVariable)

src/lib/core/components/visualizations/Visualizations.scss
moontrip: I can demo if you want, for sure 😃

src/lib/core/components/BoxStatsTable.tsx
moontrip: No problem! I was also confused about that when I made it 😄

moontrip:
@asizemore thank you for your detailed information! 👍 yes if the backend can provide a formatted data, e.g., headers, # of rows, etc., then a generic table may be possible: though, it would not be that easy task to prepare for that generic format for all table stuffs 🤞 Perhaps your info may be achieved with standardized viz refactoring

Please feel free to leave comment(s) or contact me if you need my hands 😉

asizemore:
Note - waiting for https://github.com/VEuPathDB/plot.data/issues/183 and https://github.com/VEuPathDB/web-eda/issues/1253

Issue #1178, opened 2022-06-07
Apps: thumbnail page configuration styling improvements


asizemore:
After discussion at an eda ux meeting, let's improve the styling of the app config on the thumbnail page. See below:

<img width="779" alt="Screen Shot 2022-06-07 at 10 13 41 AM" src="https://user-images.githubusercontent.com/11710234/172418427-9cb3bae2-edbc-41de-a290-2fbb6c51c7a1.png">

Figma file [here](https://www.figma.com/file/0Z6cCnTNg1yAaFGEOXQZeg/Microbiome-apps?node-id=526%3A4557)

Happy to provide more details if helpful!

asizemore:
Assigning to @dpbisme for QA

PR #1179, opened 2022-06-07
Change style of computation's configuration display header

jernestmyers:
Addresses #1178 

My implementation splits the configuration's display string at an arbitrarily chosen/inserted `&;&` string, effectively creating "data" and "method" substrings. I opted for this simple approach because my [current refactoring of the implementation details](https://github.com/VEuPathDB/web-eda/pull/1180) may affect this data.

image

### DIFF
--- a/src/lib/core/components/computations/plugins/alphaDiv.tsx
+++ b/src/lib/core/components/computations/plugins/alphaDiv.tsx
@@ -135,7 +135,7 @@ export function AlphaDivConfiguration(props: ComputationConfigProps) {
       const newComputation = createComputation(
         computation.descriptor.type,
         // @ts-ignore
-        `Data: ${variableObject?.entityDisplayName}: ${variableObject?.displayName}; Method: ${newConfigObject.alphaDivMethod}`,
+        `${variableObject?.entityDisplayName} > ${variableObject?.displayName}&;&${newConfigObject.alphaDivMethod}`,
         // @ts-ignore
         newConfigObject,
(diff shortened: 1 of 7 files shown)


src/lib/core/components/computations/plugins/alphaDiv.tsx
```diff
@@ -135,7 +135,7 @@ export function AlphaDivConfiguration(props: ComputationConfigProps) {
       const newComputation = createComputation(
         computation.descriptor.type,
         // @ts-ignore
-        `Data: ${variableObject?.entityDisplayName}: ${variableObject?.displayName}; Method: ${newConfigObject.alphaDivMethod}`,
+        `${variableObject?.entityDisplayName} > ${variableObject?.displayName}&;&${newConfigObject.alphaDivMethod}`,
```
asizemore, line 138: Lovely. It's occurred to me that we should also make the same change in the configuration dropdown. So for example changing line 225 in alphaDiv.tsx to something like
```
{collectionVar.entityDisplayName} {'>'} {collectionVar.displayName}
```

asizemore (approved):
Looks great! If you don't mind adding the '>' to the parameter dropdowns, then i think we're good to go!

src/lib/core/components/computations/plugins/alphaDiv.tsx
jernestmyers: Good catch! I'll make that change.

Issue #1164, opened 2022-06-02
Apps: remove `useDefaultPluginConfiguration`


asizemore:
Following the merge of #1149 

Reorganize so that the default config comes from the computation plugin. Consider looking at how we do default configurations with visualizations for inspiration (from @dmfalke )

SheenaTomko:
Nothing to qa

Issue #1163, opened 2022-06-02
Apps: remove `@ts-ignore` directives


asizemore:
Following the merge of #1149

SheenaTomko:
Nothing to qa

PR #1180, opened 2022-06-07
Improve mbio viz refactor implementation

jernestmyers:
Addresses #1164 and #1163.

Things done up to this point:
- plugins, as applicable, create a default configuration within the plugin.
- type guards have been added to remove all `@ts-ignore` statements
- unnecessary comments, console logs, and imports have been removed
- made adjustments to remove React or MUI warnings from console (however, a pesky React key warning remains in `StartPage`)


jernestmyers:
Apologies in advance that this became a bit of a behemoth of a PR.

src/lib/core/components/computations/ComputationInstance.tsx
```diff
@@ -105,32 +106,17 @@ interface AppTitleProps {
...
-// and one just below the navigation when we're working on a viz within an app. The former
-// is the "condensed" version. May make sense to break into two components when
-// further styling is applied?
 function AppTitle(props: AppTitleProps) {
   const { computation, condensed } = props;
-  const splitDisplayName = computation.displayName
-    ? computation.displayName.split('&;&')
-    : '';
+  const plugin = plugins[computation.descriptor?.type];
+  const ConfigDescription = plugin.configurationDescriptionComponent;
```
dmfalke, line 112: It occurs to me that this is unsafe, as `plugin` could potentially be `undefined`. I think the type of `plugins` needs to be updated as such. I think we should also throw an error here, if `plugin` is undefined. Perhaps an assert function?

src/lib/core/components/computations/StartPage.tsx
```diff
@@ -67,120 +68,117 @@ export function StartPage(props: Props) {
...
-              </Link> */}
                 {app.visualizations?.map((vizType, index) => {
                   const disabled = plugins[app.name] === undefined;
                   const VizSelector = plugins[app.name]
                     ? plugins[app.name].visualizationTypes[vizType.name]
                         .selectorComponent
                     : helperPlugin.visualizationTypes['scatterplot']
                         .selectorComponent;
-
+                  const plugin = plugins[app.name];
```
dmfalke, line 78: Need to assert `plugin` is not undefined

dmfalke (changes requested):
I think this looks great! The only change I'm requesting is to check that `plugin` is defined in the places I commented.

src/lib/core/components/computations/StartPage.tsx
jernestmyers: Apologies, that chunk of code changed with [this PR](https://github.com/VEuPathDB/web-eda/pull/1213) and I forgot to push the merged code.

src/lib/core/components/computations/Utils.ts
```diff
@@ -47,3 +50,177 @@ function createRandomString(numChars: number) {
     .toString(36)
     .slice(2, numChars + 2);
 }
+
+export function assertComputationWithConfig<ConfigType>(
+  computation: Computation,
+  decoder: t.Type<Computation, unknown, unknown>
+): asserts computation is Computation<ConfigType> {
+  const onLeft = (errors: t.Errors) => {
```
asizemore, line 58: for my education, do you mind explaining a bit about this function? I think it's goal is to assert the correct type, but how does the `onRight`, `onLeft` come into play?

src/lib/core/components/computations/Utils.ts
```diff
@@ -47,3 +50,177 @@ function createRandomString(numChars: number) {
...
+  updatedConfiguration: ConfigType,
+  visualizationId: string,
+  url: string,
+  history: RouterChildContext['router']['history']
+) {
+  // when a config value changes:
+  // 1. remove viz from current computation
+  // 2. check if the newConfig exists
+  // Y? move viz to the found computation, "existingComputation"
+  // N? create new computation
```
asizemore, line 103: Do we need 3. check if the original computation is now empty and delete if it is? From a user's perspective, all is working great, so perhaps 3. is unnecessary.

src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -1,25 +1,86 @@
...
   configurationComponent: AbundanceConfiguration,
+  configurationDescriptionComponent: AbundanceConfigDescriptionComponent,
   visualizationTypes: {
     boxplot: boxplotVisualization,
     scatterplot: scatterplotVisualization,
   },
+  createDefaultComputationSpec: createDefaultComputationSpec,
 };
 
+function AbundanceConfigDescriptionComponent({
```
asizemore, line 34: This is great. Leaves us lots of room to personalize the configuration description per app, which is going to be really helpful

src/lib/core/components/computations/plugins/alphaDiv.tsx
```diff
@@ -1,25 +1,86 @@
...
+        <span style={{ fontWeight: 300 }}>
+          {configuration.alphaDivMethod[0].toUpperCase() +
+            configuration.alphaDivMethod.slice(1)}
+        </span>
+      </h4>
+    </>
+  );
+}
+
+function createDefaultComputationSpec(rootEntity: StudyEntity) {
```
asizemore, line 71: also really nice. Will make it easy to add more complex logic when picking default configurations

asizemore (approved):
looks really great!

src/lib/core/components/computations/Utils.ts
jernestmyers: So `decode` will return either an object called `left` or an object called `right`. As you could guess, `left` represents an error. The [top-level "Idea" section](https://github.com/gcanti/io-ts/blob/master/index.md#the-idea) of the `io-ts` docs handle it this way. I think it could also be done in a more readable format with the use of `isLeft` and/or `isRight`, along these lines:
```js
if (isLeft(decoder.decode(computation)) throw new Error('bad config')
```

dmfalke (approved):

Issue #1089, opened 2022-04-19
Apps: hide no data toggle


asizemore:
Consider ranked abundance scatter

<img width="1591" alt="Screen Shot 2022-04-19 at 5 13 05 PM" src="https://user-images.githubusercontent.com/11710234/164102838-e8ba2a80-64e9-4dd9-a6b2-e7b05e02af12.png">

Does it make sense to have the no data toggle here? We could possibly end up with a "No data" group for the computed var, which is sort of odd since we're in the app in order to see the outputs of the computation.

@dpbisme and @d-callan what do you think?

asizemore:
We could have a rule that any viz with a computed var won't have the no data toggle. That's maybe aggressive because sometimes the computed var is a strata var but other times is an axis var, but as a rule is not too complicated

asizemore:
Note in plot.data we have a rule that we can't have the no data toggle on _and_ have a collection var. We'll need to replicate this rule at least in web-eda to prevent users getting lots of errors

danicahelb:
@asizemore @moontrip can we move this out of "this cycle" as it has a milestone of b59 and we are in crunch time for b58? thanks!

asizemore:
The work for this is already done, it's just in final review. It'll be simpler and faster to just approve/merge the PR then to try to hold off and have to maintain the branch. I'll finish the review today so it'll be ready for @moontrip on Monday.

asizemore:
Assigning to @dpbisme for qa

PR #1183, opened 2022-06-07
hide no data control for mbio alphaDiv and abundance

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/1089. Initially I tried to make a new prop to conditionally control the option, however, I realized that it also depended on onShowMissingnessChange function. Thus, I have made a condition to show and hide it: only for boxplot and scatterplot Vizs.

It is hidden for mbio alphaDiv and abundance, while it shows for other apps, e.g., pass app, mbio distributions, mbio X-Y relationships, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.6.1",
+  "version": "3.6.2",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
@@ -693,7 +693,12 @@ function BarplotViz(props: VisualizationProps) {
               data.value?.completeCasesAxesVars
           }
           toggleStarredVariable={toggleStarredVariable}
-          onShowMissingnessChange={onShowMissingnessChange}
+          // this can be used to show and hide no data control
+          onShowMissingnessChange={
+            computation.descriptor.type === 'pass'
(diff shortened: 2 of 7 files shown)


asizemore:
Hey @moontrip ! Lovely!! Works great!

Two notes:
1. After some discussion, mbio would like to remove the no data toggle from all visualizations in any app. So instead of the logic centering around the computation type, can it instead be a flag for app type? Then only the pass app would have the normal no data toggle. This logic avoids having to hardcode something like "if project === 'microbiome'".
2. i expect solving this issue would be simpler after the viz refactor because within each app plugin we can just say if there should or should not be a no data toggle. 

courtesy tag @dmfalke

moontrip:
@asizemore Thank you for your comments. I am not that clear of No.1 but eventually it can be achieved by viz refactoring as you mentioned in the No.2? BTW, I have slightly changed the condition in this PR so that it checks app type.

One question is whether I need to change other Viz under mbio like histogram (at distribution app), mosaic (counts and proportions app), etc. not to have that no data control as well?

asizemore:
> One question is whether I need to change other Viz under mbio like histogram (at distribution app), mosaic (counts and proportions app), etc. not to have that no data control as well?

Yes please! Seems like we should be able to do that same check of if app type is pass or not (I saw the logic change you made - looks good!)

moontrip:
@asizemore Thank you for your confirmation! 👍 I will do that

moontrip:
@asizemore done! 😃

asizemore (approved):
Tested on mbio and clinepi backends. Looks great @moontrip !

moontrip:
@asizemore thank you for your review! I will merge it later 👍

PR #1185, opened 2022-06-08
add optional subtitle to OutputEntityTitle

### LINKED ISSUE #1173: Ranked abundance: display computation information
Ranked abundance returns information about the top N taxa in the collection, but often there are details about which are returned. For example, since the app is ranking taxa by median, if only 2 taxa have median > 0, then only those two can really be ranked by their median (every other taxa ties each other). So the app returns only those 2 taxa with median > 0. However, the user doesn't know this so sometimes gets a boxplot with 10 taxa and sometimes with 2, based on the collection.

We'd like to add a spot for this type of computation information. In the live mbio site it lives in the plot title, but in the eda it may make sense for it to live somewhere else. Additionally, other apps will have similar computation information that is returned and displayed to the user, so we should ensure our solution is generalizable. 

Currently in mockup phase. Nothing to implement yet.

asizemore:
Resolves #1173 

In ranked abundance and future beta diversity, the app contains (or returns) information about the computation that is helpful for interpreting the plot. In the ranked abundance case, the computation removes all vars with rankingMethod = 0. That means sometimes a user sees a boxplot with 5 boxes, sometimes with 10, and sometimes with 1. 

For the ranked abundance app, the added subtitle won't change based on how the computation went. Though this differs from the live site, it's much less convoluted in terms of logic around what subtitle to display. [Here](https://github.com/VEuPathDB/MicrobiomeWebsite/blob/0d4e0b97bdc07677d338673a27dbe2382314861d/View/lib/R/shiny/apps/abundance_app/server.R#L210-L234) is logic currently used on the live site abundance app.

For the beta diversity app (future), the computation itself will return information worth displaying to the user. For example, it might say if the computation converged or how good of a job it did at finding the right projection. That info will be returned with the response, and when the time comes just assigned to the subtitle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.5.22",
+  "version": "3.5.23",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/src/lib/core/components/visualizations/OutputEntityTitle.tsx
+++ b/src/lib/core/components/visualizations/OutputEntityTitle.tsx
@@ -1,4 +1,5 @@
 import { makeClassNameHelper } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
+import { gray } from '@veupathdb/coreui/dist/definitions/colors';
 
 import { StudyEntity } from '../../types/study';
 import { makeEntityDisplayName } from '../../utils/study-metadata';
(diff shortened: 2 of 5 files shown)


jernestmyers (approved):
Looks to be working okay. Stylistically, it does get a bit lost but I don't have specific suggestions. I think in general the visualizations just have a lot going on, which obscures things like subtitles.

PR #1186, opened 2022-06-08
Map: basic implementation of continuous variable markers with gradient colormaps

bobular:
@d-callan is going to make some changes to the back end, but in the meantime we should get this merged and on QA because the maps are broken there, and this will fix it.


bobular:
Needs numbers sorting as NUMBERS!

moontrip (approved):
@bobular Sorry, nevermind. Somehow my local did not update your recent commit. I could confirm it worked fine

mapviz-legend-sort2

Issue #1188, opened 2022-06-08
Apps: Deleting viz thumbnail should delete computation if no more visualizations exist


jernestmyers:
When deleting a visualization in either the thumbnail view or in the full screen visualization view, the entire computation should be deleted from the `analysisState` when a computation no longer has any more visualizations. Currently, the computation remains and thus a header will display without any actual viz thumbnails.

**Before deleting the thumbnail:**
![image](https://user-images.githubusercontent.com/69446567/172676066-99c99693-6eb1-4573-9418-80efa399d87d.png)

**After deleting the thumbnail:**
![image](https://user-images.githubusercontent.com/69446567/172676134-c16820bb-f441-4411-8802-42a5efdcdfe7.png)

asizemore:
assigning to @dpbisme for qa

Issue #1187, opened 2022-06-08
Apps: Improper routing attached to fullscreen viz delete buttons


jernestmyers:
When viewing a visualization in the Full Screen component, the delete button re-routes to `{baseUrl}/visualizations/{computationId}`. This route displays only the visualizations of a computation, and does not include any information likes its configuration or app type. Also, this route does not include the "add visualization" button. Therefore, if the computation does not include any visualizations, it will display a blank page.

**Before clicking the "delete" button in the top right:**
![image](https://user-images.githubusercontent.com/69446567/172674772-a1ef7a30-3e9b-4621-b6ea-861f9f0171cc.png)

**After clicking the "delete" button:**
![image](https://user-images.githubusercontent.com/69446567/172674839-80fb29ad-bfd0-4be2-ab85-bb74c72762f6.png)

asizemore:
Assigning to @dpbisme for qa

PR #1192, opened 2022-06-10
Fix logic in delete visualization buttons for when singleAppMode is falsy

jernestmyers:
Addresses #1187 and #1188 

Specifically:
- Added some logic that will delete a computation if the last remaining visualization is deleted
- Improved readability by coercing `singleAppMode` into a boolean called `isSingleAppMode`. This coercion takes place in the `ComputationRoute` when passing it into `ComputationInstance` (which then passes it to `VisualizationsContainer`).
```js
  // lines 258 - 265 in ComputationRoute.tsx
  <ComputationInstance
      {...props}
      computationId={routeProps.match.params.id}
      computationAppOverview={app}
      visualizationTypes={plugin.visualizationTypes}
      baseUrl={`${url}/${computation?.computationId}`}
      isSingleAppMode={!!singleAppMode}
  />
```
- Added a ternary based on the `isSingleAppMode` property that will route a deletion in the `FullScreenVisualization` component to the appropriate `url`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.6.2",
+  "version": "3.6.3",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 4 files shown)


moontrip (approved):
I have checked and tested this PR and it worked well 👍

PR #1193, opened 2022-06-10
Map viz: add support for new backend response features

### LINKED ISSUE #1191: Map viz: incorporate improved back end behaviour for "overlay" data requests
The back end (`map-markers-overlay`) now does several new things
1. For categoricals with >8 values, the "best 7 + 1" will be reported for the viewport data, not total subset
2. It can accept bin range from the client, and will return a list of bin labels in numeric order, including any empty bins.

On the client side we have fully specified the overlay bin range and width (like we do for histograms) for continuous vars.

When there are >8 values for categoricals (only testable in UMSP at the moment) then we have some special logic for resetting any manually checked legend items.

### QA advice

Don't test UMSP (because it isn't public, performance is tricky, and we possibly have some issues with date variables in `map-markers-overlay`), just use **SCORE Seasonal** and **SCORE Mozambique**

So no need to test >8 value categoricals. Please just test the hell out of those two studies (they don't have many variables).  There are no date variables we can test in these studies either.

bobular:
Resolves #1191 
☝️ 
Testing advice in the linked issue.

If the vocabulary is small, use it as-is, to make sure that the colours are consistent wherever you are on the map.

If >8 values in a continuous, then of course the colours will change as you move around.  Also, as you move around, it's likely that any checkbox selections will have to be reset, so a check is made for this too (no pun intended).

Also added client-specified (from study metadata) bin specification for numeric continuous vars. This means that these stay the same as you move around the map, which we think is desired behaviour.  Only categoricals >8 values should change the legend as you move around the map.


asizemore:
hey @bobular do you mind reminding me which studies have geolocation data please? Thanks!

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -461,25 +491,42 @@ function MapViz(props: VisualizationProps) {
...
+  /**
+   * Reset checkedLegendItems to all-checked (actually none checked)
+   * if ANY of the checked items are NOT in the vocabulary
+   * OR if ALL of the checked items ARE in the vocabulary
+   *
+   * TO DO: generalise this for use in other visualizations
+   */
+  useEffect(() => {
+    if (vizConfig.checkedLegendItems == null || vocabulary == null) return;
+    // could possibly optimize by also returning if vocabulary.length <= 8 ?
```
asizemore, line 513: this makes sense to me, since we would always see all legend items if `vocabulary.length <= 8`

asizemore (approved):
Changes look good! I have a couple general map questions, now that i'm playing around with it a bunch.

1. For a categorical var pie plot, is the expected behavior that the counts do not change as i toggle on and off the legend items? For example, below in SCORE Seasonal i've toggled off all but Baseline, but the counts within each donut are the same as with all legend items on. However, this image alone suggests that all the counts that i see on the donuts come from Baseline alone, no? Maybe we've already discussed this at length so i apologize if im reopening a closed can of worms!

Screen Shot 2022-06-14 at 2 46 52 PM

2. I found the following error - is this expected? Looking on fir for the logs i see `2022-06-14 14:49:22.594 [rid:C3n8y] DEBUG RequestLogger:42 - Request end: GET /apps/pass/visualizations/map-markers-overlay 405` but otherwise i didn't see anything terribly helpful.
Screen Shot 2022-06-14 at 2 51 42 PM

asizemore:
Just went back and tried 2. again - and this time it worked? So must have been an odd fleeting error... very curious

bobular:
Ready for a quick once-over, @asizemore  - thanks for your comments!

asizemore:
Looks lovely @bobular !!

Yeah, you're right this should probably get discussed soon at any of the meetings. Although i'm probably not being helpful - I liked seeing all of the "1"s in proportion mode :)

asizemore:
conclusions from the meeting today:
- count changes only for the pie plot, not for either bar option

Some thoughts behind this decision:
When a legend item is toggled off in bar proportion mode, the y-axis stays the same and we see that the proportions are still the proportions of the total. In the pie plot case, because the colored arcs must fill the pie, it's now showing each colored arc as a proportion of the sum of all colored arcs. This is equivalent to changing the y axis in the bar plot proportion mode to be the sum of the proportions that are still there, instead of always being 1.

package.json
```diff
@@ -19,7 +19,7 @@
     "@material-ui/core": "^4.12.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
-    "@veupathdb/components": "^0.16.5",
+    "@veupathdb/components": "^0.16.6",
```
bobular, line 22: Just noting that I upgraded @veupathdb/components to bring in https://github.com/VEuPathDB/web-components/pull/366

Issue #1157, opened 2022-05-24
vizualizations need toggles to log scale axes


danicahelb:
building on the emergency ticket (VEuPathDB/web-eda#1153) to add log scale toggles to the scatter plots in time for David's PRISM SAB presentation, we want to also add log scale toggles to other viz types.

Next step is for the data viz group to discuss which axes of which vizs we need to do this for.

To do:

- [x] Change "Log Scale" to "Log scale"
- [x] Disable log toggle while in scatterplot line-fit modes (instead of current behaviour that forces the user back to "raw" mode)
- [x] disallow log scale in binned line plot 
- [x] ensure at least two tic marks on log scale axes (in `web-components`)
- [x] if a user-inputted axis range is active, and it starts at or below zero, then when switching on logscale for that axis, the zero or negative value must be reset to minPos
- [x] provide a notification when all values are negative for a particular axis (scatterplot and lineplot at least) - example to test with is Weight for age Z-score vs age group in a line plot (all the means are negative) - can also use filters on household wealth, for example
- [x] disable logScale when error bars are in force

asizemore:
At the data viz meeting, we decided to allow log axes for continuous vars when not dates or bins:
- line plot
- scatter
- gradient colormaps (when they come along)
- map markers y axis of bars

Additional decisions
-  For scatter plot, we should not allow trendlines and log scale at the same time. The log scale can always be available, but if i'm in smoothed mean mode and click log scale, the plot mode goes back to raw
- when choosing new var, should go back to linear scale
- Want to disallow log scaling if the data has already been transformed. So we should add a var attribute to flag if the var has already been log transformed
- If plotting a var with 0s or negative values and a user toggles on the log scaling, the data truncation indicators should appear to let the user know we can't plot 0 or negative values.

danicahelb:
WRT disallowing log scaling if the data has already been transformed...
Currently, if data is provided to us on a log scale, we load that variable as is and annotate `"scale" = log` (or ln, etc) in the .owl file. The database is not currently reading this information in, but should be by b60 (see: https://github.com/VEuPathDB/EdaSubsettingService/issues/64)

In the future, we will allow the user to transform data, possibly via derived variables or some other mechanism. How would the user-log transformed variables be flagged?

danicahelb:
@moontrip i moved this out of "this cycle" as it is only a nice to have and we are in crunch time for b58. thanks!

danicahelb:
[...]

PR #1194, opened 2022-06-11
adding log scale to lineplot and changing scatter plot for its behavior

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1157


moontrip:
This commit addressed two issues among others: a) adding log scale to lineplot Viz; b) "For scatter plot, we should not allow trendlines and log scale at the same time. The log scale can always be available, but if i'm in smoothed mean mode and click log scale, the plot mode goes back to raw"

PR #1201, opened 2022-06-13
Attempt to resolve Chrome map/thumbnail performance issues

bobular:
Resolves https://github.com/VEuPathDB/web-components/issues/360  (different repo - maybe move it here)
Depends on https://github.com/VEuPathDB/web-components/pull/365

This PR effectively debounces the calls to the plot thumbnailer (because the Task.run() returns a cancel function, that is returned as the useLayoutEffect cleanup function).  Using a 5000ms debounce, this seems to sort out the performance problems in Chrome (spinner and map locking up).  

We had thought about adding an extra call to the thumbnail grabber when the visualization is closed, otherwise the thumbnails could be very out of date or missing altogether.  I tried this using the second (commented out) `useLayoutEffect` in hooks/thumbnails.ts - but the asynchronous call to the screenshotter fails, presumably because the map has already gone away by this point. I think somehow we'd have to do it synchronously... ??

Reducing the "debounce" from 5000 to 2000 still seems to solve the Chrome freezing problem (even on my quite slow laptop) - and minimises the risk of having a very out-of-date screenshot. I think this would be an adequate b58 deliverable.

I'll leave this in draft (with 5000ms) for discussion

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.6.3",
+  "version": "3.6.4",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.6.6",
+  "version": "3.6.7",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


dmfalke (approved):

bobular:
Sorry about not merging before the yarn version upgrade... Should be OK now

Issue #1079, opened 2022-04-14
In line plots, choosing a Y-axis variable with only one value throws an error message


SheenaTomko:
The example below comes from PRISM without any filtering. I believe this error message is coming because choosing Country on the Y-axis should enable proportions, but there is only one value for the variable Country in this study. Some possible solutions: 1) remove variables with single values as options in the drop down menu, 2) give people a more distinct error message, 3) update how proportions work when there is only one possible value.

![image](https://user-images.githubusercontent.com/54597319/163398028-b31185d1-6051-4f28-b662-adcd082686fc.png)

bobular:
I think this will be obviated by https://github.com/VEuPathDB/EdaDataService/issues/178

asizemore:
In the data viz meeting, decided to allow vars with only one unique value in proportions. Also decided to allow the numerator=denominator.

danicahelb:
I'm removing the dependency label as we have decided to NOT impose the constraints indicated in https://github.com/VEuPathDB/EdaDataService/issues/178

moontrip:
@danicahelb I think that I have addressed this in the following PR. I made a couple of screenshots in there, so can you please see if it is appropriate outcome? https://github.com/VEuPathDB/web-eda/pull/1204#issue-1269899970

sweverschulman:
I have confirmed that the error message is removed, I was able to configure the proportion and that the y =1.

It was a little confusing that I can't click on mean and median - but that can be discussed later.

PR #1204, opened 2022-06-13
addressing a couple of lineplot viz stuffs

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1079, which the required tasks are summarized at https://github.com/VEuPathDB/web-eda/issues/1079#issuecomment-1152401516.

Here are example screenshots with PRISM data

- ### before applying this PR: the first screenshot is without removing the error message (taken from the ticket) and the second one is by simply removing the error message in the code, which causes 500 error instead 

a) 
163398028-b31185d1-6051-4f28-b662-adcd082686fc

b)
lineplot-error-msg1

- ### after applying this PR: now proportion can be used without error message & numerator = denominator & showing 1
lineplot-temporary error1

- ### after applying this PR: other variable for test the case of numerator = denominator & showing 1
lineplot-numerator-denominator1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.6.7",
+  "version": "3.6.8",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -437,10 +437,6 @@ function LineplotViz(props: VisualizationProps) {
       if (categoricalMode && !valuesAreSpecified) return undefined;
 
       if (categoricalMode && valuesAreSpecified) {
-        if (isEqual(vizConfig.numeratorValues, vizConfig.denominatorValues))
-          throw new Error(
-            'Numerator and denominator value(s) cannot be the same. Numerator values must be a subset of the denominator values.'
-          );
         if (
           vizConfig.numeratorValues != null &&
           !vizConfig.numeratorValues.every((value) =>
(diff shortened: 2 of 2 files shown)


moontrip:
@bobular Since you are the most familiar with the Lineplot Viz, I chose you as a reviewer 😄 Can you please check if this PR satisfies the requirement summarized by Danica, https://github.com/VEuPathDB/web-eda/issues/1079#issuecomment-1152401516? Above screenshots may be helpful too, I suppose 🤔

danicahelb:
hi @moontrip this is what we want, thanks.

can you confirm that -- if you have subsetted the data in a way that the numerator and denominator ends up being the same -- that the plot is not throwing an error and we still get a line at y=1?

For example, in GEMS1 case control, apply a filter for Country = Bangladesh
In the line plot, you can plot x=sleeping rooms in dwelling count & y=country. then configure the proportion so that the numerator is Bangladesh & India and denominator is all countries:

image

moontrip:
@danicahelb I have tested the case and unfortunately, it failed: if I manually adjust the numerator and denominator as Bangladesh, then it worked. 

I have briefly looked at the code and it looks to me that the list of numerator and denominator does not respect the filter. So, I think that if we limit the list by considering filters, then it would work, I suppose. I will try and see if I can do that myself, as the proportion was made by Bob

danicahelb:
hey @moontrip the other issue I identified has its own ticket: https://github.com/orgs/VEuPathDB/projects/11/views/1

maybe check in with @asizemore about implementing?

moontrip:
@asizemore Thank you for your explanation and sounds great to me 👍 I didn't know the presence of the backend work you were doing, so I thought it might have been a frontend stuff. I will see if I can make it from my side as well, though it's a bit tricky. 

When your work is merged, please let me know so that I can test it (with GEMS1 Danica tested) too!

bobular:
> I have briefly looked at the code and it looks to me that the list of numerator and denominator does not respect the filter. So, I think that if we limit the list by considering filters, then it would work, I suppose. I will try and see if I can do that myself, as the proportion was made by Bob

Hi @moontrip 
I don't think it will be easy to customize the numerator and denominator choices based on the current filters. Well, it might be possible with a whole new back end request, but I don't think we want to add new complexity like that at the current late stage in release 58!

PR #1208, opened 2022-06-15
respecting filters for lineplot viz proportion

moontrip:
This draft PR is a trial to address https://github.com/VEuPathDB/web-eda/pull/1204#issuecomment-1155393275 at frontend side. Certainly, there is a backend work to address the issue (https://github.com/VEuPathDB/plot.data/pull/177), but perhaps this PR would also be helpful, if working fine. So, it would be great if you can test this PR when you are available.

The basic idea of this work is to respect filters for numerator and denominator values of Lineplot Viz's proportion option concerning categorical variables. To this end, check variableId of filters (array) and selected yAxisVariable, and then make an array which will be used as checkbox list at both numerator and denominator.

As I made a comment in the code, ideally, valuePicker component will need to be changed to behave like Legend, e.g., disabling checkbox (grayed out) per filters, but it is not that simple task, especially with upcoming release, so skipped it for now.

I will show some screenshots below.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -885,7 +885,7 @@ function LineplotViz(props: VisualizationProps) {
 
   /**
    * changed allowedValues to respect filters
-   * ideally, valuePicker component needs be changed to behave like Legend, e.g., disabling checkbox (grayed out) per filter
+   * ideally, valuePicker component will need to be changed to behave like Legend, e.g., disabling checkbox (grayed out) per filter
    * but it is not that simple task, especially with upcoming release, so skipped it for now
    */
   const proportionInputs = (
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -298,19 +298,29 @@ function LineplotViz(props: VisualizationProps) {
   );
 
   // update numeratorValues and denominatorValues depending on filters and filteredProportionArray
-  useEffect(
-    () =>
-      updateVizConfig({
-        denominatorValues:
-          yAxisVariable != null
-            ? filteredProportionArray != null &&
-              filteredProportionArray.length > 0
-              ? filteredProportionArray
-              : yAxisVariable?.vocabulary
-            : undefined,
-      }),
-    [filters, filteredProportionArray]
-  );
+  useEffect(() => {
+    updateVizConfig({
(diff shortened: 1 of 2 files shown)


moontrip:
This example is based on GEMS1 study. Let's assume that Country's Bangladesh and India are subsetted (filtered): certainly there may be additional filters for other variable but here I just used Country as yAxisVariable.

### - when selecting xAxisVariable: Sleeping rooms in dwelling count; yAxisVariable: Country - as you can see the denominator only shows the two countries.

lineplot-filtered-proportion1

### - clicked numerator to see the list of checkbox
lineplot-filtered-proportion2

### - if I selected Bangladesh as a numerator only, it plots something 😄 
lineplot-filtered-proportion4

### - if I selected both Bangladesh and India as a numerator, it plots 1
lineplot-filtered-proportion3

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -321,8 +349,14 @@ function LineplotViz(props: VisualizationProps) {
           ? {}
           : {
               numeratorValues: undefined,
+              // change denominatorValues following filteredProportionArray
               denominatorValues:
-                yAxisVar != null ? yAxisVar.vocabulary : undefined,
+                yAxisVar != null
+                  ? filteredProportionArray != null &&
```
bobular, line 355: I think `filteredProportionArray` needs to be a dependency of the useCallback

bobular (commented):
Hi DK,

Interesting... Using the filter data directly allows us to restrict the options for *directly filtered* values.  If we "accidentally" filter out a country because we filtered on "Household.has_TV=yes" when a country has no TV owners, this approach won't know.  Something to be discussed more widely at an EDA UX meeting, I think.  The code looks solid, it's just a question of "do we want to do this?"

What is the back end work we speak of?

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: In my opinion, this is a sort of nice to have, not must have, because Ann's backend work would make it happen without throwing error at least.

moontrip:
A new commit is made to address missing dependency reported by @bobular and edge cases that were found by me. The cases are, e.g., assuming that Country's Bangladesh and India are subsetted, making Viz with yAxisVariable = Country and select Bangladesh as numerator value. Then moving to Browser and Subsetting and remove subset Bangladesh. Without this PR, the numerator values are still mixed with Bangladesh and India. A correct behavior is to regenerate the list of numerator values so that it does not misbehave

PR #1241, opened 2022-06-30
Viz log range truncation

moontrip:
Adresses #1157 and more

This PR relies on https://github.com/VEuPathDB/web-components/pull/371.
This is branched out from https://github.com/VEuPathDB/web-eda/pull/1194 to address two things for scatter and line plot Vizs:

a) set log scale range correctly
b) truncation

This has become a MASSIVE PR...

Some hints for reviewers:

1. Various `useDefaultXXXRange` hooks have been consolidated into one `useDefaultAxisRange` function
2. To achieve this, it is now up to the individual viz to calculate min/max (and minPos) where appropriate and send these to the new hook. Some vizs were doing this already, others were passing data through to many different hooks and functions, so it is now simplified a bit. Some of the original functions are preserved and are in src/lib/core/utils/axis-range-calculations.ts 
3. minPos is required when logscale is used, because the truncation bar calculations need to know the axis range explicitly (previously, many ranges were being passed through to the PlotlyPlot components as `[NaN, 123.45]`)
4. log-based truncation yellow bars are now the same size as non-log truncations
5. all-negative values for lineplot and scatterplot axes get a new `<Notification>` warning in the plot controls (search for "AllNegative")
6. truncation-config-utils had two versions: viz and subsetting, now unified
7. Instead of using 4 decimal places for axis range values, we now use 4 significant figures. Previously some problems were occurring when the range.min was being rounded down to 0 (from 0.00001, for example) - logs don't like zeroes!
8. Log axis handling has been made consistent for HistogramFilter (subsetting) too
9. mutually exclusive combinations of plot controls have been simplified a bit by making sure "illegal" combinations can never be reached.
10. When switching to or from log-scale, axis ranges are reset - it's a bit heavy handed, and we could check to see if any offending values are there (e.g. range.min <= 0 when going to log scale) but the current solution is by FAR the simplest, implementation-wise.

OK, I think that's enough for now.  I'll merge in the latest `main` and let you all know it's ready for review.

I left one remaining task for SVG Wizard, @moontrip : https://github.com/VEuPathDB/web-components/issues/380


asizemore:
Can we please disallow the log scale toggle when the x axis is categorical? Below i'm in GEMS1A Case Control and i can put Age Group (categorical) on the x-axis, then toggle on the log scale.

Screen Shot 2022-06-30 at 5 19 17 PM

asizemore:
Additionally, we should disallow log scale on the x axis of line plots if the x axis is binned

moontrip:
@asizemore Thank you for your suggestions! 👍 I will reflect them

bobular:
Hi @asizemore @moontrip 

I'm kind-of back! Still not 100% however.

Do we want to show yellow truncation bars if the line plot's error bars are chopped off, like in the PRISM ICEMR Cohort example below?

image

asizemore:
Happy to see you're at least a little back @bobular ! 

Hmm good question. I'd say yes we should show the yellow truncation bars if the error bars are chopped off, because the error bars are part of the requested data for the plot.

bobular (commented):
Hi DK,
I'm going to leave this for today and let you have my comments so far.
(see also the web-components PR for some map-related comments)

I am currently wondering why LineplotVisualization doesn't use `useDefaultDependentAxisRange` ?  Is that my fault?  Probably is!  I can see some duplication regarding `smallestPositiveDependentAxisValue`.

bobular:
Hi @asizemore 

I did a little bit more testing (GEMS1)

There's no warning when negative values (e.g. Household wealth numerical) disappear in log mode - is that the plan?

image

asizemore:
Hmmm good catch @bobular . If there are negative values and we turn on the log scale, we decided we should see the data truncation indicators

bobular:
I'm going to work on this today.

bobular:
> Hmmm good catch @bobular . If there are negative values and we turn on the log scale, we decided we should see the data truncation indicators

It seems that this is working fine for the dependent axis, but not the independent axis at the moment. I'm looking into it.

bobular:
Suggestions from UX - change wording from negative or zero to zero or negative

Make ticket for stripy truncation markers

asizemore:
the demo today was awesome!! I plan to dig into the code tomorrow!

moontrip:
@bobular Like https://github.com/VEuPathDB/web-components/pull/371, thank you for taking over this PR, and fixing bugs & adding features! 👍 I have looked at your changes and tested, and they looked good to me! As you noticed, I can't review this myself but I approve this PR anyway 😄

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -543,66 +568,48 @@ function LineplotViz(props: VisualizationProps) {
...
+        : undefined,
+    [data]
+  );
+
+  const defaultDependentAxisRange = useDefaultAxisRange(
+    yAxisVariable,
+    data.value?.yMin,
+    data.value?.yMinPos,
+    data.value?.yMax,
+    vizConfig.dependentAxisLogScale
```
asizemore, line 611: line plot dependent axes are marked by means and medians. For the same reason we disallow log scale on boxplot y and when errorbars are on, should we not allow log scale for y axes on line plot y axis at all?

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -1194,8 +1237,28 @@ function LineplotWithControls({
...
+    xMinMaxDataRange.max < 0;
+
+  const [
+    dismissedDependentAllNegativeWarning,
+    setDismissedDependentAllNegativeWarning,
+  ] = useState<boolean>(false);
+  const dependentAllNegative =
+    vizConfig.dependentAxisLogScale &&
+    yMinMaxDataRange?.max != null &&
+    yMinMaxDataRange.max < 0;
```
asizemore, line 1260: less than or equal to 0 here? If the max is 0 we're still in a bad place with the log scale

src/lib/core/utils/default-axis-range.ts
```diff
@@ -4,51 +4,54 @@ import { NumberOrDateRange } from '@veupathdb/components/lib/types/general';
...
-  // make universal range variable
-  if (
-    variable != null &&
-    (plotName === 'scatterplot' || plotName === 'lineplot')
-  ) {
-    // this should check integer as well
+  if (Variable.is(variable)) {
     if (variable.type === 'number' || variable.type === 'integer') {
       const defaults = variable.distributionDefaults;
+      if (logScale && observedMinPos == null) return undefined; // return nothing - there will be no plottable data anyway
```
asizemore, line 19: i don't follow - i would have expected `if (logScale && observedMax <= 0)`. Why if the minPos is null then we'd have no plottable data?

asizemore: oh wait i get it. Assume only one point is positive. That point would be the observedMinPos. 
So if there are no minimum positive values there must be no positive values. 👍

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -1227,13 +1262,34 @@ function ScatterplotWithControls({
...
-      truncationConfig(defaultUIState, vizConfig, defaultDependentAxisRange),
+      truncationConfig(
+        {
+          independentAxisRange: xMinMaxDataRange,
+          dependentAxisRange: yMinMaxDataRange,
+        },
+        vizConfig,
+        {
+          // truncation overrides for the axis minima for log scale
+          // only pass values that you want to take effect!
```
asizemore, line 1273: what do you mean by "take effect" here?

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -1383,23 +1457,41 @@ function ScatterplotWithControls({
...
+              disabled={independentValueType === 'date' || valueSpec != 'Raw'}
             />
           </div>
+          {independentAllNegative && !dismissedIndependentAllNegativeWarning ? (
+            <Notification
+              title={''}
+              text={
+                'Nothing can be plotted with log scale because all values are negative or zero'
+              }
+              color={'#5586BE'}
```
asizemore, line 1476: should probably get the warning color from CoreUI. Can be left for another ticket though!

asizemore (approved):
Looks great @bobular ! 
Just a few questions from me but nothing that should prevent merging.

src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -298,21 +295,35 @@ export function HistogramFilter(props: Props) {
   // stats from foreground
   const fgSummaryStats = data?.value?.series[1].summary;
 
+  const minPosVal = useMemo(
```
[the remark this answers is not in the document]
bobular: Yes the undefined value is caught later.

src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
```diff
@@ -331,7 +332,9 @@ function BoxplotViz(props: VisualizationProps<Options>) {
             : [],
           showMissingness: vizConfig.showMissingness ? 'TRUE' : 'FALSE',
         },
-        computeConfig: computation.descriptor.configuration,
+        ...(computation.descriptor.configuration
```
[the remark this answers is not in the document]
bobular: Oops this wasn't supposed to stay in the PR. Thanks!

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
bobular: I think your hunch is right, at least for the mean, which is calculated in regular arithmetic space.  The median is based on ranking, so will be the same in log and regular space.  I think we leave this for now.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -543,66 +568,48 @@ function LineplotViz(props: VisualizationProps) {
...
+      // This useBinning edge case handling is a bit of kludge.
+      // The same situation affects the histogram independent axis (all the time).
+      // The problem is if you make a data request for an x-axis viewport (aka zoom to range), then
+      // the xMin and xMax are calculated from the response, which is just for the requested
+      // viewport.  Using the data-derived xMin and xMax works great for client-side zooming,
+      // because the data doesn't change. But for server-side zooming we can't use xMin and xMax from
+      // the response.
+      // Two possible solutions I can think of:
+      // 1. Client makes a request for unzoomed data to get the xMin and xMax, and remembers this.
+      // 2. Server returns the truncation flags for us, somehow.
```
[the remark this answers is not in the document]
bobular: I am too. But I looked into it and it's too big/risky to put in for this release.  So I made a... guess what... [ticket](https://github.com/VEuPathDB/web-eda/issues/1295)!

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
bobular: Yes, good catch!  I won't rename the variable though.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
bobular: I just mean that if you passed `{ truncationConfigIndependentAxisMin }` as the override object, then it would override if the value of `truncationConfigIndependentAxisMin` was true or false.  I've hopefully improved the comments.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
bobular: Yes, I'm happy to search for them with some regexp magic when we do this (rather than make a ticket for this specifically)

PR #1242, opened 2022-07-05
Render VariableCoverageTable only when completeCases is truthy

### LINKED ISSUE #1072: Hiding of variable completeness table
As mentioned in https://github.com/VEuPathDB/web-eda/issues/924, hide the variable completeness table until it is populated.

Current behavior is that if NO variables are selected, the table is hidden:
image

However, if 1 variable is selected, the table appears but is not really populated yet:
image

We should hide the table until all required selections are made & the plotted data appears, for example, hide until this point:
image

jernestmyers:
Resolves #1072

Rendering when `completeCases` is defined ensures that the `VariableCoverageTable` and the `BirdsEyeView` are both now hidden until all required data is provided.

X and Y vars selected:
image

Proportion set:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.6.18",
+  "version": "3.6.19",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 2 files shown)


bobular (approved):
Looks and works great

Issue #1177, opened 2022-06-06
Viz tab: add * next to labels for all required fields


danicahelb:
Add an asterisks next to the label for all required fields in all viz tools:

Main* (histograms / bar plots / maps)
X-axis* (scatter / line / mosaics/ box plots)
Y-axis* (scatter / line / mosaics/ box plots)
Function* (line plots only, to be implemented in https://github.com/VEuPathDB/web-eda/issues/1038)
Proportion* (line plots only, to be implemented in https://github.com/VEuPathDB/web-eda/issues/1038)

For example:

<img width="875" alt="image" src="https://user-images.githubusercontent.com/25450900/172232177-26114a47-9139-442e-8684-06eb861439b3.png">

danicahelb:
Hey @jernestmyers we discussed this in the 6/6/2022 EDA UX meeting. We decided to go with black asterisks... it will be more subtle but that is ok. The team preferred visual simplicity.

Thanks!

jernestmyers:
Looks good on the feature mbio backend! Screenshot below is from `Bangladesh 5yr > Alpha Div > Box plot` but I also checked other config/viz combos and they all checked out.

![image](https://user-images.githubusercontent.com/69446567/177813524-f0c3433b-602d-4332-9ad3-ce8f6bf79dbe.png)

asizemore:
Great! Thank you for checking @jernestmyers ! And with that confidence i have now deployed the same data service fix to qa so it should be on clinepi shortly if not already.

asizemore:
@jernestmyers just a heads up - the data service upgrade broke a few other things on mbio so i reverted to an older tag. The data service upgrades you need are still on clinepi qa though!

nkittur-uga:
Looks good. Required components are in red with (*) and turn black once selected.

Issue #1038, opened 2022-04-11
Update UI for line plot viz tool, including addition of geometric mean option


SheenaTomko:
Came up in the Uganda meetings. Geometric mean can be more useful for looking at variables with exponential growth, like parasite densities. We'd like to add geometric mean as an option for aggregating and plotting continuous data on the Y-axis of a line plot

d-callan:
i have a number of questions here about how wed like to actually find this value w possible implications to the ux around doing so.  For ex, the presence of 0, negative or % values will be trouble for various reasons. So i guess my questions are:

1. Do we know if we definitely do or dont have percent values for variables? If we do, are there too many to annotate and potentially constrain against? Should we risk allowing users to calculate geometric means on percents directly?
2. Should we consider a client-side limitation that geometric mean only become available if the variable metadata range min is > 0. (I imagine some day w derived variables this would mean users could transform anything they liked to an appropriate range, but they wouldnt be able to easily 'transform' the geometric means in vizs back to something understandable directly relative to the original data range. i personally think thats probably ok particularly given this seems otherwise the safest route.)
3. Should we try to come up w a reasonable rule-of-thumb for how to find the geometric mean in cases where the min value is <= 0 and hope statisticians dont get too mad at us for potentially doing things differently to what theyd have done or for not handling some edge case well or something?

d-callan:
1. good to know there are relatively few percents.. is that 55 the number of IRIs that are known to represent percents? or the number of times we see one of those IRIs used in a study? If we constrain to only variables w >0 or >=0 values, do we know what proportion of those terms represent percents? (it seems still kind of bad to me if a user toggles to only compatible variables for some study and a large number of those were percents, but its probably unlikely?)
2. i guess theres a question whether it should limit against 0 arising vs just negative values? I can probably imagine scenarios where treating 0 as empty would be bad as well.. and it seems a thing only the user would know whether they wanted to do based on context. This was my original concern w point 3 really. if we make that decision, what are the chances itd be confusing or frustrating to people in the general case? id be personally inclined i think to answering the questions here around negative and 0 values consistently at least, even if we cant identify percents to handle those consistently as well.
3. If we impose a limitation i think we should never see an error? user just wont be able to select geometric mean for variables w incompatible values. (though admittedly that limitation would be based on the entire study rather than the subset, so you could never filter away values until you had only compatible ones remaining)

d-callan:
depends on veupathdb/plot.data#162

[...]

PR #1245, opened 2022-07-07
Render asterisks for required inputs' labels

jernestmyers:
Partially addresses #1177.

Currently only renders the asterisks next to inputs that are classified as `isRequired` in the `constraints` prop. This does not include the function or proportion inputs. The asterisks will be added to the function and proportion inputs as part of the work outlined in issue #1038.


bobular (commented):
So sorry for the long turnaround time, Jeremy.

I love the logic/implementation, but from the user's point of view they will need to know what the asterisk means. Either with a legend/footnote somewhere, or a mouse-over.

We should also consider removing the "(optional)" from the Stratification variables section.

Is this still going through some UX decision process? @danicahelb

jernestmyers:
Great points. I agree that we have no need for "(optional)" now that we're indicating what is required, thereby implying all other inputs are optional. I'll wait to hear what Danica thinks both about excising "optional" and about providing the user information about the asterisk.

Thanks for the feedback.

jernestmyers:
I just pushed the following changes:
- Removed "(optional)" from `Stratification variables (optional)` label
- Replaced the `Y-axis aggregation`'s tooltip with the `HelpIcon` from `web-components`; this serves two purposes: fixes the slow render of the tooltip (was using the default 1 second hover before rendering) and moves toward unifying help icons to look/feel the same throughout the site
- Also replaced the `BirdsEyeView`'s tooltip with the `HelpIcon`
- Added a tooltip to the axis labels that reads `Required input`

danicahelb:
@jernestmyers please see comments in https://github.com/VEuPathDB/EdaNewIssues/issues/407

jernestmyers:
Addresses #1322

dmfalke (approved):
This all looks good to me!

PR #1260, opened 2022-07-20
Add tooltip to BirdsEyeView and adjust styling

### LINKED ISSUE #994: Standardise birds eye plot mouse-over with (?) icon and it should not be active when plot is hidden

jernestmyers:
Resolves #994 in conjunction with [this PR from `web-components`](https://github.com/VEuPathDB/web-components/pull/377).

This PR adds a tooltip to the top right corner of the `BirdsEyeView` and adds a border and background color to match the `VarCompletenessTable`.

In doing this work, I noticed that `PlotLegend` was being passed into `PlotLayout` as a fragment even if no overlay variable was selected. This causes an unnecessary top margin on the `BirdsEyeView` which means it's not flush to the top of the plotted visualization. I added ternaries in the visualization implementations to prevent the behavior.

**Shows the extra space that renders above the BirdsEyeView**
image

**Shows updated version without the extra space**
image


dmfalke:
I don't think this is quite there. I'd actually prefer a white background, more padding, and box-shadow (similar to the plot). These thoughts are inspired by [this issue](https://github.com/VEuPathDB/web-eda/issues/1270).

**Current:**
image

**Proposed (roughly):**
image

jernestmyers:
To that end, I suppose we should wait until the UX meeting to reach a final verdict. I don't have strong opinions either way.

dmfalke (approved):

Issue #1176, opened 2022-06-06
Enable column sorting on subsetting results tables


danicahelb:
It would be great if we could allow users to sort the subsetting results tables by one of the columns, and allow both ascending and descending ordering. Thanks!

dmfalke:
FYI, the service already supports this: https://veupathdb.github.io/EdaSubsettingService/api.html#resources:/studies/{study-id}/entities/{entity-id}/tabular:post (see `reportConfig.sorting`)

SAK22:
Looks good!

PR #1261, opened 2022-07-21
Enable column sorting for subset download modal

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1176
Requires https://github.com/VEuPathDB/CoreUI/pull/93


src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
```diff
@@ -420,7 +427,12 @@ export default function SubsetDownloadModal({
                       },
                     },
                   },
+                  icons: {
+                    inactiveColor: gray[300],
```
asizemore, line 431: because of the background color on the header cells, the inactive color would need to be at least gray[500] to pass accessibility for icons. So perhaps gray[500] for inactive and gray[900] for active?

src/lib/workspace/Subsetting/hooks.ts
```diff
@@ -58,6 +59,7 @@ export const processGridData = (
       return {
         Header: `${entity.displayName} ID`,
         accessor: `${entity.id}/${columnID}`,
+        disableSortBy: true,
```
asizemore, line 62: curious who uses disableSortBy? Clearly someone does since it works lol i just couldn't find it.

asizemore (commented):
Interested in your opinion on the grays, but otherwise looks great!

src/lib/workspace/Subsetting/hooks.ts
chowington: Good question! It's passed through to the `DataGrid` prop `columns`, whose type is an array of `Column`s, a type that's imported straight from `react-table` in CoreUI's `DataGrid/index.tsx`. I realize this means the full API is opaque to the CoreUI user, so I added a link to the `react-table` docs in the `columns` prop description.

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
chowington: Thanks for checking on it! Unfortunately on my screen I think 500 and 900 are not distinguishable enough to determine active vs inactive easily. Can we settle with 400 and 900?

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
```diff
@@ -220,6 +225,14 @@ export default function SubsetDownloadModal({
     ({ pageSize, pageIndex, sortBy }) => {
       if (!currentEntity) return;
       setDataLoading(true);
+      setSortBy(sortBy ?? []);
+
+      const selectedVariableIds = selectedVariableDescriptorsWithMergeKeys.map(
+        (descriptor) => descriptor.variableId
+      );
+      const filteredSortBy = sortBy?.filter((column) =>
```
asizemore, line 233: am i reading this right that this line is basically a sanity check to ensure that everything we're about to sort by actually is in the variables we selected in the table? 
Was there a way to sort by a var that wasn't selected? Or was this just a cautionary add?

asizemore (approved):
👍 

> Unfortunately on my screen I think 500 and 900 are not distinguishable enough to determine active vs inactive easily. Can we settle with 400 and 900?

I thought about this for longer than i should admit 😅 . I even tried removing the gray background in the header cell:
Screen Shot 2022-07-26 at 5 33 07 PM
but that didn't look great.

Still, i realized that the icon's inactive state is actually the least of the accessibility issues lol. Specifically, our text color for the data itself doesn't pass the contrast guideline haha. I don't think we should try to address all of that now so i'll make an issue in CoreUI so we don't forget to revisit soon!
All that to say, yes let's go with your suggestion of 400 and 900!

Thanks @chowington !

src/lib/workspace/Subsetting/hooks.ts
asizemore: ahh that's great! Thank you!

chowington:
> 👍
> 
> > Unfortunately on my screen I think 500 and 900 are not distinguishable enough to determine active vs inactive easily. Can we settle with 400 and 900?
> 
> I thought about this for longer than i should admit 😅 . I even tried removing the gray background in the header cell: Screen Shot 2022-07-26 at 5 33 07 PM but that didn't look great.
> 
> Still, i realized that the icon's inactive state is actually the least of the accessibility issues lol. Specifically, our text color for the data itself doesn't pass the contrast guideline haha. I don't think we should try to address all of that now so i'll make an issue in CoreUI so we don't forget to revisit soon! All that to say, yes let's go with your suggestion of 400 and 900!
> 
> Thanks @chowington !

Thanks for putting thought into this, Ann! I agree we should rethink those design details.

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
chowington: Not quite a sanity check---I found a bug caused by the fact that `sortBy`'s state was lagging behind the currently selected variables when we removed a column with the X button. This meant in our request we were asking the backend to sort by a column that we didn't ask it to give us, so it was understandably confused about this and gave us a 500 instead. :)

I didn't see a good way to fix the inconsistency due to the logic flow, but paring down `sortBy` to have only selected variables works just as well.

PR #1262, opened 2022-07-21
Use undefined for empty compute config

dmfalke:
This PR fixes an issue where a `null` value was being passed as `computeConfig` to the "pass" app. By using the value `undefined`, the property will not be included in the payload, which will address the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.7.1",
+  "version": "3.7.2",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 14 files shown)


jtlong3rd (approved):

bobular (commented):
I tested and this still leaves old saved analyses broken for scatterplot and boxplot visualizations.

I think we need to address this.

dmfalke:
> I tested and this still leaves old saved analyses broken for scatterplot and boxplot visualizations.
> 
> 
> 
> I think we need to address this.

You're right. I think I have a different approach that will work. I'll update soon.

bobular:
> > I tested and this still leaves old saved analyses broken for scatterplot and boxplot visualizations.
> > I think we need to address this.
> 
> You're right. I think I have a different approach that will work. I'll update soon.

I gave it 5 mins of thought and figured that `<FullScreenVisualization>` could modify the `computation.configuration` from null to undefined before passing it to the `<vizPlugin.fullscreenComponent>`

dmfalke:
I pushed some changes to address analyses with `null` compute configs.

jtlong3rd (approved):

PR #1263, opened 2022-07-22
dont startCase if string includes rRNA

### LINKED ISSUE #1257: Mbio downloads: extra spaces and capitalization getting added
Only observed in the downloads tab for 16S assay entities.

The entity name is "16S rRNA (V4) assays". Curious why some of the text is getting transformed to uppercase (the r in rRNA) and some extra spaces are appearing?

Screen Shot 2022-07-13 at 4 49 11 PM

asizemore:
Resolves #1257 

We're using [startCase](https://lodash.com/docs/) to transform the entity display names in the downloads buttons. Sometimes we have biological terms (such as rRNA) that we'll never want to be startCase-ed. In order to handle these terms we could either
1. Have some complicated logic that only startCases the non-biological words
2. Just don't startCase the whole entity name (implemented now until we have a decision on the real question... see below)

Which brings me to the real question - was startCase-ing in this case an intentional decision we made a while back? Maybe because buttons should always be startCase-ed? Or do we want the button text to match the case of the entity diagram (sentence case)?
@danicahelb @SheenaTomko maybe you remember?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.7.3",
+  "version": "3.7.4",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/src/lib/workspace/DownloadTab/MySubset.tsx
+++ b/src/lib/workspace/DownloadTab/MySubset.tsx
@@ -76,9 +76,7 @@ export default function MySubset({
         <FloatingButton
           key={index}
           text={`${data.filteredCount?.toLocaleString()} of ${data.totalCount?.toLocaleString()} ${
-            data.displayNamePlural?.includes('rRNA')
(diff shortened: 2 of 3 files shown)


SheenaTomko:
I don't remember if this was intentional or not. I would be ok with it matching the entity diagram (sentence case)

asizemore:
Interestingly, google and apple are on different sides of the text case [debate](https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98)

dmfalke:
I don't think the client code should ever manipulate display text. This makes it more difficult to be consistent.

asizemore:
@jernestmyers and @dmfalke  - i think only one of you is needed for review of this tiny PR 😄

jernestmyers (approved):

dmfalke (approved):

PR #1282, opened 2022-07-26
computed and collection vars in coverage table working for boxplot

### LINKED ISSUE #1086: Apps: Allow computed vars to appear in variable coverage table

asizemore:
Resolves #1086 

This PR incorporates both computed and collection variables into the variable coverage table. In the former case, it's quite straightforward to add any computed var. In the latter case, it's more complicated because we're adding more rows than the current table expects. 

# To Dos

- [x] Have all variables and appropriate numbers show up in the variable coverage table for boxplot
- [x] Use display labels for vars in a collection var
- [x] Repeat for scatter

# For next time
- i added an [issue](https://github.com/VEuPathDB/microbiomeComputations/issues/20) for improving the computed variable response. I'll add more details to that issue soon but that should then prevent the web-eda plugins from defining a `variableId`.


dmfalke (approved):
Overall, the logic seems good. I wonder if some of it could be moved into `VariableCoverageTable` somehow, or handled in a utility function? I'm specifically thinking about [this code](https://github.com/VEuPathDB/web-eda/blob/8da327d7df4efbdeecdee73881dfb1a5346cf97c/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L836-L847), but there might be more that can be generalized.

asizemore:
@dmfalke there is definitely more that can be generalized. I'm off next week so i might end up just merging this and making a new tech debt ticket to improve the organization 😞 . I'll link it here if i do! Otherwise i'll work on it when im back!
I have a plan for how to improve it i just didn't get the time this week unfortunately!

dmfalke:
That is a-ok with me. Just something to keep in mind as we grow our collection of vizs supporting provided vars

Issue #787, opened 2021-12-22
More back end requests (`distribution` and `users/.../analyses`) need debouncing to pre-empt performance problems/back end stability with heavy use


bobular:
While testing that the `count` requests are being properly debounced when a user clicks on lots of categorical filter checkboxes, I noticed a bottleneck with other requests, namely `distribution` and `/eda-user-service/users/nnnn/analyses/xxxxx` 

The problem is so bad that if you click about 10-20 times in filter checkboxes, you can wait a minute or more before all requests are processed.

bobular:
This didn't make it to the new board.

If I do 5 rapid clicks on this subsetting table (as if I was selecting 5 countries, for example)
![image](https://user-images.githubusercontent.com/308639/174493715-9bc6df7a-5b58-473f-9804-77f1227f25dc.png)

The client makes 5 `distribution` requests:
![image](https://user-images.githubusercontent.com/308639/174493796-944b85ce-85df-4161-9584-389d237cf9d4.png)
and 7 user database updates:  (e.g. to /eda-user-service/users/578013513/analyses/ClinEpiDB/aKEdcNb)
![image](https://user-images.githubusercontent.com/308639/174493830-45ed4296-9aa0-4b2d-a818-e0b77240f44a.png)

Note the long times for the user analyses requests. These are performed in the background, and don't really affect the user (unless they close a tab and lose their latest work), but the load on the server cannot be good news.

I have also seen a massive backlog of the user analyses requests while navigating around the map viz.

The cost to "debounce" these requests is probably low, unless it opens a can of worms

dmfalke:
This issue doesn't have a UX component to it, so I'm going to move it to **Complete**

PR #1283, opened 2022-07-27
Data usage improvements

### LINKED ISSUE #1045: Reduce duplicated fetch calls in SubsettingDataGridModal
This issue extends [this issue regarding the performance of the download modal](https://github.com/VEuPathDB/web-eda/issues/988).

Specifically, when `handleSelectedVariablesChange` is first triggered on a new study, the `fetchPaginatedData` method is called 3 times. Subsequent click events only trigger one fetch call, as desired.

The fetch call is made [in this `useEffect`](https://github.com/VEuPathDB/web-eda/blob/76bd9137c126f216c8b6e073d8210076b0c2b767/src/lib/workspace/Subsetting/SubsettingDataGridModal.tsx#L237-L242).

dmfalke:
fixes #1045 #787

~This is a WIP. So far, #1045 has been addressed. I'd love some preliminary feedback from @jtlong3rd, especially regarding the changes to the analysis hooks.~

This is ready for review.

### DIFF
--- a/src/lib/core/hooks/analysis.ts
+++ b/src/lib/core/hooks/analysis.ts
@@ -61,8 +61,14 @@ const analysisCache: Record<string, Analysis | undefined> = {};
 
 export function usePreloadAnalysis() {
   const analysisClient = useAnalysisClient();
-  return async function preloadAnalysis(id: string) {
-    analysisCache[id] = await analysisClient.getAnalysis(id);
+  /**
+   * @param id Identifier of saved analysis.
+   * @param analysis Optional analysis object to use for the cache. This can be
+   * used if an analysis object was already fetched from the service. This
+   * should be used sparingly.
+   */
+  return async function preloadAnalysis(id: string, analysis?: Analysis) {
+    analysisCache[id] = analysis ?? (await analysisClient.getAnalysis(id));
   };
 }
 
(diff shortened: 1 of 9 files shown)


src/lib/core/hooks/analysis.ts
```diff
@@ -61,8 +61,14 @@ const analysisCache: Record<string, Analysis | undefined> = {};
...
-    analysisCache[id] = await analysisClient.getAnalysis(id);
+  /**
+   * @param id Identifier of saved analysis.
+   * @param analysis Optional analysis object to use for the cache. This can be
+   * used if an analysis object was already fetched from the service. This
+   * should be used sparingly.
+   */
+  return async function preloadAnalysis(id: string, analysis?: Analysis) {
+    analysisCache[id] = analysis ?? (await analysisClient.getAnalysis(id));
   };
```
jtlong3rd, line 72: Since `preloadAnalysis` with a passed `analysis` doesn't actually "preload" anything, it might be clearer to make a dedicated `cacheAnalysis` utility which receives an `id` and `analysis` object, and simply stores that `analysis` in the cache. You could then: 

1. Refactor `preloadAnalysis` to use `cacheAnalysis`, and don't allow an optional `analysis` argument
2. In the `useWorkspaceAnalysis` hook, directly use `cacheAnalysis` instead of `preloadAnalysis`

If you do this, I would add a comment that `cacheAnalysis` should be used sparingly, similar to what you did for `preloadAnalysis`.

```suggestion
  return async function preloadAnalysis(id: string) {
    cacheAnalysis(id, await analysisClient.getAnalysis(id));
  };
```

jtlong3rd (commented):
I think the overall approach you're taking ("cache a copy of the saved analysis which reuses the 'new' descriptor") makes sense.

src/lib/core/hooks/analysis.ts
dmfalke: This makes a lot of sense. I was thinking that the optional argument meant I was overloading the semantics of the function :slightly_smiling_face:

src/lib/core/hooks/entityCounts.ts
```diff
@@ -8,26 +8,24 @@ import { isStubEntity, STUB_ENTITY } from './study';
...
 export type EntityCounts = Record<string, number>;
 
-export function useEntityCounts(filters: Filter[] = []) {
+export function useEntityCounts(filters?: Filter[]) {
   const { id, rootEntity } = useStudyMetadata();
   const subsettingClient = useSubsettingClient();
 
-  const isStub = isStubEntity(rootEntity);
-
   // use JSON version in dependencies to prevent unnecessary recalculations
```
jtlong3rd, line 15:
```suggestion
```

jtlong3rd (approved):

bobular:
Hi! I started reviewing in a few spare minutes but don't have time to resolve the conflict right now. I guess I should roll back web-components - ah yes, that was easy enough to do.

dmfalke:
I'm looking into this conflict.

dmfalke:
> I'm looking into this conflict.

Resolving the conflict is a little tricky and will take me a little bit of time. I haven't yet had a chance to give it my full attention this morning.

dmfalke:
> > I'm looking into this conflict.
> 
> Resolving the conflict is a little tricky and will take me a little bit of time. I haven't yet had a chance to give it my full attention this morning.

The merge conflict is resolved now

PR #1284, opened 2022-07-28
Cleanup use analysis

### LINKED ISSUE #560: Handle subsequent updates when transitioning from a saved, to a new analysis
Currently, when a _new_ analysis becomes _saved_, the user can continue to add filters while the _saved_ analysis is being created on the back end. These additional filters are lost when the client receives the _saved_ analysis and replaces the analysis state.

We want to keep track of additional filters and apply them to the saved analysis once received.

See the description of #553 for more context.

### LINKED ISSUE #597: Fleeting "Visualization not found" warning shown after clicking on [+ New visualization]
This appears in a small font just below "View study details"

This only seems to happen in a brand new analysis / or for the first visualization added.

Might only be visible on dev instances this side of the Atlantic. No I see it on https://qa.clinepidb.org/ce.b19/app too

### LINKED ISSUE #787: More back end requests (`distribution` and `users/.../analyses`) need debouncing to pre-empt performance problems/back end stability with heavy use

dmfalke:
closes #787
closes #597
closes #560 

This PR is a first pass at simplifying some of the logic inside of the `useAnalysis` hook.

Notable changes are:
1. Setters are "pure" and only call state setters.
2. The `createAnalysis` function argument no longer returns the new analysis id. Because it performs the redirect, there isn't much reason to return the id, and this change simplifies expectations.
3. To that end, all redirect logic not occurs in `createAnalysis` (see changes in `StartPage.tsx`). This has a side benefit of addressing the pesky "Visualization not found" that appears when a new viz is created in a new analysis.
4. Some uses of `useState` were removed.

We might want to consider combining `useAnalysis` and `useWorkspaceAnalysis` into a single hook. I'm finding the indirection that passing in `createAnalysis` introduces to be confusion and maybe unnecessary.


jtlong3rd:
> We might want to consider combining `useAnalysis` and `useWorkspaceAnalysis` into a single hook. I'm finding the indirection that passing in `createAnalysis` introduces to be confusion and maybe unnecessary.

I see the merit - as I recall, `useAnalysis` was originally meant to be a generic "core interface" for interacting with EDA analyses, and potentially used in UIs other than the EDA workspace (e.g., the standalone MapVeu app). It's come to pass that we're not really using `us[ing]Analysis` outside of the EDA workspace, so I think it's fine to combine `useAnalysis` and `useWorkspaceAnalysis`.

dmfalke:
> > We might want to consider combining `useAnalysis` and `useWorkspaceAnalysis` into a single hook. I'm finding the indirection that passing in `createAnalysis` introduces to be confusion and maybe unnecessary.
> 
> I see the merit - as I recall, `useAnalysis` was originally meant to be a generic "core interface" for interacting with EDA analyses, and potentially used in UIs other than the EDA workspace (e.g., the standalone MapVeu app). It's come to pass that we're not really using `us[ing]Analysis` outside of the EDA workspace, so I think it's fine to combine `useAnalysis` and `useWorkspaceAnalysis`.

We've also been homing in on the idea that MapVeu will use the EDA workspace

jtlong3rd (commented):
Off to a great start! Getting `useSetter` out of the business of creating analyses has definitely simplified things. If/when `useAnalysis` and `useWorkspaceAnalysis` are combined, all the analysis creation logic could be concentrated in the `saveAnalysis` callback, which is a good thing.

(Worth noting: part of the reason why `useSetter` was creating analyses was to accommodate [an old requirement](https://github.com/VEuPathDB/web-eda/issues/283). We seem to have regressed some time in the past few months - now, trying to sort a filter in "browse and subset" triggers the creation of a backend analysis.)

dmfalke:
Aha, we should not "create" if only uiState is modified. I think this can be supported easily

dmfalke:
> Aha, we should not "create" if only uiState is modified. I think this can be supported easily

See https://github.com/VEuPathDB/web-eda/pull/1284/commits/50c8ba82867ac3f189eeed88856404bda5e8b621

dmfalke:
@jtlong3rd I've consolidated the hook logic and removed `useWorkspaceAnalysis`. Take a look when you get a chance and let me know what you think. Thanks!

dmfalke:
I think this is ready for review. I plan to merge this into `main` once #1283 is merged. I will then open a new PR to handle batching analysis update requests to the user service.

src/lib/core/types/AnalysisStore.ts
```diff
@@ -0,0 +1,32 @@
+import { AnalysisClient } from '../api/analysis-api';
+import { Analysis } from './analysis';
+
+interface SubscribeCallback {
+  (analysis: Analysis): void;
+}
+
+export class AnalysisStore {
```
jtlong3rd, line 8: Looking forward to seeing how this gets used!

jtlong3rd (approved):

src/lib/core/types/AnalysisStore.ts
dmfalke: Oh no! I did not mean to include that :facepalm:

dmfalke:
I removed the file I accidentally added, and I rebased on the latest from the target branch.

jtlong3rd (approved):

dmfalke:
> Shall I make a ticket for removing all the MapVeu [sic] code, prior to work starting on full screen viz's?

Ah, yes, that's probably a good idea.

bobular:
Cool - here it is https://github.com/VEuPathDB/web-eda/issues/1300

Issue #1207, opened 2022-06-15
Fix scrolling behavior of subset download modal


chowington:
The way scrolling works in the subset download modal isn't the best UX. See https://github.com/VEuPathDB/web-eda/issues/1169

danicahelb:
Discussed in 7/1/2022 EDA UX meeting

The separate scrollable sections for the variable tree and for the data table are looking really nice!

The header is wasting a lot of the modal's real estate. Please make these changes:
- For longer study titles, do not wrap the title onto 2 lines. Instead, shorten with ellipses and show the full title on mouseover
- Make the analysis name the 2nd line in the header, under the study title
- analysis name and entity counts can be placed on the upper left side of the modal, below the header

sweverschulman:
I can confirm that three dots and the mouse over do work. However, the mouse over is a little bit delayed, as if it needs to "think"

Could @chowington or @danicahelb confirm that this is ok, before I officially sign off.


![Image](https://user-images.githubusercontent.com/62298921/182877693-dbf1e618-0307-4d89-a439-da77ab7c9978.png)

danicahelb:
I agree, it takes a long time for the mouseover text to appear. @chowington how difficult would this be to speed up?

chowington:
Not bad. This is due to us using different tooltip methods in different areas of the site. The styling group will discuss harmonizing tooltips at a future meeting. For now I will use a tooltip component that appears faster than the current method.

sweverschulman:
The mouseover text speed has increased dramatically and there is no longer a long lag.

PR #1285, opened 2022-07-28
Fix study name truncation for modal title

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1207 (QA comment)

This should work as long as we don't get a study with < or > in the title text itself, which we don't have at this point

### DIFF
--- a/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
+++ b/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
@@ -706,21 +706,32 @@ export default function SubsetDownloadModal({
 
   // ~18px (round to 20px) per character for medium title size
   const maxStudyNameLength = Math.floor(modalHeaderWidth / 20);
-  const studyName =
-    studyRecord.displayName.length > maxStudyNameLength ? (
-      <span title={studyRecord.displayName}>
-        {safeHtml(
-          studyRecord.displayName.substring(0, maxStudyNameLength - 2) + '...'
-        )}
-      </span>
-    ) : (
-      safeHtml(studyRecord.displayName)
+  const regexAllTags = /<.*?>/g;
+  const studyNameNoTags = studyRecord.displayName.replaceAll(regexAllTags, '');
+  let studyNameElement: JSX.Element;
+
+  if (studyNameNoTags.length > maxStudyNameLength) {
+    const tagCharCount =
+      studyRecord.displayName.length - studyNameNoTags.length;
+    const studyNameTrunc = studyRecord.displayName.substring(
+      0,
+      maxStudyNameLength + tagCharCount - 2
     );
+    const regexUnclosedTag = /<(?!.*>).*/;
+    const studyNameNoUnclosedTag = studyNameTrunc.replace(regexUnclosedTag, '');
+    studyNameElement = (
+      <span title={studyNameNoTags}>
+        {safeHtml(studyNameNoUnclosedTag + '...')}
+      </span>
+    );
+  } else {
+    studyNameElement = safeHtml(studyRecord.displayName);
+  }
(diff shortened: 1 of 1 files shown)


src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
```diff
@@ -706,21 +706,32 @@ export default function SubsetDownloadModal({
...
   // ~18px (round to 20px) per character for medium title size
   const maxStudyNameLength = Math.floor(modalHeaderWidth / 20);
-  const studyName =
-    studyRecord.displayName.length > maxStudyNameLength ? (
-      <span title={studyRecord.displayName}>
-        {safeHtml(
-          studyRecord.displayName.substring(0, maxStudyNameLength - 2) + '...'
-        )}
+  const regexAllTags = /<.*?>/g;
+  const studyNameNoTags = studyRecord.displayName.replaceAll(regexAllTags, '');
```
jernestmyers, line 710: I did a similar thing not long ago, and it was suggested that I reach instead for [this `stripHTML` utility function](https://github.com/VEuPathDB/WDKClient/blob/1d179fe1825dac38134273fa0e1d8de670e96f22/Client/src/Utils/DomUtils.ts#L285-L292) instead.

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
```diff
@@ -706,21 +706,32 @@ export default function SubsetDownloadModal({
...
+    const studyNameTrunc = studyRecord.displayName.substring(
+      0,
+      maxStudyNameLength + tagCharCount - 2
+    );
+    const regexUnclosedTag = /<(?!.*>).*/;
+    const studyNameNoUnclosedTag = studyNameTrunc.replace(regexUnclosedTag, '');
+    studyNameElement = (
+      <span title={studyNameNoTags}>
+        {safeHtml(studyNameNoUnclosedTag + '...')}
       </span>
```
dmfalke, line 725: FYI, you can pass props to `safeHtml`. It defaults to using a `<span>` tag to wrap the content passed in as the first argument.
```suggestion
      {safeHtml(studyNameNoUnclosedTag + '...', { title: studyNameNoTags })}
```

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
dmfalke: The current solution can introduce false positives:
image

chowington:
Thanks for the feedback @jernestmyers and @dmfalke! I implemented your suggestions 👍

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
```diff
@@ -706,21 +707,30 @@ export default function SubsetDownloadModal({
...
-      safeHtml(studyRecord.displayName)
+  const studyNameNoTags = stripHTML(studyRecord.displayName);
+  let studyNameElement: JSX.Element;
+
+  if (studyNameNoTags.length > maxStudyNameLength) {
+    const tagCharCount =
+      studyRecord.displayName.length - studyNameNoTags.length;
+    const studyNameTrunc = studyRecord.displayName.substring(
+      0,
+      maxStudyNameLength + tagCharCount - 2
```
asizemore, line 718: the -2 is just an extra buffer?

asizemore (approved):
Works great!

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
chowington: It's to account for adding the `...` :)

PR #1290, opened 2022-08-02
temporary solution for SVG log scale

moontrip:
This relies on https://github.com/VEuPathDB/web-components/pull/381. Also, this draft PR is branched out from https://github.com/VEuPathDB/web-eda/pull/1241 as this needs map's log scale functionality

This is a temporary solution, just minimizing gap between small and large (using mouse tool) charts. Please see this comment for more detailed info: https://github.com/VEuPathDB/web-components/issues/380#issuecomment-1203128901


bobular:
Please see the comment in the original ticket: https://github.com/VEuPathDB/web-components/issues/380

I don't think a temporary solution is needed.

moontrip:
This PR is no longer needed so closing

PR #1294, opened 2022-08-05
Use MUI Tooltip component in subset download modal

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1207 (tooltip QA comment)


jernestmyers:
I'm wondering if it makes more sense to pull the `Tooltip` component from web-components so that it would have the same yellow-ish background as the ones used in the CheckboxTree?

I get this is probably a bit of an ad hoc fix and that we are hoping to unify these eventually, but it may be easier to unify an existing component than to revisit each instance of a MUI Tooltip (though it may be more likely that we put the Tooltip in CoreUI which would render that point moot anyway!).

chowington:
This is even more ad hoc than is apparent, because I purposefully didn't update the tooltip for the X button on the columns, as changing it away from just using a `title` attribute lead to a confusing bug where that tooltip would be hidden by the tooltip for sorting columns. Therefore it's currently necessary to keep that tooltip as a `title` attribute, which gives it a dark background, and I think for consistency we can keep the rest of these tooltips having a dark background as well. I believe the only yellow background tooltip on the subset modal comes from the help icon, and I'm okay with that having a different tooltip style for now because it can in fact be seen as a different kind of tooltip (one attached to a help icon whose sole purpose is to provide the tooltip).

jernestmyers (approved):
Okay, so maybe we should at the very least open an issue to unify the tooltips on this modal. In that issue, I think it would be helpful to provide some details on the issues you encountered when trying to add a tooltip to the column's remove buttons.

chowington:
Thanks @jernestmyers! We'll be sure to tackle that after making the CoreUI Tooltip component 👍

PR #1296, opened 2022-08-08
Temporary svg log scale

moontrip:


moontrip:
no need to have this so closing

Issue #1252, opened 2022-07-12
Swap toggles for CoreUI Switch


asizemore:
CoreUI has a `Switch` component intended to replace the toggles, and listens to the site theme. Let's exchange instances of toggles for the CoreUI `Switch`.

Not a high priority, but fun and low cost i'd guess.

dmfalke:
I scheduled this for this release. It's a nice bit of cleanup :slightly_smiling_face:

chowington:
@asizemore Do you think we should also replace the custom toggle (and other custom widgets as appropriate) in web-components? They're used mainly in the plot controls

asizemore:
@chowington yes i think that'd be great! Thank you! They should be using the primary theme color i'd think?

SheenaTomko:
Looks great! Replaced in drop down menus, below stratification variables, and in plot parameters below plots

PR #1298, opened 2022-08-08
Use CoreUI switch

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1252
Requires https://github.com/VEuPathDB/CoreUI/pull/101


src/lib/core/components/Switch.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const Switch = ({
+  selectedOption,
+  disabled,
+  onOptionChange,
+  labels,
+  size = 'medium',
+  styleOverrides,
+}: SwitchProps) => {
+  return (
+    <FilledSwitch
```
jernestmyers, line 26: Could you help me understand why we'd want to create a `Switch` component in `web-eda` instead of just directly importing `FilledSwitch` from `CoreUI`? I understood why we'd refactor the existing `web-component`'s toggle to use the `CoreUI` switch as its internals. Though we may in the future want to hunt down and replace all imports of the `web-components` switch to instead come from` CoreUI`, it's a simpler and quicker approach to just update the existing component. But I'm not quite seeing a benefit to `web-eda` also having a `Switch` component.

chowington: Yeah! It's just because we use this same styling of the switch three different places in the EDA, and I thought duplicating the styling would not be very DRY:

```
      options={[false, true]}
      styleOverrides={{
        default: [
          { backgroundColor: grey[500] },
          { backgroundColor: blue[800], knobColor: 'white' },
        ],
        hover: [
          { backgroundColor: grey[500] },
          { backgroundColor: blue[800], knobColor: 'white' },
        ],
        disabled: { backgroundColor: grey[300] },
      }}
```

dmfalke: I'm not sure, but I think we should use the same styling everywhere. If we really want different styles in some places, maybe we should make variants in coreui, similar to what was done with [buttons](https://github.com/VEuPathDB/CoreUI/tree/main/src/components/buttons)?

asizemore: Can the switch just use theme colors? For example, blue[800] doesn't match clinepi's teal or mbio's mutedBlue themes. I don't think there was originally a reason that it needed to be gray. 
If we want it to be colored when on and gray when off, could we use the default 'disabled' styling for the component?

chowington:
Ready for review again!

src/lib/workspace/AllAnalyses.tsx
```diff
@@ -336,20 +336,18 @@ export function AllAnalyses(props: Props) {
...
-              style={{
-                padding: '1em',
-              }}
+              labelPosition="right"
+              value={sortPinned}
+              onChange={setSortPinned}
               disabled={pinnedAnalyses.length === 0}
+              styleOverrides={{
+                container: { marginLeft: '1em' },
+                mainColor: { hue: red, level: 500 },
```
asizemore, line 347: why is this red instead of the theme color?

asizemore (commented):
Looking good! Just left one question.

Is there a reason to not swapping out the other switches as well? For example the no data switch and log scale switch.

src/lib/workspace/AllAnalyses.tsx
chowington: And about the red---whoops, that was just me testing that prop. Forgot to replace it. Good catch!

jernestmyers (approved):

PR #1304, opened 2022-08-10
Prevent map overlay back end 500 with correct binSpec range using fancy significant figures rounding up/down

### LINKED ISSUE #1301: SCORE Mozambique variable gives 500 error in map viz overlay for School-age children in village count
Looks like this bug https://github.com/VEuPathDB/plot.data/issues/179 is back!

Although this time it's **School-age children in village count** (last time it was School-age children **treated** in village count, which works fine now)

Study: **SCORE Mozambique S. haematobium Cluster Randomized Trial**
Viz: map
Variable: School-age children in village count

In the logs the error is now:
```
Error in plot.data::validateBinRange(xVals, binRange, "NUMBER", FALSE) : 
  Invalid bin range provided: Bin range cannot represent a subset of the data range.
```

Update for testers:

If there is no pink warning, all is good!

bobular:
Fixes #1301


moontrip (commented):
@bobular It works fine without error but I have one question and one suggestion: Q) what will happen if newValue === value at the RoundUp/Down functions? 
My suggestion is that actually the two functions RoundUp & Down are identical except the if condition: please correct me if I am wrong. So perhaps it can be combined with condition? e.g., 

`const adjusted = newValue < value ? newValue + epsilon : newValue - epsilon;`

Above code is incomplete for the case of newValue === value. If the equal condition should be excluded then we may use if-else if at a single function like 

`if (newValue < value) {
...
} else if (newValue > value) {
...
} `

moontrip:
Or we may introduce a new prop at a single function like upDown: "up" (or "down") to round-up or round-down if necessary

moontrip (approved):
Yes I realized the necessity for distinguishing round-up and down so I am fine with two functions @bobular But perhaps it may still be simplified if introducing 3rd prop on the function to set up up/down functionality, but I will leave the decision to you

bobular:
If we went with the unified function, we'd need to roll in the functionality of the existing `numberSignificantFigures` (for when neither 'up' nor 'down' is provided).

In that case do we continue to do the "if integer, return unchanged" behaviour?

The function has only been used in the default axis range function, and the motivation was to avoid ridiculous 0.111111111111111 appearing in the range control inputs.

I think we could keep the old integer behaviour in the unified function.  I'll have a think about it.

Issue #1266, opened 2022-07-25
Remove hardcoded clinepi-specific text


asizemore:
<img width="1723" alt="Screen Shot 2022-07-25 at 10 40 02 AM" src="https://user-images.githubusercontent.com/11710234/180803872-493385ed-8bdf-45fc-958d-2cd08d767e50.png">

While on mbio, i get a warning about making the analysis appear on the ClinEpiDB page. I propose we change the language to say
"This analysis will appear on the Public analyses page." Alternatively, we do get the project name so we might be able to use that in the text.

dmfalke:
A quick search shows that we use the string "ClinEpiDB" in a few places. We should never hardcode a website name in the project.

asizemore:
Seems like it should be possible to use the `projectId` var which we use to filter apps ([line](https://github.com/VEuPathDB/web-eda/blob/8e149f94a57fb4be89918aa47da26698e6e378d7/src/lib/workspace/ComputationRoute.tsx#L41))? 

Otherwise we should just make the text here and in [Variables.tsx](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/workspace/Variable.tsx) general. 

Between Variables.tsx and this public analysis warning, i think that's all the hardcoded website names!

dmfalke:
We should either make it generic or use https://github.com/VEuPathDB/WDKClient/blob/6d01ca1a547f9c8df65c5bd6d3ff597da9fe3313/Client/src/Service/ServiceBase.ts#L73. If the latter, we might want to make it a toplevel prop for `WorkspaceRouter` and make it available via a hook.

SheenaTomko:
Looks good

PR #1306, opened 2022-08-10
Remove references to Clinepidb in site text

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1266


src/lib/workspace/sharing/ConfirmPublicAnalysis.tsx
```diff
@@ -72,7 +72,7 @@ export default function ConfirmPublicAnalysis({
                 target="_blank"
                 style={{ color: theme?.palette.primary.hue[600] }}
               >
-                ClinEpiDB Public analyses page.
+                Public analyses page.
```
bobular, line 75: I know the original has a period, but this isn't a sentence and likely doesn't need one.

How does it look on the webpage (could you provide a screenshot pls - I am lazy!)

bobular (dismissed):
In lieu of Ann, I did this gruelling review instead.  Had one comment/question!

src/lib/workspace/sharing/ConfirmPublicAnalysis.tsx
chowington: Hi Bob! It is a sentence actually---the sentence says "This analysis will appear on the Public analyses page." Here's the modal:

image

dmfalke (commented):
I would ask for feedback on the verbiage related to the original variable name. I think it's okay, but they might prefer including something about the data being "integrated".

src/lib/workspace/Variable.tsx
```diff
@@ -84,8 +84,7 @@ export function VariableDetails(props: Props) {
               {showMore && providerLabelLeftover}
               &nbsp;
               <HelpIcon>
-                The name of this variable in the data files that were integrated
-                into ClinEpiDB
+                The name of this variable in the orignal data files
```
chowington, line 87: @dmfalke That makes sense. @danicahelb or @SheenaTomko, any feedback on the above change?

SheenaTomko: I'm ok with it. @danicahelb? I know the files we integrate aren't always "original". If you want something like "The name of this variable in the data file it was integrated from" I would be ok with that too

asizemore (approved):
The changes look good to me! But I agree with the other comments that suggest @SheenaTomko and @danicahelb are probably the ones who need to approve the verbiage

src/lib/workspace/Variable.tsx
chowington: @danicahelb To be clear, which verbiage did you approve? Mine above or Sheena's below mine?

chowington: I believe in the meeting she said the suggestion I put forth above. I'll merge, but feel free to correct me, @danicahelb!

chowington:
@bobular Seems like you specifically need to approve my changes since you submitted a change request?

dmfalke (approved):
@chowington I could be wrong, but I think the reason you can't merge is that the branch is out of date with `main`. Can you try merging the latest and see if you're able to merge this PR?

chowington:
@dmfalke I don't think so! I just merge in main and here's what I'm seeing (you might be seeing it too):

image

dmfalke:
> @dmfalke I don't think so! I just merged in main and here's what I'm seeing (you might be seeing it too):
> 
> image

Ah, ok. I saw a different message yesterday. Let me see what I can do.

dmfalke:
I "dismissed" @bobular's review, so you should be able to merge now.

PR #1308, opened 2022-08-11
Upgrade dev deps

dmfalke:
This PR upgrades several development dependencies. The major update is to react-scripts@5.


dmfalke:
I pushed a few more changes, including typescript!

PR #1312, opened 2022-08-12
polyfill process

dmfalke:
Fixes the error `process is not defined`


jtlong3rd (approved):

bobular:
Can confirm the "process" error has gone away in the map.

Can also confirm that while testing this I found a new bug in the map chart markers (EDA-side) :cry:

dmfalke:
Is the new bug related to the recent changes I made?

PR #1321, opened 2022-08-16
Only request study release data if recordClass inludes the DownloadVersions table

### LINKED ISSUE #1145: Download tab: disable whole study section for mbio beta release
The whole dataset files for mbio are unlikely to be ready by the initial beta site release. Is it simple to either put "Coming soon!" or just totally remove the Full Dataset section of the Downloads tab? 

If i understand correctly, the My Subset portion is just generated by the eda so as long as the eda works, that section will work.

dmfalke:
fixes #1145

I noticed that the MicrobiomeDB WDK model does not currently include the 'DownloadVersions' table. This is the current source of the error when visiting the **Downloads** tab (the WDK service will return a 500 when request a field for a record that does not exist).

This PR addresses the error by checking that the table exists in the record class. If it doesn't, then we don't request the table. Once the download files are ready, we can add that table to the dataset record in MicrobiomeDB. If we need to test that the table works before pushing to QA, we can add the table to the model on a development website and test there.

### DIFF
--- a/src/lib/core/hooks/study.ts
+++ b/src/lib/core/hooks/study.ts
@@ -12,7 +12,10 @@ import {
   getScopes,
   getNodeId,
 } from '@veupathdb/wdk-client/lib/Utils/CategoryUtils';
-import { AnswerJsonFormatConfig } from '@veupathdb/wdk-client/lib/Utils/WdkModel';
+import {
+  AnswerJsonFormatConfig,
+  RecordInstance,
+} from '@veupathdb/wdk-client/lib/Utils/WdkModel';
 
 // Definitions
 import {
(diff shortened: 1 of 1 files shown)


dmfalke:
This is the relevant line to update in the model https://github.com/VEuPathDB/EbrcModelCommon/blob/ab61b5ef1587546e242ce7c4ba7370ad8de4c06a/Model/lib/wdk/model/records/datasetRecords.xml#L444

asizemore (approved):
Really clear. Thank you @dmfalke !

dmfalke:
> > This is the relevant line to update in the model https://github.com/VEuPathDB/EbrcModelCommon/blob/ab61b5ef1587546e242ce7c4ba7370ad8de4c06a/Model/lib/wdk/model/records/datasetRecords.xml#L444
> 
> 
> 
> I'm sorry i didn't see this earlier! I went ahead and updated the model for fear of forgetting to update it later (now that we won't see an error when we visit the page). I'm sorry to cause a bit of duplicated effort!

If adding that table to the mbio model addresses the error message, then I'm happy!

dmfalke:
It looks like adding the table to the model is sufficient, so I am going to close this PR and associated issue.

PR #1339, opened 2022-08-25
add download tab warning banner for improperly formatted mbio studies

### LINKED ISSUE #1337: Microbiome downloads: warn users to use live site downloads for some studies
As noted [here](https://github.com/VEuPathDB/EdaLoadingIssues/issues/37#issuecomment-1225770900) by Sheena, the download files aren't quite right for the mbio longitudinal studies because of how the data was loaded. Since we can't fix the data in time for the release, Sheena had the good idea of just adding a banner that warns users that they should use download files from the live site instead of the beta site.

Plan: add a banner to the download tab for certain files, similar to how the viz tab has a banner for large clinepi studies.

asizemore:
Resolves #1337 

In QAing [this issue](https://github.com/VEuPathDB/EdaLoadingIssues/issues/37#issuecomment-1225928487), @SheenaTomko found some of the download files are not correct, specifically the Participant and Sample files for longitudinal studies. She proposed we add a warning banner to direct users to the live site for downloads until we can fix the data issue (b60).

This PR adds the warning banner to any study that is a) longitudinal (including anything that said "prospective cohort design") and b) is missing participant repeated measures. The approach follows that taken for adding the warning banner on the visualization tab for large studies having slow visualizations.

I struggled with the language for the banner quite a bit. @SheenaTomko I'd be very grateful for your advice on this! I think some files (assays) are usually fine, but the particpant and sample ones are odd. They're not showing incorrect data as much as the mapping between variable and entity doesn't really make sense. So anyways I struggled with how to convey this in a friendly way that does not compromise trust in the future when we will have sparkly perfect download files :)

### DIFF
--- a/src/lib/workspace/DownloadTab/index.tsx
+++ b/src/lib/workspace/DownloadTab/index.tsx
@@ -14,6 +14,7 @@ import { DownloadClient } from '../../core/api/DownloadClient';
 // Components
 import MySubset from './MySubset';
 import CurrentRelease from './CurrentRelease';
+import Banner from '@veupathdb/coreui/dist/components/banners/Banner';
 
 // Hooks
 import { useWdkStudyReleases } from '../../core/hooks/study';
@@ -25,6 +26,7 @@ import { useAttemptActionCallback } from '@veupathdb/study-data-access/lib/data-
 import { useWdkService } from '@veupathdb/wdk-client/lib/Hooks/WdkServiceHook';
 import { Action } from '@veupathdb/study-data-access/lib/data-restriction/DataRestrictionUiActions';
 import { getStudyRequestNeedsApproval } from '@veupathdb/study-data-access/lib/shared/studies';
+import { useLocalBackedState } from '@veupathdb/wdk-client/lib/Hooks/LocalBackedState';
 
 type DownloadsTabProps = {
   downloadClient: DownloadClient;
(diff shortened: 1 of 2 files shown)


src/lib/workspace/DownloadTab/index.tsx
```diff
@@ -62,6 +64,38 @@ export default function DownloadTab({
...
+  const [
+    shouldShowWarning,
+    setShouldShowWarning,
+  ] = useLocalBackedState<boolean>(
+    true,
+    SHOULD_SHOW_DOWNLOAD_WARNING_KEY,
+    (boolean) => String(boolean),
+    (string) => string !== 'false'
+  );
+  const studiesForDownalodWarning = [
```
jernestmyers, line 79: Small typo: `studiesForDownalodWarning` should be `studiesForDownloadWarning`

src/lib/workspace/DownloadTab/index.tsx
```diff
@@ -93,6 +128,20 @@ export default function DownloadTab({
           fontSize: '1.4em',
         }}
       >
+        {/* The following is part of the temporary solution for funky mbio download files */}
+        {studiesForDownalodWarning.includes(datasetId) && shouldShowWarning && (
```
jernestmyers, line 132: Will need to update this typo as well.

jernestmyers (commented):
Brilliant approach overall! 🥇 

I also like Sheena's tweaks to the banner message.

asizemore:
Thank you both! Typos fixed and warning message updated. 
Screen Shot 2022-08-25 at 11 20 34 AM

> Brilliant approach overall! 🥇
Lol i learn from the best!

jernestmyers (approved):

PR #1341, opened 2022-08-25
filter out any provided variable from the required variable logic

### LINKED ISSUE #1340: Remove provided variables from required variables list
Provided variables are not part of the constraints and so do not have an `isRequired` property. This causes an error for all apps. 

(from my dev site with the latest web-eda code as of this morning)
Screen Shot 2022-08-25 at 12 17 11 PM

Please remove provided variables from the required vars logic

asizemore:
Resolves #1340 

Provided vars won't be in the constraints, so I filtered them out of the logic using the readonlyValue (should only be defined for provided vars)

### DIFF
--- a/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/src/lib/core/components/visualizations/InputVariables.tsx
@@ -238,7 +238,9 @@ export function InputVariables(props: Props) {
                 <h4>{sectionInfo[inputRole ?? 'default'].title}</h4>
               </div>
               {inputs
-                .filter((input) => input.role === inputRole)
+                .filter(
+                  (input) => input.role === inputRole && !input.readonlyValue
+                )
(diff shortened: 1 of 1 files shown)


jernestmyers:
Sorry for introducing this error. I should have tested [my required inputs work](https://github.com/VEuPathDB/web-eda/pull/1245) on mbio as well. I think I'd prefer to just re-work this logic such that the fixed variables are still rendered. To me, that provides a more consistent UI across our visualizations.

If that sounds okay to you, maybe we could close this PR and [its follow-up issue](https://github.com/VEuPathDB/web-eda/issues/1342) and I'll re-work the rendering logic instead of filtering out `!input.isReadOnly`.

asizemore:
No worries! Sounds like a good plan to me! I'll close this PR and the followup issue

PR #1344, opened 2022-08-28
Use dataset id, rather than study id

dmfalke:
When this code was refactored, I used the EDA study ID, rather than the WDK dataset ID for the `studyId` field of the analysis object. This highlights the confusion around using the property name `studyId`. I will make a new issue about this soon.


jtlong3rd (approved):
Crisis averted!

aurreco-uga (approved):

dmfalke:
It's been a while, but I made issue #1524 to address the confusion between `studyId` and `datasetId`.

PR #1345, opened 2022-09-01
More control over visualization plot layouts and `useMapMarkers` hook

### LINKED ISSUE #1333: Full Screen Map: Update visualizations to work with Full Screen Map
- [ ] Include `Option` to disable/hide the faceting input.
- [ ] Include `Option` to allow a provided overlay variable.

bobular:
Fixes #1333 

Made a start on this, looking for feedback while still in draft please.

I extended the `XyzVisualization.withOptions()` approach to allow the optional customization of `layoutComponent` (see new `GlobalVisualizationOptions`).

By default each component will use the regular `PlotLayout` but you can instantiate a visualization with a different layout method. See the commented out code in `pass.tsx` that gives you a very minimal Boxplot (inputs, plot, and nothing else).

The worst part was disentangling the plot from the controls (`BoxplotWithControls` -> two separate components/functions). It wasn't difficult, just fiddly and a lot of moving stuff around. TS helps massively of course.

I have also refactored the fetching of map marker data into its own custom hook (`useMapMarkers`) but currently this only works when an overlay variable is provided.   It has a new prop `requireOverlay` which when `true` makes the hook behave as desired for the old map vizualization, and when `false` produces simple count-only markers when no overlay variable (confusingly also called xAxisVariable, but that does also make sense) is provided.

HistogramVisualization major refactor to avoid a large amount of prop-passing.

Now all visualizations are refactored to avoid having a "XyzPlotWithControls" sub-component with a ton of prop-passing. It means that the main component function is monolithic, but everything is named clearly and all vizs are the same structure now, so further refactoring should be easier.

All visualizations **apart from** Mosaic and Map (which we won't make available as floating visualizations for now) now have the `options.getOverlayVariable()` override for use as a floating visualization.

### DIFF
--- a/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/src/lib/core/components/computations/plugins/abundance.tsx
@@ -57,7 +57,7 @@ export const plugin: ComputationPlugin = {
           };
         }
       },
-      getOverlayVariable(config) {
+      getComputedOverlayVariable(config) {
         if (AbundanceConfig.is(config)) {
           return config.collectionVariable;
         }
--- a/src/lib/core/components/computations/plugins/pass.tsx
+++ b/src/lib/core/components/computations/plugins/pass.tsx
@@ -30,9 +30,16 @@ export const plugin: ComputationPlugin = {
     // densityplot: scatterplotVisualization,
     barplot: barplotVisualization,
     boxplot: boxplotVisualization,
-    // or...
-    // boxplotVisualization.withOptions({
-    //   layoutComponent: FloatingLayout,
-    //}), /// TEMPORARY ONLY!!! ///
+    //     or...
+    //    boxplot: boxplotVisualization.withOptions({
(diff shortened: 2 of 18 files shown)


moontrip:
@bobular interesting to see this separation/decomposition 👍 Just one comment from my glancing - I will checkout this and see more details. So, since you also tried to separate Controls as controlsNode at PlotLayout component, I think that you missed it at FacetedPlotLayout's props?

bobular:
> @bobular interesting to see this separation/decomposition +1 Just one comment from my glancing - I will checkout this and see more details. So, since you also tried to separate Controls as controlsNode at `<PlotLayout>`, I think that you missed it at FacetedPlotLayout's props (and its handling at `<FacetedPlotLayout>` like done at `<SinglePlotLayout>`?

Thank you @moontrip - you're correct - the controls are missing for faceted boxplots and scatterplots! I will fix that.

Issue #1334, opened 2022-08-24
Full Screen Map: Implement FullScreenMapApp


dmfalke:
This depends on #1326, #1327, #1333

- [x] Define UI state
- [x] Create a modified set of visualizations based on those used by the pass app. This will include passing options to disable faceting and provide an overlay variable.
- [x] Render map using the FullScreenApp component (see #1327)
- [x] #1368

jernestmyers:
Closed in 09/28/2022 EDA dev meeting.

PR #1349, opened 2022-09-06
Fsm all

### LINKED ISSUE #1327: Full Screen Map: full screen component
This issue is about implementing the React component that will be used for the full screen computation/app. See the [mock-ups](https://docs.google.com/presentation/d/17Yg4Kz00pW6wbNo3bK0DpQjW4cVEzrq0VQ75pw1xZNc/edit). Note that for this issue, we are planning to only support **Visualize** mode. Thus, we will not need the **Subset | Visualize** toggle.

In addition to the MapVEuMap component, we will need the following:

- [x] #1347
- [x] Viz selector modal. (See https://docs.google.com/presentation/d/17Yg4Kz00pW6wbNo3bK0DpQjW4cVEzrq0VQ75pw1xZNc/edit#slide=id.g13e1893ab5c_0_310)
- [x] Floating viz component for active viz. (See https://docs.google.com/presentation/d/17Yg4Kz00pW6wbNo3bK0DpQjW4cVEzrq0VQ75pw1xZNc/edit#slide=id.g13e1893ab5c_0_333)
- [x] Update computation config based on map interactions.

Note that we might want to use an icon for viz thumbnails since thumbnails can frequently become invalid since they are sensitive to the map viewport.

Also note that we still need to finalize what the computation configuration (aka, ui config) will look like.

dmfalke:
Fixes #1327 #1334

This PR is a continuation of (and supersedes) #1346. It is also based on the branch `fsm-alt-viz-layouts`, although this PR is currently targeting `main`.

You can compare with `fsm-alt-viz-layouts` here: https://github.com/VEuPathDB/web-eda/compare/fsm-alt-viz-layouts...fsm-all.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -101,6 +101,7 @@ import { useDefaultAxisRange } from '../../../hooks/computeDefaultAxisRange';
 import { useVizConfig } from '../../../hooks/visualizations';
 import { createVisualizationPlugin } from '../VisualizationPlugin';
 import { histogramDefaultDependentAxisMinMax } from '../../../utils/axis-range-calculations';
+import { LayoutOptions } from '../../layouts/types';
 
 export type HistogramDataWithCoverageStatistics = (
   | HistogramData
@@ -165,8 +166,11 @@ export const HistogramConfig = t.intersection([
   }),
 ]);
 
-function HistogramViz(props: VisualizationProps) {
+interface Options extends LayoutOptions {}
+
+function HistogramViz(props: VisualizationProps<Options>) {
   const {
+    options,
     computation,
     visualization,
     updateConfiguration,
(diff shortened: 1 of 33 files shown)


moontrip:
@dmfalke merged legend work by Connor to this PR. Checked that the legend works fine in this branch/PR after merge, as exampled in the screenshot. Note that I intentionally unchecked 9-12 years to test and then captured the screenshot :)

fsn-legend1

dmfalke:
The following items have been addressed by my most recent commits (https://github.com/VEuPathDB/web-eda/pull/1349/files/76941e8878fb095a9372bb4f3ce3ea8ad8a1d62e..90e1ed3b8ba9ec7bf54e5ff53256077e0ceb4474):

- Save analysis before loading the FSM.
- Update styling for close button, so it is easier to find.
- Add starred variables to FSM.
- Propagate checked legend items to linked visualizations.

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -878,15 +672,38 @@ function HistogramPlotWithControls({
...
+      data,
+      checkedLegendItems,
+      vizConfig.dependentAxisLogScale,
+      vizConfig.dependentAxisRange,
+    ]
+  );
+
+  const overlayLabel = variableDisplayWithUnit(overlayVariable);
+
+  const histogramProps: HistogramProps = {
```
moontrip, line 688: @dmfalke I am testing this grand merge (😄) for my past commits like axis label power format and proportion bugs fir histogram and barplot Vizs. They seem to work fine for me. However, I found a minor bug on histogram Viz, which is probably inherited from Bob's Viz separation works. There is a missing prop in this histogramProps, so the following should also be added - histogram plot style is not being applied without it, e.g., no box shadow, height, etc.

`containerStyles: !isFaceted(data.value) ? plotContainerStyles : undefined,`

dmfalke: Thanks @moontrip! I will add this.

dmfalke: Actually, I think this is addressed here:
https://github.com/VEuPathDB/web-eda/blob/289293be9c1f6a5a738de6e258fa0e0f8559854e/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx#L719-L741

dmfalke: Oh, ignore that. That is something different.

dmfalke:
I think this is ready to be merged into `main`. I have added the feature flag we discussed, so the feature will not appear on qa sites by default.

moontrip:
@dmfalke I have tested your feature flag and it worked out by adding `REACT_APP_ENABLE_FULL_SCREEN_APPS=true` at env.local 👍 Also quite extensively tested Vizs with this PR and they seem to work fine too. Found one bug for Lineplot Viz that does not have spinner (data loading image) but Bob is addressing it at his PR, https://github.com/VEuPathDB/web-eda/commit/6da6d2a994ec597065775fa92fc3d934eee674c5, so probably no need to change it in this PR.

Didn't test full-screen map (FSM) itself much yet, but actually I have one question about the floating visualization under FSM. Do we need to allow to select `(Geolocation) map Viz` under FSM (at Select a visualization modal that is shown when pressing New visualization at top right of FSM) ? Not quite sure if it is intentional but small map inside FSM may not mean much? If we allow it, then it's layout itself is broken because the map size is larger than floating Viz size: see below screenshot. 

broken map1

dmfalke:
> Do we need to allow to select `(Geolocation) map Viz` under FSM [...]

That's a good question. Maybe @bobular can answer? Either way, I don't think we have to address that in order to merge this. I will give @bobular a few hours before taking any action, but I would like to merge this today.

moontrip:
> > Do we need to allow to select `(Geolocation) map Viz` under FSM [...]
> 
> That's a good question. Maybe @bobular can answer? Either way, I don't think we have to address that in order to merge this. I will give @bobular a few hours before taking any action, but I would like to merge this today.

@dmfalke Sounds good and I agree with your opinion, so I will leave my approval 👍

moontrip (approved):

src/lib/core/hooks/checkedLegendItemsStatus.ts
```diff
@@ -15,3 +15,52 @@ export function useCheckedLegendItemsStatus(
 
   return checkedLegendItems;
 }
+
+/**
+ * WIP custom hook to wrap all necessary checked legend items functionality
+ * into one
+ */
+
+export function useCheckedLegendItems(
```
bobular, line 24: This doesn't seem used, but I think that was the plan (to wait until after the mega merge)

bobular (approved):
Approve and will merge.

PR #1351, opened 2022-09-07
Streamlined `useMapMarkers` props

bobular:
Much cleaner props handling for `useMapMarkers`

May be safer if you do the merge, @dmfalke !!


dmfalke (approved):
This is great, thanks! I will go ahead and merge it, since I have more updates to push to `fsm-all`

PR #1354, opened 2022-09-09
Replace viz variable selectors with new CoreUI components

jernestmyers:
~~Depends on [this PR in CoreUI](https://github.com/VEuPathDB/CoreUI/pull/112).~~

Of note:
- I opted to add an `asDropdown` prop to both `VariableTree` and `VariableList`, which is passed down from `VariableTreeDropdown`. This allowed for the least amount of refactoring to accommodate the `CheckboxTree` implementation both in the variable selectors and in the `Browse and Subset`.
- The [clear selection button functionality](https://github.com/VEuPathDB/web-eda/pull/1354/files#diff-57aa3250ae63fcc91c2beb0589357ac2ff5e461a21263b9e4dd94f7414d3430aR81-R94) moved out of `VariableTreeDropdown` to `VariableTree` and gets passed into `VariableList` via a new prop. This eliminates any real functionality in `VariableTreeDropdown` aside from its descriptive name and the existing container classes.
- The majority of changes in `VariableList` are CSS adjustments.


jernestmyers:
At this point, the implementation/styling of the following contexts are ready for review:
1. All variations of variable selectors are pulling from `CoreUI`, including `SelectTree` (in lieu of `VariableTreeDropdown`), `SingleSelect` (line plot function selection), and `SelectList` (line plot proportion selections).
2. The variable tree in the `Browse and Subset` tab
3. The multi-selection variable tree in the download modal

Note: the latest CoreUI version used in this branch is sufficient for review, meaning a simple `yarn start` should get you up and running just fine. However, there is [a CoreUI branch](https://github.com/VEuPathDB/CoreUI/pull/116) with very minor tweaks like more space beneath the search box, vertically centering the lock icon in the label (download modal), and some other nitpicks.

dmfalke (changes requested):
I've looked at the code and test this, and there only seem to be a couple of things to address.

1. See comment about duplicate props in `VariableList.tsx`.
2. Can we get rid of `VariableTree.scss`, or at least remove unused styles?
3. I noticed that I cannot tab into the `SingleSelect` component in  `LineplotVisualization`. I think this is probably a CoreUI issue. I will make one there.

Otherwise, this is looking really good!

jernestmyers:
> 2. Can we get rid of VariableTree.scss, or at least remove unused styles?
`VariableTree.scss` is till serving some styling purposes in the download modal. I did look into removing its dependency on the style sheet and decided against doing it for now, but I can reevaluate that decision. The download modal instance even uses classes like `wdk-CheckboxTreeItem` and `wdk-CheckboxTreeNodeContent`.

> 3. I noticed that I cannot tab into the SingleSelect component in LineplotVisualization. I think this is probably a CoreUI issue. I will make one there.
I'm not 100% following this comment. I am able to tab onto the button, though the tab order of the viz buttons is a bit strange. Tabbing onto the `SingleSelect` comes after tabbing through the overlay and facet buttons (and the help icon). Once the menu is open, we can use arrows to navigate the menu. We disabled tabbing through the menu options because 1) it messed up the focus logic and 2) native select elements don't allow tabbing through options.

jernestmyers:
I _**think**_ I managed to excise all uses of `VariableList.scss` and have deleted it accordingly. I also went ahead and DRY'd up the `sharedProps` between `CheckboxTree` and `SelectTree`. And as I mentioned above, the tabbing of `SingleSelect` behaves as I expect it to, but perhaps I'm misunderstanding something.

dmfalke:
> I'm not 100% following this comment. I am able to tab onto the button, though the tab order of the viz buttons is a bit strange. Tabbing onto the `SingleSelect` comes after tabbing through the overlay and facet buttons (and the help icon). Once the menu is open, we can use arrows to navigate the menu. We disabled tabbing through the menu options because 1) it messed up the focus logic and 2) native select elements don't allow tabbing through options.

I didn't notice that the elements were tabbing out of order, so ignore that comment! It would be nice to fix the ordering, but that's not pertinent to this PR.

dmfalke (approved):

PR #1357, opened 2022-09-12
enabling mouseMode

moontrip:
@dmfalke @bobular @chowington 
I have tested FSM with legend too and it seems to work fine overall, esp. for demo purpose 👍 I noticed that mouseMode was implemented but not functional so I examined the issue and found that it actually missed a prop, showMouseToolbar. Also enabled checkedLegendItems as it seems to work fine too. 
For this reason, this PR is branched out from fsm-legend but one can just change the fsm-legend directly.

Here is an example screenshot with mouseMode at FSM

mouseMode1


moontrip:
These are now addressed at https://github.com/VEuPathDB/web-eda/pull/1349 thus closing this draft PR

Issue #1360, opened 2022-09-13
correct dependent axis range for Lineplot Viz proportion


moontrip:
Currently, dependent axis range for Lineplot Viz proportion is not showing any value before making a user input. Per https://github.com/VEuPathDB/web-eda/issues/1099, it should have default range of [0, 1] or [minPos, 1] for log scale. 

Here are screenshots of the current behaviors.

![lineplot-pro1](https://user-images.githubusercontent.com/12802305/189921837-9e7c8418-3193-4e28-90ea-2a1663347db0.png)

![lineplot-pro2](https://user-images.githubusercontent.com/12802305/189921858-644065df-bb88-465a-ac05-23ac31e68bed.png)

danicahelb:
@moontrip I don't think there was a bug here.

We never show the axis values until ALL required parameters are selected. For your example above in the line plot, configuring the proportion IS a required parameter (this parameter just gets activated only after the y-axis variable is selected).

I think the current behavior is fine

moontrip:
@danicahelb Thank you for your review! One of the problems is that dependent axis range is missing even if all the required ones are selected: please see the second screenshot where plot is drawn with the selection of the numerator but no dependent axis range is shown - it should be [0,1] without log scale per https://github.com/VEuPathDB/web-eda/issues/1099

bobular:
Oh, I see @danicahelb has already commented on the potential non-issue status.

But then I read your reply!  Yes that is a bug, that the y-axis range input boxes are empty, even when data is plotted.

danicahelb:
oh yes, now i see. that is a bug! thanks @moontrip

moontrip:
This is now addressed at https://github.com/VEuPathDB/web-eda/pull/1369

SheenaTomko:
This is fixed :)

Issue #1097, opened 2022-04-21
Implement new Axis Range Controls on viz tab


danicahelb:
We want to include (as part of the axis controls) a button per axis to zoom to the range of data actually being plotted.

after mockups are approved (VEuPathDB/web-eda#1100) we will implement the zoom to plotted data buttons

dmfalke:
I'm tentatively costing this as **medium**. We may need to adjust when mock-ups are done.

danicahelb:
Discussed in 7/25, 7/29, & 8/1 EDA UX meetings:

[Mockups (see page labeled "Axis controls")](https://www.figma.com/file/LVDoPrzKkd1rCYJgdgCCKC/New-viz-tab-buttons%3A-(1)-clear-parameters%3B-(2)-reset-to-defaults-%2F-zoom-to-plotted-data?node-id=127%3A500) 

<img width="797" alt="image" src="https://user-images.githubusercontent.com/25450900/182264159-593dab58-7866-441d-bb0e-079054c52170.png">

**_We will have radio buttons for Axis range options: `Full` vs `Auto-zoom` vs `Custom`_**

**Full range:** Default axis range selection: `Full` (user will need to actively select `Auto-zoom` or `Custom` to change the axis range)
- Full range will always includes the origin, even if the dataset does not include 0. 
- Full range **minimum** should either be (i) a negative number (if the minimum value in the full dataset or annotated range is negative) or (ii) zero (if the minimum value in the full dataset is 0 or a positive number) 
- Full range **maximum** will be the maximum value in the full dataset or annotated range + any buffer needed for error bars, Confidence intervals, etc

**Auto-zoom**
- Auto-zoom range will always include the origin, even if the user has applied a filter that would have removed 0 from the subsetted data
- If a user chooses `Auto-zoom`, the axis range will be be zoomed to the range where there is corresponding y-axis data for every x-axis point (and visa-versa)… 
- Note: auto-zoom range can be different from the range of the variable in the subset. For example, in LLINEUP, participants in the study are between 0-110 years of age but microscopy was performed only on kids 2-10 years. If I filter LLINEUP to include participants 0-50 years, plot Age vs microscopy, and then select `Auto-zoom` the x-axis would extend from 0-10 (keep the origin!) even though the subset ranges from 0-50.

**Custom range:** User can have full control over the axes ranges
- If user types into the axis range selector boxes, the radio button selection auto-updates to `Custom`

**_Range selector boxes_**
[...]

PR #1361, opened 2022-09-13
correcting dependent axis range of Lineplot Viz proportion

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1360.

Initially I tried to change useDefaultAxisRange hook but current solution is more intuitive and much simpler and works just fine.

Screenshots after fixing:

lineplot-pro-fix1

lineplot-pro-fix2

### DIFF
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -5,13 +5,7 @@ import LinePlot, {
 } from '@veupathdb/components/lib/plots/LinePlot';
 
 import * as t from 'io-ts';
-import {
-  useCallback,
-  useMemo,
-  useState,
-  useEffect,
-  useDebugValue,
-} from 'react';
+import { useCallback, useMemo, useState, useEffect } from 'react';
 
 import DataClient, {
   LineplotRequestParams,
@@ -62,6 +56,7 @@ import {
   NumberOrTimeDelta,
   NumberOrTimeDeltaRange,
   TimeDelta,
+  NumberRange,
 } from '@veupathdb/components/lib/types/general';
 import {
   LinePlotDataSeries,
@@ -641,13 +636,21 @@ function LineplotViz(props: VisualizationProps) {
     [data]
   );
 
-  const defaultDependentAxisRange = useDefaultAxisRange(
+  let defaultDependentAxisRange = useDefaultAxisRange(
     yAxisVariable,
     data.value?.yMin,
     data.value?.yMinPos,
     data.value?.yMax,
     vizConfig.dependentAxisLogScale
   );
+
+  if (vizConfig.valueSpecConfig === 'Proportion')
+    if (vizConfig.dependentAxisLogScale)
+      defaultDependentAxisRange = {
+        min: data.value?.yMinPos,
+        max: 1,
+      } as NumberRange;
+    else defaultDependentAxisRange = { min: 0, max: 1 };
 
   // custom legend list
   const legendItems: LegendItemsProps[] = useMemo(() => {
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -644,7 +644,7 @@ function LineplotViz(props: VisualizationProps) {
     vizConfig.dependentAxisLogScale
   );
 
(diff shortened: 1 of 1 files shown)


src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -641,14 +636,22 @@ function LineplotViz(props: VisualizationProps) {
     [data]
   );
 
-  const defaultDependentAxisRange = useDefaultAxisRange(
+  let defaultDependentAxisRange = useDefaultAxisRange(
```
bobular, line 639: I don't know how strict we are about using constants, but I guess this could be a `const` and then below we could reassign the values within, e.g. `defaultDependentAxisRange.min = ...`  But it would be more verbose

bobular (commented):
I'm not really sure how serious this "bug" is.  If I understand correctly, when no numerator(s) are selected, the plot cannot be made and therefore the user is not expecting to fiddle with the ranges. It is maybe a bit strange that the x-axis ranges are filled in (0 to 62) but the y-axis range is empty, but I could live with that (I honestly hadn't noticed before).

Upon applying this fix, the y-axis range is (0 to 1) for non-log mode and (BLANK to 1) in log mode. So there is still a blank remaining, which cannot really be fixed unless it is replaced with a generic 0.001 or something.

I think maybe this is a non-issue?  I'll link to this comment from the issue and get some input from the outreach folk.

bobular:
> I'm not really sure how serious this "bug" is. If I understand correctly, when no numerator(s) are selected, the plot cannot be made and therefore the user is not expecting to fiddle with the ranges. It is maybe a bit strange that the x-axis ranges are filled in (0 to 62) but the y-axis range is empty, but I could live with that (I honestly hadn't noticed before).
> 
> Upon applying this fix, the y-axis range is (0 to 1) for non-log mode and (BLANK to 1) in log mode. So there is still a blank remaining, which cannot really be fixed unless it is replaced with a generic 0.001 or something.
> 
> I think maybe this is a non-issue? I'll link to this comment from the issue and get some input from the outreach folk.

Oh, I see Danica had the same thoughts and then you corrected her! I stand corrected too.  Though I think it's more important to show the numbers **when there is data plotted** rather than before the numerator has been selected. 

https://github.com/VEuPathDB/web-eda/pull/1361#pullrequestreview-1110704284

moontrip:
@bobular Thank you for your review and comment! 👍 It is an easy fix so I just committed a change to cope with your suggestion: just added a condition to check whether data.value exists - this will guarantee that y-axis range is shown when data is available/plotted.

bobular (approved):
Good solution checking for data.value

I'll approve this.  I suppose having a variable variable instead of a constant "variable" that is used as a dependency for useMemo/callback/whatever means that we could accidentally change its value somewhere. If it's a `const` the compiler won't let us.

moontrip:
@bobular I left my thought at Slack where you made a discussion with Dave 👍

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
dmfalke: Just wanted to say that using `let` for this variable could be avoided by moving this logic into a function, which can use a `let` variable. It has the benefit of keeping the mutable variable out of the scope of the component function.

moontrip: @dmfalke Thank you for your comment! 👍 I also thought of that approach but didn't pursue it as I was not sure of making a function for that specific case only. I will try to do that.

moontrip: @dmfalke I have addressed your suggestion by making a custom hook that wraps the logic. Is it alright for you?

moontrip:
This may need to be overridden by the implementation of the new axis range control, https://github.com/VEuPathDB/web-eda/issues/1097 so it is converted to draft PR for now.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
dmfalke: @moontrip this looks great. (Sorry for the delayed response!)

dmfalke (approved):

moontrip:
@dmfalke no problem! Thank you for your review. Please note that this branch is closed because it is now incorporated at https://github.com/VEuPathDB/web-eda/pull/1369 in part.

PR #1362, opened 2022-09-14
FSM: handle constraints on provided overlay variable for floating visualizations

bobular:
Fixes https://github.com/VEuPathDB/web-eda/issues/1368

No longer requires web-components PR.

- [x] add snackbar when the overlay variable is automatically disabled by the new hook
- [x] unstratified plots should NOT use maroon. Use black (very dark grey?) until Ann is back!
- [x] barplot
- [x] histogram
- [x] lineplot
- [x] scatterplot
- [x] boxplot

### DIFF
--- a/src/lib/core/hooks/visualizations.ts
+++ b/src/lib/core/hooks/visualizations.ts
@@ -69,6 +69,9 @@ export function useVizConfig<ConfigType>(
  * Doesn't return anything!
  */
 export function useProvidedOptionalVariable<ConfigType>(
+  optionGetter:
+    | ((config: unknown) => VariableDescriptor | undefined)
+    | undefined,
   inputName: keyof ConfigType,
   providedVariableDescriptor: VariableDescriptor | undefined,
   /* storedVariableDescriptor is basically vizConfig.overlayVariable */
@@ -88,6 +91,7 @@ export function useProvidedOptionalVariable<ConfigType>(
   // if the new variable is compatible with variable constraints
   useEffect(() => {
     if (
+      optionGetter == null ||
       storedVariableDescriptor == null ||
       isEqual(providedVariableDescriptor, storedVariableDescriptor)
     )
--- a/src/lib/core/hooks/visualizations.ts
+++ b/src/lib/core/hooks/visualizations.ts
@@ -12,6 +12,7 @@ import {
   VariablesByInputName,
 } from '../utils/data-element-constraints';
 import { isEqual } from 'lodash';
+import { ColorPaletteAddon } from '@veupathdb/components/lib/types/plots';
 
 /**
  * Decodes config from back end or uses default config if there's an error (including no config at all)
@@ -134,5 +135,25 @@ export function useFlattenedConstraints(
       dataElementConstraints &&
       flattenConstraints(selectedVariables, entities, dataElementConstraints),
     [dataElementConstraints, selectedVariables, entities]
+  );
+}
+
+/**
+ * Simple hook to create the appropriate PlotlyPlotProps to make the plot markers grey
+ * when there's an externally provided overlay variable that is currently not enabled
+ * as an overlay.
+ */
+
+export function useNeutralPaletteProps(
+  overlayVariableDescriptor: VariableDescriptor | undefined,
+  providedOverlayVariableDescriptor: VariableDescriptor | undefined
+): ColorPaletteAddon {
+  return useMemo(
+    () =>
+      overlayVariableDescriptor == null &&
+      providedOverlayVariableDescriptor != null
+        ? { colorPalette: ['#333'] }
+        : {},
+    [overlayVariableDescriptor, providedOverlayVariableDescriptor]
   );
 }
--- a/src/lib/core/hooks/visualizations.ts
+++ b/src/lib/core/hooks/visualizations.ts
@@ -152,7 +152,7 @@ export function useNeutralPaletteProps(
     () =>
       overlayVariableDescriptor == null &&
       providedOverlayVariableDescriptor != null
-        ? { colorPalette: ['#333'] }
+        ? { colorPalette: Array(8).fill('#333') }
         : {},
     [overlayVariableDescriptor, providedOverlayVariableDescriptor]
   );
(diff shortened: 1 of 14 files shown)


dmfalke (approved):
I'm happy with this. I think the snackbar and tooltip message can be added in a follow-up.

bobular:
I just finished off this work but there's no rush to approve or test. It can wait until your return if necessary :-)

moontrip:
@bobular I have started to test your PR. At the very first test, I have noticed strange bugs concerning legend check/uncheck behavior - I have tested with two studies and both have the same issue so perhaps they are bugs. Thus, I just stopped testing other features and am reporting them here. Will probably do other tests sometime tomorrow.

- started with Histogram and found below issues
if selecting Overlay at FSM, plot histogram by selecting Main variable at floating viz, and Overlay radio button option at floating viz is set to be none. Under these conditions, if uncheck any legend, then (a1) somehow graph is also disappeared; (a2) it behaves an infinite loop, alternating checked and unchecked at legend automatically and the plot appears and disappears continuously; (a3) Overlay radio button option at floating viz is set to be an overlay chosen at FSM overlay (top right), then legend uncheck/check seem to work fine but 2) still occurs.

Note that Boxplot, Scatterplot, and Lineplot have the same behaviors with histogram above. Somehow barplot does not have (a2) and (a3) symptoms but (a1) still happens (when setting overlay none at floating viz).

bobular:
> @bobular I have started to test your PR. At the very first test, I have noticed strange bugs concerning legend check/uncheck behavior - I have tested with two studies and both have the same issues so perhaps they are general bugs. Thus, I just stopped testing other features and am reporting them here. Will probably do other tests sometime tomorrow.
> 
>     * started with Histogram and found below issues
>       if selecting Overlay at FSM, plot histogram by selecting Main variable at floating viz, and Overlay radio button option at floating viz is set to be none. Under these conditions, if uncheck any legend, then (a1) somehow plot is also disappeared; (a2) it behaves an infinite loop, alternating checked and unchecked at legend automatically and the plot appears and disappears continuously; (a3) Overlay radio button option at floating viz is set to be an overlay chosen at FSM overlay (top right), then legend uncheck/check seem to work fine but (a2) still occurs.
> 
> 
> Note that Boxplot, Scatterplot, and Lineplot have the same behaviors with histogram above. Somehow Barplot does not have (a2) and (a3) symptoms but (a1) still happens, so it may give you some hints?

Ooh, nasty bug, thanks!

bobular:
I noticed that the geohash grid wasn't enabled for FSM, so I have added it in this branch.

moontrip:
@bobular 
My two cents: I think you might have already known, but just in case, the following commit was what Dave did to interconnect between fsm and floating viz for legend behavior (check/uncheck). So, a conditional pass for the corresponding prop by checking whether the Overlay option at floating viz is none or overlay or something like that may solve the issues in part or full?

https://github.com/VEuPathDB/web-eda/commit/90e1ed3b8ba9ec7bf54e5ff53256077e0ceb4474

bobular:
> @bobular My two cents: I think you might have already known, but just in case, the following commit was what Dave did to interconnect between fsm and floating viz for legend behavior (check/uncheck). So, a conditional pass for the corresponding prop by checking whether the Overlay option at floating viz is none or overlay or something like that may solve the issues in part or full?
> 
> [90e1ed3](https://github.com/VEuPathDB/web-eda/commit/90e1ed3b8ba9ec7bf54e5ff53256077e0ceb4474)

Thank you, I am already looking into `useCheckedLegendItemsStatus` but that commit makes it clearer what is going on with the upstream wiring.

bobular:
Yay, that suggestion solves the disappearing plots, thank you @moontrip 

Now I will look at the cycling problem for another 20 minutes. I had hoped they were connected, but it seems not.
I also ruled out it being something to do with updated settings coming back from the back-end after a few seconds because if I comment out the following in FullScreenContainer
```
      analysisClient.setPreferences({
        [key]: state,
      });
```
it still cycles between checked and unchecked...

moontrip:
@bobular glad it worked, at least for that disappearing one :) For the cycling issue, as I mentioned before, only barplot does not have that symptom for whatever reason, so it may be helpful to figure out the issue?

bobular:
Hi @moontrip - yes I looked carefully at the differences between barplot and for example boxplot.

I found a [difference](https://github.com/VEuPathDB/web-eda/pull/1362/commits/adbe4269ce84ad14900cd7378914351c0acbe7bc), and fixing it solves the issue! 

However, I'm not quite sure how thumbnail updates cause the looping.  These are adding unnecessary load for the client for floating visualizations so I will add a tech debt ticket for disabling them.  That may also solve the #1367 problem. The timeframe seems similar!

moontrip:
> Hi @moontrip - yes I looked carefully at the differences between barplot and for example boxplot.
> 
> I found a [difference](https://github.com/VEuPathDB/web-eda/pull/1362/commits/adbe4269ce84ad14900cd7378914351c0acbe7bc), and fixing it solves the issue!
> 
> However, I'm not quite sure how thumbnail updates cause the looping. These are adding unnecessary load for the client for floating visualizations so I will add a tech debt ticket for disabling them. That may also solve the #1367 problem. The timeframe seems similar!

@bobular Wow I didn't expect that was the reason and quite interesting. Anyway I am pleased that it is resolved 👍 👍

src/lib/core/hooks/visualizations.ts
```diff
@@ -39,3 +50,115 @@ export function useVizConfig<ConfigType>(
...
+ * If vizConfig[inputName] is not null and the variable does not meet the
+ * constraints, then set vizConfig[inputName] to undefined (which has
+ * the effect of moving the radio button to "None".
+ *
+ * Also enqueues a snackbar. Make sure you are in a child of a SnackbarProvider.
+ *
+ * Doesn't return anything!
+ */
+export function useProvidedOptionalVariable<ConfigType>(
+  optionGetter:
```
bobular, line 72: Should this just be an `isEnabled: boolean` prop?

moontrip (approved):
recent commits by Bob fixed legend-related issues so I am approving this PR

src/lib/core/hooks/visualizations.ts
bobular: I think if it was a named prop, it would be OK to use a boolean, but these are unnamed, so booleans are a bit opaque. Leaving it as-is.

PR #1363, opened 2022-09-15
Add navbar to EDA dev workspace and improve link names

jernestmyers:
This is a trivial enhancement to `web-eda`'s dev environment. I often find it inconvenient to switch between studies or analyses. Thus, I added a navbar to the dev server's heading. Also note that I changed the first link's name from `EDA Workspace` to `My analyses`.

image


moontrip:
I like to have this as I often changed url directly to select other study 👍

PR #1364, opened 2022-09-16
Prevent access to unreleased studies if client is not configured for it

dmfalke:
Addresses https://github.com/VEuPathDB/ClinEpiWebsite/issues/60

This PR add protection against accessing datasets that are unreleased, from websites that are not configured to include such datasets.

### DIFF
--- a/src/lib/workspace/AnalysisPanel.tsx
+++ b/src/lib/workspace/AnalysisPanel.tsx
@@ -87,6 +87,7 @@ interface Props {
   /** API client that will be used in the Download Tab */
   downloadClient: DownloadClient;
   singleAppMode?: string;
+  showUnreleasedData: boolean;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -87,6 +87,7 @@ interface Props {
   /** API client that will be used in the Download Tab */
   downloadClient: DownloadClient;
   singleAppMode?: string;
+  showUnreleasedData: boolean;
```
bobular, line 90: Good that it's not an optional prop :+1: I don't have to check it's provided everywhere!

src/lib/workspace/WorkspaceRouter.tsx
```diff
@@ -62,6 +62,7 @@ export function WorkspaceRouter({
   sharingUrlPrefix,
   showLoginForm,
   singleAppMode,
+  showUnreleasedData = false,
```
bobular, line 65: safe default

bobular (approved):
Followed all the logic through (no actual testing) and all looks good.

jernestmyers (approved):

PR #1366, opened 2022-09-19
Proposed fix for double data requests

bobular:


dmfalke (approved):
I think this is good for now. We might want to consider moving this logic into a custom hook, since it could apply in other places (like the viz tab). It's also possible that we could improve some logic in the viz components to prevent the double data request. I don't have anything concrete to add, just some knee-jerk thoughts.

PR #1369, opened 2022-09-19
new axis range options

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1097, having new axis range option such as Full, Auto-zoom, and Custom

### WIP
- [x] Scatterplot Viz - X & Y
- [x] Lineplot Viz - X & Y
- [x] Barplot Viz - Y
- [x] Boxplot Viz - Y
- [x] Histogram Viz - X & Y
- [x] Date handling for dependent axis range, esp. for the Auto-zoom option, e.g., Scatterplot and Lineplot Vizs
- [x] Proportion - Barplot Y, Histogram Y, and Lineplot Y
- [x] Checking computed apps with Microbiome (such as alphadiv, abundance, etc.)
- [x] merge main branch


moontrip:
applied the new axis controls to Lineplot Viz and fixed bugs for Scatterplot Viz.

Since this new Lineplot Viz's axis control is related to the ticket, https://github.com/VEuPathDB/ClinEpiWebsite/issues/61, screenshots were made using the same example in the ticket.

- Y-axis Full option: metadata-based min/max (including 0)
lineplot-prism-control1

- Y-axis Auto-zoom option: data-based min/max  (including 0): this will resolve the issue in the ticket, I suppose
lineplot-prism-control2


- Y-axis Auto-zoom option for a Faceted plot
lineplot-prism-control3

moontrip:
Handling date for Auto-zoom option is addressed, which is for some Vizs like Scatterplot and Lineplot Vizs that allow date for dependent axis. Here are screenshots for Lineplot Viz example (GEMS1)

- Full option for dependent axis
lineplot-control-date1


- Auto-zoom option for dependent axis
lineplot-control-date2

moontrip:
applied to Barplot and Boxplot Vizs. Also considered Barplot Viz's proportion case. Screenshots with Barplot Viz:

- Barplot Viz, Proportion, Full option 
barplot-proportion-1


- Barplot Viz, Proportion, Auto-zoom option
barplot-proportion-2

moontrip:
histogram Viz now have new axis range controls for both x-axis and y-axis. Also proportion is considered. Just in case x-axis range is now computed, but honestly not quite sure if it is useful. Screenshots are:

- Full option for dependent axis, proportion mode
histogram-control-proportion1


- Auto-zoom option for dependent axis, proportion mode
histogram-control-proportion2

moontrip:
Considered Lineplot Viz proportion, which will also address https://github.com/VEuPathDB/web-eda/issues/1360. Screenshots are:

- Full option, no proportion is selected (thus no range is shown)
lineplot-proportion1

- Full option, proportion is selected (thus range is shown as 0 to 1)
lineplot-proportion2

- Auto-zoom
lineplot-proportion3

moontrip:
Tested computed apps with this PR such as alphaDiv, abundance, and etc. and found a couple of issues concerning log scale, axis range control, and decimal points. Now they are fixed - screenshots with alphaDiv and Scatterplot Viz:

- Full option, no log scale
alphadiv1

- Full option, log scale
alphadiv2

- Auto-zoom, no log scale
alphadiv3

moontrip:
finally merged main branch into this PR - significant conflicts and auto-merge messed things up further. From my extensive tests, they seem to work correctly.

bobular (commented):
Histogram looks pretty good. Just a few comments in the code.

I'll continue with testing the other plot types shortly.

I've noticed an unrelated bug (present in QA too) so will make that first!

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -451,12 +471,26 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
 
-  const defaultIndependentRange = useDefaultAxisRange(xAxisVariable);
+  const independentAxisMinMax = useMemo(
+    () => histogramDefaultIndependentAxisMinMax(data),
+    [data]
+  );
+
+  const defaultIndependentRange = useDefaultAxisRange(
+    xAxisVariable,
+    vizConfig.independentAxisValueSpec === 'Full'
```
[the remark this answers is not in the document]
moontrip: @bobular Thank you for your tests and comments 👍 Yes, when implementing the new axis range controls for histrogram x-axis, I realized that actually Full and Auto-zoom options did not mean much. The need to have those two tests for `vizConfig.independentAxisValueSpec === 'Full'` depends on whether we will distinguish Full and Auto-zoom options for histogram x-axis: Case 1) if we don't need to have different values for those two options, then we can simply use old logic `const defaultIndependentRange = useDefaultAxisRange(xAxisVariable);` and remove the new works. Case 2)  If we need to distinguish them, then we need new additions, `independentAxisMinMax` and `defaultIndependentRange` with the two `vizConfig.independentAxisValueSpec === 'Full'` conditions for `useDefaultAxisRange` hook. Note that without the two conditions, both Full and Auto-zoom may lead to data-based min/max, not metadata-based min/max, e.g., GEMS1 Age (for main variable) where metadata-based min/max is [0,62] while data-based min/max is [0,63] with the default bin width of 3.

Which one do you prefer?

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -755,6 +806,8 @@ function HistogramViz(props: VisualizationProps<Options>) {
         max: truncationConfigDependentAxisMax,
       },
     },
+    independentAxisValueSpec: vizConfig.independentAxisValueSpec,
```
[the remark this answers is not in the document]
moontrip: @bobular oh yes, they are not necessary. Somehow I did the same thing to other Vizs, so I just removed them from other Vizs too 👍

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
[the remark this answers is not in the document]
bobular: For me, the following is working fine for Histogram -> GEMS1 -> Age

Full goes 0 to 62, Auto-zoom goes 0 to max-of-my-subset.

```
  const defaultIndependentRange = useDefaultAxisRange(
    xAxisVariable,
    independentAxisMinMax?.min,
    undefined, // never logScale on this axis so no need for minPos
    independentAxisMinMax?.max,
    undefined, // never logScale - I would prefer `false` here
    vizConfig.independentAxisValueSpec
  );
```

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2185,3 +2283,44 @@ function isSuitableCategoricalVariable(variable?: Variable): boolean {
...
+    dependentAxisLogScale,
+    dependentAxisValueSpec
+  );
+
+  // include min origin: 0 (linear) or minPos (logscale)
+  if (data.value != null && valueSpecConfig === 'Proportion')
+    if (dependentAxisLogScale)
+      defaultDependentAxisRange = {
+        min: data.value?.yMinPos,
+        // in case data.value.yMinPos === 1, then use 1.0001 for max range for better display
```
bobular, line 2303: Does the 1.0001 avoid points being chopped off at the top? Don't we add padding already, somewhere else?

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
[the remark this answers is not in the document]
moontrip: @bobular Can you please check it out again? If we use it like your listed, then both Full and Auto-zoom goes to [0,63] (with the default bin width of 3) because [independentAxisMinMax.min, independentAxisMinMax.max] = [0,63] and Full option will compute max among metadata-based, data-based, etc. I am getting both [0,63] - just in case, please hard refresh your browser 😃 (CTRL + F5 key in Chrome) or check it at private mode :)

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: @bobular For Lineplot proportion, there is a case that both min and max are unity (1) (e.g., numerator = denominator, etc.), then the range becomes [1,1]. It plots okay without error but the y-axis tick labels become strange - so it is just for a better display purpose - you can see it without setting 1.0001 😃.

bobular: Oh, I see.  I mis-read line 2305 as checking the max, not min!

I have just been looking at this, and comparing with barplot and histogram.  Can they all use the same logic to handle proportion mode?  I see that barplot and histogram both have a bit of [cut-and-pasted similar code](https://github.com/VEuPathDB/web-eda/blob/5065d207a7e636a3b140eb1e17c944c7052b8f0c/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx#L477-L488) - could we add a `isProportion : boolean` prop to useDefaultAxisRange and handle everything in there?  It would be better for future maintainability.

Maybe we could deal with the `min === max` edge case in `web-components` extend-axis-range-truncations.ts?

moontrip: > I have just been looking at this, and comparing with barplot and histogram. Can they all use the same logic to handle proportion mode? I see that barplot and histogram both have a bit of [cut-and-pasted similar code](https://github.com/VEuPathDB/web-eda/blob/5065d207a7e636a3b140eb1e17c944c7052b8f0c/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx#L477-L488) - could we add a `isProportion : boolean` prop to useDefaultAxisRange and handle everything in there? It would be better for future maintainability.

@bobular Actually I tried to make it be common but it was not that simple task because: a) there exist many conditions (log scale, dataType, dataRange, etc.) and it adds more complexity; b) I cannot remember correctly, but it caused type issue. Due to those reasons, I deferred it and made things work at first. Perhaps technical debt item.
 
> Maybe we could deal with the `min === max` edge case in `web-components` extend-axis-range-truncations.ts?

Yeah it may be better to handle those edge case in there. Another tech debt item? :)

bobular: Great, let's get this merged relatively soon so we can get feedback on various functional aspects.

I've added a new [tech debt ticket here](https://github.com/VEuPathDB/web-eda/issues/1398).

bobular (approved):

PR #1386, opened 2022-09-23
Handle case where a dataset record's is_public attribute value is "NA"

### LINKED ISSUE #1387: User datasets are being treated as unavailable when the client is configured to not show unreleased data.
This is because we are currently checking if the value of a dataset's `is_public` attribute is not "true". Instead, we should check if the value is "false". This is because user datasets use the value of "NA". Such datasets will only be included in the response if the user owns the dataset, so we can treat them as "released" in this context.

This is related to https://github.com/VEuPathDB/web-study-data-access/issues/21

dmfalke:
Fixes #1387.

Since the `is_public` attribute of a dataset can have the value "NA", it's better to check if the value is "false", rather than is not "true", when determining if the dataset is unreleased.

A value of "NA" does strictly mean the dataset has been released, but for the context in which the value is being checked, we can assume the user has access to the dataset, since only user datasets have this value, and a user will only see such datasets if they are owned by the user making the request.

When this PR is merged, we should also make a PR for branch `v3.8.20-patch` and cherry-pick the commit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.9.0",
+  "version": "3.9.1",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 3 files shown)


dmfalke:
Since this is urgent, I merged this without review

PR #1394, opened 2022-09-29
Replace alphaDiv and rankedAbundance config select elements

jernestmyers:
The configurations for `alphaDiv` and `rankedAbundance` use native select elements. This PR replaces those native select elements with the new `SingleSelect` component from CoreUI.


src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -153,6 +155,23 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
...
+    selectedDataConfig?.entityDisplayName +
+    ' > ' +
+    selectedDataConfig?.displayName;
+
+  const [selectedDataValue, setSelectedDataValue] = useState('');
+
+  useEffect(() => {
+    if (!selectedDataValue.length) return;
+    changeConfigHandler('collectionVariable', JSON.parse(selectedDataValue));
+  }, [selectedDataValue]);
```
dmfalke, line 173: The `useState` and `useEffect` hooks seem unnecessary here. It looks like you are doing this because the value needs `JSON.parse()` to be applied to it, and the combination of this `useState` and `useEffect` hook allows you to use the `changeConfigHandler` function, while also using it for changing ranking method.

I think a preferable solution would be to do something like this:
```
<SingleSelect
  ...
  onSelect={flow(
    JSON.parse,
    partial(changeConfigHandler, 'collectionVariable')
  )}
  ...
/>
```
This follows the same pattern you're using for rankingMethod.

An even better solution would be to allow `SingleSelect` values to be anything, rather than just strings. Then, you wouldn't need to using `JSON.stringify` and `JSON.parse`. I can make an issue for this in coreui.

This all applies to the `alphaDiv` plugin, as well.

dmfalke (changes requested):
I think it would be good to remove the unnecessary use of component state.

jernestmyers:
This work now depends on https://github.com/VEuPathDB/CoreUI/pull/119, which changes the type of `SingleSelect`'s `value` prop to be `any`.

Issue #1400, opened 2022-10-05
use sentence case for plot mode options


danicahelb:
"Count" and "Proportion" Plot Mode options on the histogram and bar plot viz tools are currently all lower case. Please capitalize the first letter of both words.

<img width="215" alt="image" src="https://user-images.githubusercontent.com/25450900/194105991-ce97da20-ecb0-4ae9-a33e-2ed75874e606.png">

Issue #1127, opened 2022-05-06
Standardize behavior of `Reset to defaults` buttons across viz all tools


bobular:
In histogram it resets the bin width, but in lineplot it doesn't.  It should probably be consistent, but feedback from ClinEpi folk would be helpful.

danicahelb:
Discussed in 7/25 EDA UX meeting

Folks are generally happy with the placement & functionality of the reset button in the [mockups](https://www.figma.com/file/LVDoPrzKkd1rCYJgdgCCKC/New-viz-tab-buttons%3A-(1)-clear-parameters%3B-(2)-reset-to-defaults-%2F-zoom-to-plotted-data?node-id=10%3A154):
<img width="405" alt="image" src="https://user-images.githubusercontent.com/25450900/180867955-b2adca5f-01c4-47b8-8d43-fc32aa31d716.png">

@Ann Blevins and Styling WG should pick a better icon for the “reset to default” buttons

danicahelb:
reset to default button: https://github.com/VEuPathDB/CoreUI/issues/111

moontrip:
For Outreach/UI/UX team: Current plan is to use Undo icon only to replace Reset to defaults button as shown in https://github.com/VEuPathDB/web-eda/issues/1127#issuecomment-1194586363. But Bob and I think that icon with text looks okay as well (see below screenshot), thus it would be great if you can choose which one is better - the change in between icon without text and icon with text is quite simple.

![196177163-e5a0e519-6dc1-4952-9fa6-c4d3cae41345](https://user-images.githubusercontent.com/12802305/196244617-5e6f225c-1e5b-4510-98b6-93672aae2e43.png)

bobular:
Oh, I thought since you (@moontrip) fixed the tooltip width problem, we'd go with the text-less version?

I prefer that, at least.

moontrip:
@bobular yes the implementation is certainly icon **without** text (merged) as suggested earlier, but I wanted to ask the team if they prefer icon **with** text like above, just in case :)

sweverschulman:
plot reset is consistent across different plots

PR #1401, opened 2022-10-06
using reset to defaults icon etc

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/1127. Also, https://github.com/VEuPathDB/web-eda/issues/1400

List of works that was handled in this PR
- [x] replacing Reset to defaults button with a new icon and change its location
- [x] reset all controls to their default values: Lineplot bin-related controls are also reset
- [x] for Lineplot Viz, added disabled for all controls when the variable is categorical (e.g., ordinal)
- [x] change reset icon color to be radio button's color
- [x] use sentence case for plot option mode Also, https://github.com/VEuPathDB/web-eda/issues/1400

A screenshot with a Scatterplot for demo purpose:
scatterplot-reset1


src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -649,16 +657,42 @@ function BarplotViz(props: VisualizationProps<Options>) {
...
+            <div style={{ marginLeft: '-2.3em' }}>
+              <Tooltip title={'Reset to defaults'}>
+                <button
+                  onClick={handleDependentAxisSettingsReset}
+                  style={{
+                    width: undoSize,
+                    height: undoSize,
+                    background: 'none',
+                    border: 'none',
+                    padding: 0,
```
bobular, line 678: I played around with an extra `margin-left: 2px` in dev tools and I think it looks a bit less crowded.

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -846,11 +854,40 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
+
+          {/* set Undo icon and its behavior */}
+          <div
+            style={{
+              display: 'flex',
+              flexDirection: 'row',
+              alignItems: 'center',
+            }}
+          >
+            <LabelledGroup label="X-axis controls"> </LabelledGroup>
```
bobular, line 866: I see LabelledGroup is now childless :cry: 

Does it cause problems if we put all the controls inside it?

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -824,13 +827,18 @@ function HistogramViz(props: VisualizationProps<Options>) {
     </>
   );
 
+  // set Undo icon props
+  const undoSize = 20;
+  const undoColor = '#006699';
```
bobular, line 832: It would be good to start using the colors from the theme.

`core/components/FilterChip.tsx` is an example of how to do it.

I would put the undo button into its own component and then all the (minor) theme-related boilerplate is only needed once.

bobular (commented):
Looks great.
I recommend factoring the new button into a component and using the theme colors. It will be easier to maintain!

src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
moontrip: Thank you for your tests, @bobular 👍 I will consider it

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: yeah it is now childless... the reason why I separate it is that I need to place label and reset button (Undo icon) in a row, so I wrapped <LabelledGroup> and reset button with <div>.

src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: I tried to make it as a component, ResetButton. However, I have made four examples with Boxplot Viz to hear your opinion (https://github.com/VEuPathDB/web-eda/pull/1401/commits/fb3925d0d439b5f87e4afadb84fe1b8ee035cd92), so it would be great if you can test them. Basically it is comprised of three approaches, CoreUI, MUI (Material-UI), and original (previous one using pure button/Tooltip). In the following screenshot there are four kinds of button icons starting from CoreUI (icon only - without text), CoreUI (icon with text), MUI, and original. CoreUI one has mouse hover event (you can test it), and somehow icon with text is not that bad. One thing I don't like for the first one, CoreUI - icon only without text, is that it looks like its tooltip size depends on the button size so the tooltip text, Reset to defaults, are shown in the three lines: the others are okay.

It would be great if you can share your thoughts on which one would be best fit to our purpose.

boxplot-undo1

bobular:
I agree that the icon+text button looks OK - though maybe it doesn't need a tool-tip (unless that is a more wordy version).

I would vote for either of the CoreUI versions. The long texty one is good for now. (Some people may complain about the shouty capitalization?)  We could raise a ticket in CoreUI to address the tooltip dimensions/wrapping?  Or perhaps play with non-breaking spaces? Unfortunately the tooltip prop only accepts `string` so I couldn't provide HTML for the tooltip contents.

image

moontrip:
@bobular I made changes to use CoreUI approach (icon only for now) as mentioned above. I tested more carefully for wrapped tooltip, however, I noticed that it worked just fine at Storybook - added a tooltip and tested. So, I concluded that the icon width was not sufficiently long, even if it looked okay, thus I resolved the issue by increasing the width of the reset (undo) icon long enough - currently set to 50%. Now, the tooltip text is shown as single line and no visual issue is found even with the increase of the width.

tooltip1

bobular (approved):
Looks very nice.  :+1:

Issue #1367, opened 2022-09-19
Full screen map has dragging issues


bobular:
This seems to only happen when floating visualizations are open. I suspect it might be some kind of circular state (something coming back from the saved user analysis?) but I can't see an obvious culprit.

To reproduce (in `fsm-all`)
1. open FSM
2. put up a barplot
3. move the map
4. shortly after the barplot has finished updating, drag the map again
5. you should see the map jump back to where you started dragging (and the barplot will update again)

Tested also with boxplot - same issue (and assume same for all)

bobular:
Also, try cancelling the overlay variable (top right panel) at step 4 instead of dragging the map.  That also gets reset.

Wait a few seconds first and it's fine.

nkittur-uga:
No dragging issues seen.

PR #1403, opened 2022-10-06
Make thumbnail creation optional

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1367

Looks like for the FSM dragging issue, @bobular's intuition was correct---the thumbnail creation was the problem. Nice one, Bob :)

However the issue I experience with quick dragging the map viz seems to be a different problem; it's still happening for me. However I'm not sure anyone else has seen this issue, so maybe it's just a quirk from my environment....


moontrip (approved):
With this PR, I could not experience the issue Bob stated, so no thumbnail to FSM seems to be the solution. Other Vizs still generate thumbnail so this worked fine for me 👍

bobular (approved):
Problem solved!
Code looks good. Sorry I left BoxplotViz as the odd-one-out!

Issue #1397, opened 2022-10-03
Lineplot visualization: switching from non-date to date x-variable causes runtime exception


bobular:
We've had this [problem before](https://github.com/VEuPathDB/web-eda/issues/1061). It may have crept back in.  The solution is to put guards in place to prevent treating old values for axis ranges as dates.

aaronwlsong:
Looks it has been fixed again. Closing ticket.



![Image](https://user-images.githubusercontent.com/61077845/197025409-80db4f9e-9fae-4241-8ee8-f8742591e26d.png)
![Image](https://user-images.githubusercontent.com/61077845/197025408-a42c1084-35fe-4d20-a7b6-f20889f37a45.png)

aaronwlsong:
Looks it has been fixed again. Closing ticket.

![Image](https://user-images.githubusercontent.com/61077845/197025474-a13f0a3a-5e37-4c2b-b31b-8cc8c0fc86e5.png)

PR #1404, opened 2022-10-07
fix date issue for Lineplot Viz

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1397. 

TD;DL - it works fine with this PR

As Bob stated, there was a related work made before, https://github.com/VEuPathDB/web-eda/issues/1061, where I used Date.parse() to prevent from the error, but somehow that part disappeared - perhaps it was removed when working Viz related works. 

The error happens when switching from ordinal to date variable for xAxisvariable, e.g., for GEMS1 Case Study, switching from Age group to Birth date, etc. So, initially, I used similar approach like previous one (https://github.com/VEuPathDB/web-eda/issues/1061) using Date.parse(). However, I realized that the behavior of Date.parse() depends on web browsers, especially chrome: e.g., Date.parse('aaa') returns NaN but Date.parse('aaa 1') does not return NaN (only for chrome browser; firefox still returns NaN). To remove such edge case, I tried to use regEx to filter out a string starting with alphabet and it worked fine. Unfortunately, I found another edge case that some ordinal variable (e.g., Study year, etc.) has vocabulary that starts with number ('1', '1.5', ...) which Date.parse(numeric string) does not return NaN. So, made additional regEx to filter out the case. Then, another exception was found that that a numeric string of certain variable(s) looks like year, e.g., enrollment year that has '2007', '2008', ... 

I thought that there might be more cases not to satisfy all cases with Date.parse() approach, thus finally I relied on a library, `luxon`, which has already been installed at node_modules. By doing so, all cases are now resolved. Note that I had to install `@types/luxon` to avoid type error: added to `devDependencies`


bobular (approved):
This is nice and well worth the new library.

I thought we had other places we did the number/date/string safety check, but it looks like we have refactored them away already so this is good. (Unless they are in web-components?)

Anyway, I have added a tech-debt ticket to look into retiring the other date library we use.
https://github.com/VEuPathDB/web-components/issues/400

Issue #1408, opened 2022-10-11
independent behavior of X and Y axis range control and no origin for Auto-zoom


moontrip:
Two things to improve concerning the new axis range control:

a) As raised by Bob at Slack, he found that X and Y axis range controls were not independent: in other words, a change of one axis range control should not affect the other axis range. 

b) based on this decision, no origin is included for Auto-zoom option (https://github.com/VEuPathDB/web-eda/issues/1097#issuecomment-1273655548)

bobular:
I had previously said that x-axis bin width changes could possibly be allowed to reset the y-range (if custom), but let's see how it goes just relying on the yellow truncation bars as a warning.

sweverschulman:
I can confirm that x and y axis range control are independent of each other. 

![Image](https://user-images.githubusercontent.com/62298921/197240959-2d645c07-ade6-47da-b2bc-b5457be58f4a.png)


![Image](https://user-images.githubusercontent.com/62298921/197240970-190da3bf-8575-443e-8912-b84a3ca050ad.png)

Issue #1338, opened 2022-08-25
Tech debt: refactor multiple unnamed boolean arguments in viz's onChangeHandlerFactory callback


bobular:
These multiple unnamed boolean arguments are now getting a bit out of control/unreadable. 

At the same time, maybe we can move the snackbar enqueuement inside this handler and perhaps refactor it to a custom hook used by all viz's.  At the moment each viz is different because they handle different "interactions" between config props differently. Hopefully this can be generalised.

_Originally posted by @bobular in https://github.com/VEuPathDB/web-eda/pull/1332#discussion_r953817721_

PR #1409, opened 2022-10-11
independent axis range control btw x and y, no origin for auto-zoom

moontrip:
This will address two enhancements described in the ticket, https://github.com/VEuPathDB/web-eda/issues/1408

a) behavior of axis range controls between x and y are independent each other
b) for Auto-zoom option, origin (e.g., 0) is not included

As for b), below is the example case that Bob raised (https://github.com/VEuPathDB/web-eda/issues/1097#issuecomment-1265744498, https://github.com/VEuPathDB/web-eda/issues/1097#issuecomment-1273655548)
- X-axis Full
histogram-india1

- X-axis Auto-zoom: now the X-axis range is based on data, not including 0.
histogram-india2

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.10.5",
+  "version": "3.10.6",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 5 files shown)


bobular (approved):
Looks great, works great :+1: 

Adds more urgency to https://github.com/VEuPathDB/web-eda/issues/1338

moontrip:
Thank you for your review and tests, @bobular 👍

PR #1416, opened 2022-10-18
Use variable's field.term instead of name for SelectTree key

### LINKED ISSUE #1414: Variable picker: Can't switch between two same-named variables, e.g. GEMS1 dry mouth
This is a bug on my local dev at `main` and on QA, but not on live.

In a visualization, you cannot switch between the two different "dry mouth" variables after one has been selected (e.g. as a x-variable in a RxC mosaic).

Part of the issue may be that "dry mouth" exists as a value in a multi-variable at the Participant level, and a regular categorical variable at Participant repeated measures.  In subsetting they are displayed differently (intentionally) - so you can switch between one and the other.

UPDATE: actually the variable change IS communicated back to parent components, but the widget does not close automatically.

Could be of interest to @jernestmyers

jernestmyers:
Closes #1414

### DIFF
--- a/src/lib/core/components/variableTrees/VariableList.tsx
+++ b/src/lib/core/components/variableTrees/VariableList.tsx
@@ -722,7 +722,7 @@ export default function VariableList({
       }}
     >
       <SelectTree
-        key={dropdownLabel}
+        key={activeField?.term}
         {...sharedProps}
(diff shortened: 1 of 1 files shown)


bobular (approved):
Ooh, nice simple fix. I wish they were all like this!

Issue #1413, opened 2022-10-17
Scatterplot: inconsistent behavior when no overlay variable selected


danicahelb:
In a scatter plot with plot mode =  `smoothed mean with raw` or `best fit line with raw`, If I do not add an overlay variable, I cannot click off the data points or the lines / confidence intervals.

<img width="1002" alt="image" src="https://user-images.githubusercontent.com/25450900/196288321-a3054b41-1e66-4e1c-b12c-589c5a069b84.png">

However, if I add an overlay, i can click off values to remove all of the data points from showing up in the plot.

<img width="1030" alt="image" src="https://user-images.githubusercontent.com/25450900/196288541-2e9697cc-dc7a-492d-8e3c-efbaa104bf6c.png">

This behavior is inconsistent. Also, there are use cases where I may not have an overlay variable added but still want to click off all of the raw points. For example, LLINEUP... x-axis=`Age` & y-axis=`Plasmodium asexual stages, by microscopy result` with no overlay. The way the points are drawn prevents me from seeing that a best fit line has been drawn. I wanted to click the raw points off so that I could see the best fit line.

bobular:
Now this is fixed, we can see important trends!

We could have a think about whether the yellow truncation warnings should "respect" the checkbox selections.

![image](https://user-images.githubusercontent.com/308639/197749153-559f8597-080f-426d-8fc5-96911c522c47.png)

moontrip:
@bobular The legend is just to enable and disable each plot via plotly's built-in functionality, which does not involve data reloading, while truncation warning/data range are invoked based on new data loading. Thus, honestly, I don't know how to make your suggestion happen.

sweverschulman:
There is now consistent behavior between with or without an overlay. There is now the ability to check off values when using scatter plot with plot mode = smoothed mean with raw or best fit line with raw. 

![Image](https://user-images.githubusercontent.com/62298921/198369423-2b1e1f78-1471-46eb-a745-fbd54d95f82c.png)
![Image](https://user-images.githubusercontent.com/62298921/198369422-221d17a6-a637-47ad-b773-3eb86f1dc61c.png)
![Image](https://user-images.githubusercontent.com/62298921/198369424-f4ea191a-cc3a-4963-92ce-26b4b13601f8.png)

PR #1422, opened 2022-10-24
fix scatterplot Viz legend for single data

moontrip:
This addresses https://github.com/VEuPathDB/web-eda/issues/1413

After carefully checking the issue, I realized that an argument for useCheckedLegendItems hook was not correctly set, which was modified to cope with full screen map (FSM) before. Checked both Scatterplot Viz and FSM to work fine.

Here is a screenshot to disable data for the same example used in the ticket.
scatter-legend-bug-fix2


bobular (approved):
Good catch!

Looks great, but please could you also update the comment at line 899?  (the "no overlay" part is no longer relevant - maybe just remove it?)

moontrip:
@bobular Thank you for your review and test! 👍 I also removed the text so will merge this PR

Issue #1170, opened 2022-06-01
Subsetting: Categorical variable filters with very many values - check-all button only works for visible page when another filter is on


bobular:
Example UMSP -> Health Center -> District in Uganda

First set another filter on Clinic Visit -> Malaria diagnosis
![image](https://user-images.githubusercontent.com/308639/171433141-64301bde-bf1c-417e-b9f1-a1c647a13fe5.png)

Now in Health Center -> District in Uganda

Clicking on the top-left checkbox
![image](https://user-images.githubusercontent.com/308639/171432670-6ff5044f-1521-4165-8a56-bd3b66653f21.png)

Only selects districts A-K (actually 0 to K)

Also clicking on this again doesn't deselect anything.

When no other filter is active, "check all" box works as you would expect (districts A-Z)

dmfalke:
Related issue #1412

sweverschulman:
Currently waiting for USMP to be reloaded so that I can QA

bobular:
Hi Steph, you can QA with GEMS1 and the first variable that comes up when
you search for "reason".

sweverschulman:
In GEMS1 Case Control Study, variable: "Reason not evaluated after rehydration" if search within those variables for a term "baby" and select all. It only selects those terms that are found in the search. I do believe this confirms the correct functionality. 

@bobular mind confirming that above is correct.

bobular:
> In GEMS1 Case Control Study, variable: "Reason not evaluated after rehydration" if search within those variables for a term "baby" and select all. It only selects those terms that are found in the search. I do believe this confirms the correct functionality.
> 
> @bobular mind confirming that above is correct.

Yes, that's correct. Thank you!

PR #1428, opened 2022-10-26
Bump wdk-client version

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1170


dmfalke (approved):
The version of wdk-client used on websites will be updated separately from this. I do this on a periodic basis, and will do so tonight.

Issue #1330, opened 2022-08-19
Bug in hiding variable from variable tree


SheenaTomko:
The term EUPATH_0038405 (Cause of death, from autopsy) in SIP is supposed to be hidden in the variable tree but available on download. Right now it appears everywhere. Other terms like Village ENVO_01000773 and 'Kebele (subdistrict) in Ethiopia' EUPATH_0038722 have also been hidden from the variable tree in SIP and are working correctly. Not sure why this variable is being treated differently.

This study is not yet released and only on qa.restricted

dmfalke:
I believe the problem has to do with the logic for finding the root of the variable tree in this code: https://github.com/VEuPathDB/web-eda/blob/1f1972cbc765419ea2f59d201e6f1a887460ab19/src/lib/core/utils/wdk-filter-param-adapter.ts#L260-L294.

Note that traversal begins at the end of the function, with the variable whose parent ID is the same as the entity ID (i.e., its parent ID is not defined). No such variable exists for the entity **Participant Repeated Measures**. Thus, the variables of that entity are not being traversed. We need a more resilient way to find the root variable.

sweverschulman:
I can confirm that EUPATH_0038405 (Cause of death, from autopsy) is currently hidden on the variable tree but available for download

![Image](https://user-images.githubusercontent.com/62298921/200013432-23ac59d9-d567-44b0-bb8c-660a28247157.png)



![Image](https://user-images.githubusercontent.com/62298921/200013440-5f90b321-6ffc-42b3-a61e-155056b5749a.png)

PR #1429, opened 2022-10-26
Fix hidden variables

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1330

### DIFF
--- a/src/lib/core/utils/wdk-filter-param-adapter.ts
+++ b/src/lib/core/utils/wdk-filter-param-adapter.ts
@@ -210,6 +210,9 @@ export function entitiesToFields(
   scope: VariableScope
 ): Array<FieldWithMetadata> {
   return entities.flatMap((entity) => {
+    console.log('in entitiesToFields');
+    console.log({ entity });
+
     // Create a Set of variableId so we can lookup parentIds
     const variableIds = new Set(entity.variables.map((v) => v.id));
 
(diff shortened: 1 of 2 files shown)


dmfalke (changes requested):
This looks good! Could you add a short comment explaining what `rootParentIds` is, and how we decide what gets added to that array? Thanks!

src/lib/core/utils/wdk-filter-param-adapter.ts
```diff
@@ -286,8 +282,20 @@ export function makeHiddenVariablesInScope(
     });
   }
 
-  variablesByParentId[entity.id]?.forEach((variable) => {
-    _traverseDescendantVariables(variable, false);
+  const variableIds = new Set(entity.variables.map((variable) => variable.id));
```
chowington, line 285: Note that this is duplicated in `entitiesToFields`. Is it worth deduplicating this operation?

dmfalke: I think we should, eventually; but I think we will want to overhaul how we are handling variable trees in a larger context. You'll notice that we are doing a pseudo tree traversal of variables here, which means we have to re-implement the "parent logic" here.

An overhaul I've had in mind (but haven't had time to do, yet) is to construct the tree in the subsetting client, and to make that available as a part of the study metadata that the application code uses. I don't think there is an issue for this. I will make one in a bit and reference it here.

dmfalke: https://github.com/VEuPathDB/web-eda/issues/1434

dmfalke (approved):

Issue #1443, opened 2022-10-27
Histogram behaviour with respect to x-axis range change is inconsistent between subsetting and visualization


bobular:
Any change to the controls subsetting histogram controls causes a new back end request and so the bin width slider range is adjusted based on the response.

In the viz tool
a) at the time of writing, on the live site: changes to the x-axis range will trigger a new back end request (and the bin slider is adapted based on the response)
b) at the time of writing, the latest `main` branch client code does not trigger a new back end request (as described above)

I think the behaviour in both subsetting and viz should be as in a) above

Relates to https://github.com/VEuPathDB/web-eda/issues/1274

bobular:
Update: #1444 now fixes the b) behaviour above and reverts it to a)

danicahelb:
potentially related to https://redmine.apidb.org/issues/47693

I just talked with @d-callan about this, she commented:

> Adding bin slider specs to variable metadata so that the workflow determines optimal bin slider min, max and step values in the same way in the subsetting tab as in the viz tab is more complicated than originally anticipated. I'm not sure the minor differences between subset and viz tabs are worth the effort.

Issue #1274, opened 2022-07-18
Bin width & range selection interactions are buggy


danicahelb:
Appropriate bin widths are based on the selected data range. Behavior is currently buggy, and different between browse & subset and viz tab histograms

**Viz tab behavior:**

1. If the user sets the bin width and then changes the data range, the bin width selector box may turn red as a subtle indicator that an inappropriate width is being used based on the range of the data. For example: Histogram of Female anopheles count (LLINEUP) → change bin width to 100 and range to 0-200
- bin width selector box turns red and the plot has inappropriate bin width based on the selected range

2. However, if the user selects the data range 1st and then tries to set the bin width to an inappropriate number, behavior is different. For example: Histogram of Female anopheles count (LLINEUP) → change range to 0-200 and then try to change bin width to 100 
- 1st, even though I type “100”, the 2nd 0 is left out and the bin width selector box says “10” → the plot updates so that the bin widths are 10
- 2nd, I try to type the 2nd 0 into the box again → this time the bin width selector says “100” and the box turns red but the bin width is still at 10 on the plot

**Browse & subset behavior:**

1. If the user sets the bin width and then changes the data range, the range selector boxes show a different number than the range of the plotted axis. For example: Histogram of Female anopheles count (LLINEUP) → change bin width to 100 and range to 0-200
- range selector box indicates 0-200 but the axis is showing 0-300

2. However, if the user selects the data range 1st and then tries to set the bin width to an inappropriate number, behavior is different. For example: Histogram of Female anopheles count (LLINEUP) → change range to 0-200 and then try to change bin width to 101
- 1st, even though I type “101”, the 2nd 1 is left out and the bin width selector box says “10” → the plot updates so that the bin widths are 10
- 2nd, I try to type the 2nd 1 into the box again → this time the bin width selector says “100” and the box turns red but the bin width is still at 10 on the plot

bobular:
I'm also noticing that the bin width slider box turns red (but with no other warning) when the bin width is set to either of the extremes of the allowed range.
![image](https://user-images.githubusercontent.com/308639/193567169-57c94a21-4b9b-4432-b12e-cfb40104feda.png)
(SCORE S. mansoni Cluster Randomized Trial, Histogram, main=birth year

dmfalke:
[...]

PR #1444, opened 2022-10-27
Partially address histogram bin width UI bugs

### LINKED ISSUE #1450: Visualizations: histogram/lineplot zooming should re-bin unless binWidth has been entered
Study: India ICEMR Cohort
Filters: Fever in last 48 hours = YES

On QA currently the behaviour is:

Make a **new** viz histogram of **temperature** 
Click "Auto-zoom".
Notice no spinner and no new back end request.

The behaviour should be that the binning should be re-done, based on the data, UNLESS a binwidth has been manually inputted by the user.

bobular:
Resolves #1450 
(which also has QA notes)

Works partially toward https://github.com/VEuPathDB/web-eda/issues/1443 and https://github.com/VEuPathDB/web-eda/issues/1274

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.11.7",
+  "version": "3.11.8",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
--- a/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -471,6 +471,7 @@ function HistogramViz(props: VisualizationProps<Options>) {
       vizConfig.valueSpec,
       vizConfig.showMissingness,
       vizConfig.independentAxisRange,
+      vizConfig.independentAxisValueSpec === 'Full',
       studyId,
       filters,
       filteredCounts,
@@ -705,8 +706,8 @@ function HistogramViz(props: VisualizationProps<Options>) {
   const handleIndependentAxisSettingsReset = useCallback(() => {
     updateVizConfig({
       independentAxisRange: undefined,
-      binWidth: defaultUIState.binWidth,
-      binWidthTimeUnit: defaultUIState.binWidthTimeUnit,
+      binWidth: undefined,
+      binWidthTimeUnit: undefined,
       independentAxisValueSpec: 'Full',
     });
     // add reset for truncation message: including dependent axis warning as well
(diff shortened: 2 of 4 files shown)


moontrip:
@bobular Thank you for your work. IIRC, I made the change because I thought axis range throughout other Vizs purely controlled  what ranges are displayed in a plot, not related to data (e.g., data reloading).

bobular:
Hi @moontrip - I made lineplot only use the "annotated binwidth" when in "Full" mode and also updated the data-getting dependencies so that a change from Full to Auto-zoom or Custom would trigger a new data request.  I think the behaviour is good now.  A good combination to test with is LLINEUP Cluster Randomized Trial - lineplot - x=Age, y=Hemoglobin.

I will not address the fact that the default bin width (in Full mode) is 4.23 which is outside the allowed bin width slider range.  I am explaining this on https://github.com/VEuPathDB/web-eda/issues/1274 in a minute.

bobular:
PS. going to add the "only use annotated binwidth when in Full mode" to histogram!

moontrip:
@bobular I could confirm that the things you mentioned worked fine 👍 I also saw your comments at the ticket so I will leave this PR as is for the time being before any decision is made. But I could have approved this 😃

bobular:
> @bobular I could confirm that the things you mentioned worked fine +1 I also saw your comments at the ticket so I will leave this PR as is for the time being before any decision is made. But I could have approved this smiley

Thanks!  There's no UX discussion required for the changes made in this PR. So we could merge.  The thing that needs discussion is what to do when the response invalidates the user's binWidth choice. Currently we just outline it in red and do nothing. Re-sending a new request seems kind of clunky. Anway, discussion should be on #1274.

moontrip (approved):
@bobular Thank you for your explanation. Then I will approve then you can decide its fate 😄

bobular:
Thanks @moontrip 

@dmfalke I'm not sure what to do about QA for these fixes.

The "viz histogram back end request trigger-dependencies fix" fixes an issue we've only recently introduced onto the qa site and nobody has noticed it. This fix reverts to previous live site behaviour.

There's improved behaviour to supplement DK's axis mode work (histogram and lineplot).  Switching from Full to Auto or Custom will trigger a new back-end request (which in some cases will be necessary to re-bin the smaller x-axis range, other times it will make no difference).

There's also a small fix related to pressing the axis range reset button in histograms (it now sets bin width to `undefined` which gives us the desired default behaviour at the slightly cryptic [line 1326](https://github.com/VEuPathDB/web-eda/blob/9f4a53f355bd474846751bf2c6a2833cb68c12e6/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx#L1326). 

On reflection, I'll make an issue ticket that describes the thing I poorly described just above, with a way to reproduce it on QA currently.

PR #1446, opened 2022-10-28
Fix EdaDataService #190 failed QA

bobular:
Addresses failed QA in https://github.com/VEuPathDB/EdaDataService/issues/190


dmfalke (approved):
Looks like it's working!

PR #1451, opened 2022-10-31
Only request attributes that are defined for dataset records.

dmfalke:
This PR adds some logic to filter out attributes that are not defined for the WDK dataset record class. This is required to support https://github.com/VEuPathDB/EbrcModelCommon/pull/28

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.11.8",
+  "version": "3.11.9",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


aurreco-uga (approved):
the changes to the file data-element-constraints.ts seem unrelated but im sure they r fine

dmfalke:
> the changes to the file data-element-constraints.ts seem unrelated but im sure they r fine

Oops! I didn't mean to commit that. I may need to revert. I will check with Danielle.

PR #1454, opened 2022-11-01
Handle minNumValues var constraint

### LINKED ISSUE #1453: Variable with binary outcome cannot be chosen in 2x2 mosaic plot
In MORDOR study, the var study arm has two values, but it cannot be chosen in 2x2 mosaic, however, if I plot it in RxC mosaic plot meanwhile keep all other things the same, it works. 

In 2x2 mosaic, I cannot choose var:"study arm" for either x or y axis.
Screen Shot 2022-10-26 at 5 11 38 PM

In RXC mosaic, it works. 

Screen Shot 2022-10-26 at 5 30 32 PM

dmfalke:
Closes #1453

This depends on some backend changes (https://github.com/VEuPathDB/EdaCommon/pull/21, and https://github.com/VEuPathDB/EdaDataService/pull/219) to be able to fully test. They have not yet been merged.

In the mean time, these changes should not break any thing, since it is a new constraint. So, it is possible to test for regressions, if any one wants to take that on.

@d-callan if you can add a message here when the backend changes are merged, that would be great.

### DIFF
--- a/src/lib/core/utils/data-element-constraints.ts
+++ b/src/lib/core/utils/data-element-constraints.ts
@@ -149,6 +149,7 @@ export function filterVariablesByConstraint(
     constraint == null ||
     (constraint.allowedShapes == null &&
       constraint.allowedTypes == null &&
+      constraint.minNumValues == null &&
       constraint.maxNumValues == null &&
       constraint.isTemporal == null &&
       constraint.allowMultiValued)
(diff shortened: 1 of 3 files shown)


src/lib/core/utils/data-element-constraints.ts
```diff
@@ -352,6 +361,21 @@ export function mergeConstraints(
   );
 }
 
+export function mergeMinNumValues(
+  constraintA: DataElementConstraint,
+  constraintB: DataElementConstraint
+) {
+  const mergedMinNumValues = Math.min(
```
d-callan, line 368: Possibly this should be Math.max

d-callan (commented):
Just one thought, that when merging constraints we may want the max of the two specified mins.

dmfalke:
Yeah, good catch!

I can't help but think the merging concept is fundamentally flawed, but that's a different story

src/lib/core/utils/data-element-constraints.ts
dmfalke: fixed

d-callan (approved):

chowington (approved):
Looks good (in absence of testing)!

PR #1457, opened 2022-11-01
Adjust for new CheckboxTree default styles

jernestmyers:
This work reflects changes to the `CheckboxTree` API as a result of https://github.com/VEuPathDB/CoreUI/pull/125


dmfalke (commented):
Can we use a `title` attribute where the `Tooltip` component has been removed? This way we can still present the information to users.

dmfalke (approved):

Issue #1433, opened 2022-10-24
Line plots: default range for aggregated continuous Y-axis variables should be auto-zoom


danicahelb:
discussed in 10/24/2022 EDA UX meeting

Related to: https://github.com/VEuPathDB/ClinEpiWebsite/issues/61#issuecomment-1251253744

Currently, aggregated Y-axis values (arithmetic mean, median, geometric mean) on the line plot use the range of the raw data for their default range. This was done so that the y-axis range would be standardized, regardless of how binning was done or what the subset was. But it is problematic because the range of the raw data and the range of the aggregate data are quite different, resulting in plots that look like either a bug or no difference between the bins:

<img width="1400" alt="image" src="https://user-images.githubusercontent.com/25450900/197627850-fa1424a3-19fd-4cdf-b846-a5f0170fcf7f.png">

For aggregated Y-axis values, the default plot should be the `Auto-zoom` option. The `Full` range option should be disabled

To make it clear to the user that something special is happening on the y-axis (ie, that the data is being aggregated), we should:
1. Put the aggregation function on a separate line break from the variable label
2. Bold the font for the aggregation function

If users are not understanding why the 'Full' range option is disabled, we will add a context message in the future

danicahelb:
looks much better, thanks!

<img width="670" alt="image" src="https://user-images.githubusercontent.com/25450900/202047254-0e0ebc85-88f5-4827-b96a-b19f6b00cb6b.png">

moontrip:
@danicahelb Thank you for your tests and confirmation! 👍

PR #1460, opened 2022-11-03
lineplot y-axis range for categorical and non-categorical variable

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1433. 
Also, this depends on https://github.com/VEuPathDB/web-components/pull/406 for handling tooltip string with html tags. This actually works without web-components, but tooltip string would not be correct.

For non-categorical variable, dependent axis range is set to be 'Auto-zoom', while for it is 'Full' for categorical variable. Also, I found the way to style dependent axis label, which supports some html tags like `<b> and <br>`, so I did style it. As mentioned in  https://github.com/VEuPathDB/web-components/pull/406, tooltip would show those tags as string thus additional work was required at web-components.

Here are a couple of screenshots (used PRISM data):
- non-categorical y-axis variable, thus Auto-zoom is default. Also styling was made at dependent axis label and tooltip shows correct text without html tags
lineplot-prism1

- categorical y-axis variable, so Full is default. 
lineplot-prism2


src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -386,7 +386,12 @@ function LineplotViz(props: VisualizationProps<Options>) {
         independentAxisLogScale: false,
         dependentAxisLogScale: false,
         independentAxisValueSpec: 'Full',
-        dependentAxisValueSpec: 'Full',
+        dependentAxisValueSpec:
+          yAxisVar != null
+            ? isSuitableCategoricalVariable(yAxisVar)
```
bobular, line 391: I think you can use `categoricalMode` here (already in the dependencies)

bobular (approved):
Tested and approve with one suggestion to check out please.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: @bobular Thank you for your review, test, and suggestion! 👍 The reason why I have used `isSuitableCategoricalVariable` function directly instead of `categoricalMode` is the same reason to your previous implementation/comment, citing "we can't use the more generally scoped 'yAxisVariable' because it's based on vizConfig and is out of date" 😃 It was really helpful to implement my case as well 👍

bobular: Ahaha, oops. I was clever back then, I guess!

bobular: Can possibly remove `categoricalMode` from the deps?

moontrip: @bobular I removed the deps and still worked fine. Updated.

Issue #1195, opened 2022-06-08
Bug: addition of stratification variables resets other selections (like range)


danicahelb:
Bug: when I add a stratification variable, this resets the x-axis range graph parameters, so I have to go back to reset the range … this is annoying, and forces user to wait to reload again (doesn’t reset bin controls, which is good).

This is especially problematic when looking at large datasets like UMSP

bobular:
Poss relevant to https://github.com/VEuPathDB/web-eda/issues/1338

SAK22:
Looks good!

Issue #1295, opened 2022-08-08
Lineplot and histogram vizualizations: improve truncation warning behaviour


bobular:
### The issue

Currently, most plots only display axis truncation warnings when data from the *current subset* has been excluded by user axis ranges.

However, for lineplot (when binning) and histogram x-axes, there is a problem.

Study: India ICEMR Severe P. vivax and falciparum Cohort
Filters: filter on Temperature, so that just the middle values are present (e.g. 37 to 37.5)
Then make a histogram and adjust the ranges slightly inwards (but nowhere near the filtered range) - this is what we call "zooming".

![image](https://user-images.githubusercontent.com/308639/183443084-7ee621e7-837f-41a7-af29-319e3d1dd83f.png)

Note that the warnings are displayed even though no data was cut off.

### Implementation notes

We'll need to make a special unzoomed data request in order to get the true xMin and xMax of the filtered subset.  Though if we're not zooming, we should avoid making this twice.  We'll probably have to move the data request into a custom hook.

dmfalke:
Making a second request for the unzoomed data is not ideal. Could the data service tell us if we're leaving out data points (cc @asizemore)?

asizemore:
sorry i don't quite follow yet - do we know is different about the binned requests/responses (binned line and histo) and the non-binned requests/responses that cause the issue? I guess why do we not have the xMin and xMax of the filtered subset information for binned axes but we do in raw datapoint axes?

bobular:
Update!

As of today, the above example now has no yellow bars (my filter is slightly different, ignore that):
![image](https://user-images.githubusercontent.com/308639/202723784-b654d6ea-dfa5-4150-b954-b897858481e9.png)

But if we zoom in, we also don't get yellow bars (which ideally, we should)

![image](https://user-images.githubusercontent.com/308639/202724058-3b4c5739-ec78-4677-8aa8-8f5653d00118.png)

The problem is basically that the yellow bar check is made using min and max values determined from the back end response. When sending a `viewport` in the request to the back end (e.g. for zooming in) the back end (correctly) returns only data within that viewport.

There's an existing comment from me in LineplotVisualization.tsx along the same lines:

```
      // This useBinning edge case handling is a bit of kludge.
      // The same situation affects the histogram independent axis (all the time).
      // The problem is if you make a data request for an x-axis viewport (aka zoom to range), then
      // the xMin and xMax are calculated from the response, which is just for the requested
[...]

PR #1464, opened 2022-11-09
keep x-axis range when changing overlay variable

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1195.
Since x-axis range control exists for Histogram, Scatterplot, and Lineplot Vizs, this applied to those three Vizs. 
No screenshot is attached as it cannot demonstrate this work.


src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -1998,14 +2012,16 @@ function getRequestParams(
...
-  const viewport =
-    vizConfig?.independentAxisRange?.min != null &&
-    vizConfig?.independentAxisRange?.max != null
-      ? {
-          xMin: String(vizConfig?.independentAxisRange?.min),
-          xMax: String(vizConfig?.independentAxisRange?.max),
-        }
-      : undefined;
+  // also no viewport change regardless of the change of overlayVariable
+  const viewport = keepIndependentAxisRange
```
bobular, line 2016: I did some testing, and it seems that now there is never any `viewport` sent to the back end.  Compare this with the QA site and you'll see the 

I reverted this block of code and I get the desired behaviour, though there is an issue with missing yellow truncation bars, but that is a separate issue that I have just commented on in an old ticket: https://github.com/VEuPathDB/web-eda/issues/1295

I think we would be OK without the `keepIndependentAxisRange` state management?

moontrip: @bobular Thank you for your review and tests. Yes I intended to set not to change viewport conditionally due to two reasons: a) as you mentioned, it is due to truncation; b) another problem is that if we always use viewport but keep x-axis range, change custom range with overlay variable, and remove overlay variable, then the data itself is limited to the x-axis custom range. So, it is just not yellow truncation.

For an example of b), GEMS1 Case, Age (X), Weight (Y), Case or control participant (overlay). If one set x-axis custom range of 0 to 30, then it looks like this - looks fine for now:
case01

However, if I remove the overlay variable, then the data is limited to 0 to 30 because we asked the data from 0 to 30 via viewport. Also, "Reset to default" would not work because data was already changed to 0 to 30 range. This was one of the reasons why I didn't want to change viewport for the change of overlay variable.
case02

bobular:
So if we make a **new** lineplot like above, set the custom range to [0 to 30] and either refresh the page, or go back to the main viz tab and click on the thumbnail again, the viewport is sent to the backend and then if you click on the **x-axis range = Full** radio button, the viewport doesn't get updated see below). Please could you set the viewport for all requests (or at least for all custom range requests) and then I'll review/test again.

image

moontrip:
> So if we make a **new** lineplot like above, set the custom range to [0 to 30] and either refresh the page, or go back to the main viz tab and click on the thumbnail again, the viewport is sent to the backend and then if you click on the **x-axis range = Full** radio button, the viewport doesn't get updated see below). Please could you set the viewport for all requests (or at least for all custom range requests) and then I'll review/test again, and I'll pay more attention to the behaviour with relation to changing the overlay variable.
> 
> image

@bobular Thank you for your detailed tests and finding this issue 👍 I think that I know the reason for this. In this PR, the viewport depends on `keepIndependentAxisRange` and it is defined in `useState()`. Thus, if we destroy the current Viz like doing such behaviors you described, then we lose the value of `keepIndependentAxisRange` (getting back to its default, `false`). If it is `false`, then the viewport relies on `vizConfig.independentAxisRange`, which became the Custom value, [0 to 30],  and this was used for viewport when reloading the Viz. To resolve this issue, we will need to store `keepIndependentAxisRange` to `vizConfig` as well. I will work for that.

moontrip:
@bobular It is now set to store `keepIndependentAxisRange` to `vizConfig` so the issue you raised will be resolved with the new update. 🤞 It would be great if you can test it again.

bobular:
Hi @moontrip 

Through some combinations of clicks I'm still able to get it into this state

image

And I can also reproduce the "binning still on when it is switched off".  See next comments

bobular:
So, the "binning still on" is easily reproduced.
Make a new x:Age y:Weight lineplot, switch to custom x-range, set limits to `0 to 30`, turn binning on and then off again.

The binning remains active. I think it's something to do with the default/fallback setting of `binWidth` here. We need to provide `value: 0` in the `binSpec` request but I think `binWidth` from https://github.com/VEuPathDB/web-eda/blob/a81742b4552497a4392d1bd904b3069035fb624c/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx#L1972-L1978 is `undefined`.  It's possible the fallback at https://github.com/VEuPathDB/web-eda/blob/a81742b4552497a4392d1bd904b3069035fb624c/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx#L2003 can include `value: 0` but it would need more testing.

It's possible this bug was not a result of your work - I haven't chased all the git blames sorry!

bobular:
Update: these two issues (binning-stays-on and full range x-axis not always respected) I am reporting were not introduced in this work. They are present in `main` and on live and QA.  I think it is probably something to do with data request dependencies. I will make a new ticket. Apologies for not checking this sooner.

@moontrip please could you undo [these changes](https://github.com/VEuPathDB/web-eda/pull/1464/commits/a81742b4552497a4392d1bd904b3069035fb624c) as they may have introduced additional complexity.

moontrip:
@bobular As I mentioned in https://github.com/VEuPathDB/web-eda/issues/1544#issuecomment-1372428548, I tried to apply the method here with a new commit. When you are available, can you please check if it is alright for you?

bobular:
I've tested and I don't think that's the right UX (if I choose x-range 0 to 30 and then switch on binning, I think I want the x-range to stay at 0 to 30).  Also I'm still seeing the binning stay on when switched off, but that's not the focus of this ticket.

I think we should simply address the goals of the linked [ticket](https://github.com/VEuPathDB/web-eda/issues/1195) and leave these new issues to the new ticket. I have a hunch that we don't need to use state or stored vizConfig to solve this. I tried to play around but couldn't easily roll this back to early commits in this PR because of the back end changes.  I'll try again tomorrow.

moontrip:
@bobular Thank you for your tests and explanations 👍 From my tests, everything seems to work just fine if I set `const viewport = undefined;` (after reverting the latest change for resetting independentAxisRange when toggling useBinning) - honestly, I am not quite sure of the role of the viewport, especially with the axis range control.

moontrip:
@bobular As you could see, the `keepIndependentAxisRange` is only used for controlling viewport behavior. So if we don't need to use viewport like I said above, then keepIndependentAxisRange can be removed. The `vizConfig.keepIndependentAxisRange` only targeted for the case of going out of the plot and coming back to it (e.g., minimizing plot, etc.), i.e., remembering its state.

bobular:
> I am not quite sure of the role of the viewport, especially with the axis range control.

The viewport is essential information to the back end when doing binning on a custom x-range. If you "zoom in" with a custom x-range, then the back end can handle smaller bin widths without overloading the network/client, but only if it knows the custom x-range aka viewport.

bobular:
> The `vizConfig.keepIndependentAxisRange` only targeted for the case of going out of the plot and coming back to it (e.g., minimizing plot, etc.), i.e., remembering its state.

But we already remember `vizConfig.independentAxisRange` which is all we need to remember.

Isn't the solution to "keeping x-axis rang when changing overlay variable" simply confined to `handleInputVariableChange`?  I think I would rename `keepBin` to `keepIndependentAxis` and make sure that `independentAxisRange` and other independent axis values are maintained during the variable change.

In fact, the focus of the [ticket](https://github.com/VEuPathDB/web-eda/issues/1195) on stratification variables only is misleading.  If the yAxis variable changes I am pretty sure it's OK to keep all the xAxis settings.

moontrip:
@bobular Thank you for your comments! 👍 I left my answers next to your comments.

> > The `vizConfig.keepIndependentAxisRange` only targeted for the case of going out of the plot and coming back to it (e.g., minimizing plot, etc.), i.e., remembering its state.
> 
> But we already remember `vizConfig.independentAxisRange` which is all we need to remember.
> 

DK: If https://github.com/VEuPathDB/web-eda/issues/1544 is fixed, then perhaps we don't need to use vizConfig.keepIndependentAxisRange. Either way, I don't know about the role of the viewport, especially with the presence of independent axis range control: doesn't the independent axis range control play the same role with the viewport? Another question is whether we need to have a dependency of independentAxisRange for data request: Histogram and Lineplot Vizs are the cases, while Scatterplot Viz does not do that. Honestly, I cannot find a specific merit for asking new data with the change of independent axis range: perhaps it is related to binning but I could not find specific difference with or without data dependency of independent axis range control.


> Isn't the solution to "keeping x-axis rang when changing overlay variable" simply confined to `handleInputVariableChange`? I think I would rename `keepBin` to `keepIndependentAxis` and make sure that `independentAxisRange` and other independent axis values are maintained during the variable change.
> 
> In fact, the focus of the [ticket](https://github.com/VEuPathDB/web-eda/issues/1195) on stratification variables only is misleading. If the yAxis variable changes I am pretty sure it's OK to keep all the xAxis settings.

DK: Perhaps you are correct that x-axis range does not need to be changed by the change of y-axis variable either.

bobular:
> Either way, I don't know about the role of the viewport, especially with the presence of independent axis range control: doesn't the independent axis range control play the same role with the viewport? Another question is whether we need to have a dependency of independentAxisRange for data request: Histogram and Lineplot Vizs are the cases, while Scatterplot Viz does not do that. Honestly, I cannot find a specific merit for asking new data with the change of independent axis range: perhaps it is related to binning but I could not find specific difference with or without data dependency of independent axis range control.

Did you see my explanation [here](https://github.com/VEuPathDB/web-eda/pull/1464#issuecomment-1373626028)?

It's critical to send a new data request when binning is active - the data endpoints expect the viewport (to which we pass the x-axis range).

Anyway, my advice is as above: strip out any new state/config management (`vizConfig.keepIndependentAxisRange`) and do exactly what your PR says: "keep x-axis range [settings] when changing overlay [or any other non x-axis] variable".

So renaming `keepBin` to `keepIndependentAxisSettings` would be great. There's no need for new state or `vizConfig`, trust me!  Also remember that if you want to preserve `vizConfig` settings in `handleInputVariableChange` you can simply leave those keys out of the `updateVizConfig` call.

moontrip:
> 

@bobular Sorry that I missed your previous comment: perhaps it was due to web browser's refresh issue. Yes, I have been working to remove vizConfig.keepIndependentAxisRange and etc. as you suggested last week. Will get back to you after done.

moontrip:
@bobular I have addressed your suggestions in the latest commit. It would be great if you can check them out and let me know if you find anything I missed. 👍

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -365,12 +372,15 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
...
-        // set independentAxisRange undefined
-        independentAxisRange: undefined,
+        independentAxisRange: keepIndependentAxisRange
+          ? vizConfig.independentAxisRange
+          : undefined,
         dependentAxisRange: undefined,
         independentAxisLogScale: false,
         dependentAxisLogScale: false,
-        independentAxisValueSpec: 'Full',
+        independentAxisValueSpec: keepIndependentAxisRange
```
bobular, line 381: Are we not going to keep `independentAxisLogScale` here too?  (two lines above)
Might be best to use the same name for the variable: `keepIndependentAxisSettings`.

bobular (changes requested):
Looks and works great - just one query.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: @bobular good point! 👍 I think it can also keep its setting so will change it. Also, I realized that I did make a wrong name for the replacement of keepBin - will fix it too.

moontrip: @bobular changes are made!

moontrip:
Thank you for your detailed reviews, feedbacks, and tests, @bobular 👍

PR #1468, opened 2022-11-11
Fix client to handle new style visualization plugin responses (variable metadata mainly)

bobular:


src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -1043,22 +1045,26 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
       ? ({
           entityId: computedYAxisDetails?.entityId,
           variableId: computedYAxisDetails?.variableId,
-          displayName: data.value?.computedVariableMetadata?.displayName?.[0],
+          displayName: 'foo', // data.value?.computedVariableMetadata?.displayName?.[0],
```
bobular, line 1048: Not sure what this part of the code does.

The back end doesn't seem to return `computedVariableMetadata.displayName`.

Boxplot has the same code.  I will investigate!

d-callan: Idk if it's relevant yet, but some computed variables we will provide display names for. For ex in beta div the two axes of the scatter plot have display names that depend on the results of the compute.

bobular: Turned out that the `displayName` prop was off-spec and unused. I removed it.

bobular:
Ready for review.  I will just merge in main for you guys.

I've tested it minimally. It's painfully slow working with a local back end transatlantically so I prefer to test again when the back end is merged/rolled out.

d-callan (approved):

dmfalke (approved):
Looks good. Haven't tested, but we can suss out issue on qa.

PR #1479, opened 2022-11-15
Handle inputs without constraints

dmfalke:
This PR handles cases where an input does not have an associated constraint

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.12.3",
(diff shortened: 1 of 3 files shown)


dmfalke:
It's not clear to me why this became an issue with the new constraint logic, but it could be related to us no longer merging constraints. I will investigate this soon, but my hunch is that it is related to us no longer flattening constraints.

chowington (approved):
Thanks for taking care of this, Dave!

bobular (approved):

Issue #1481, opened 2022-11-15
Bug: Scatterplot has legend issue on QA site


danicahelb:
There is now a bug with the legend when an overlay variable is added to the plot. See this GEMS1 example.

**On the live site it looks fine**

<img width="944" alt="image" src="https://user-images.githubusercontent.com/25450900/202033718-9af0b401-1d74-41ad-82d5-ef6390709e6d.png">

**On the QA site we are getting a gradient color map but the colors don't match what is on the plot and this is a categorical variable**

<img width="980" alt="image" src="https://user-images.githubusercontent.com/25450900/202033790-ad59b01e-2166-493b-9b2f-9466253286d5.png">

SheenaTomko:
Looks great now!

Issue #1399, opened 2022-10-05
Scatterplot: allow continuous overlay variables (needs to optionally support gradient colormaps)


d-callan:
mbio beta diversity app will need a scatterplot w a gradient colormap. but i dont think (@danicahelb could confirm) that clinepi wants gradient colormaps? so our scatter viz needs to be able to support them, but also have them be optional.

my (possibly flawed) memory from ann is that the components necessary to build a gradient colormap are available in web-components, but i havent yet confirmed.

danicahelb:
clinepi would be happy to use gradient colormaps to allow continuous vars to be used as the overlay on scatter plots

danicahelb:
prioritize for immediately following b60 release since mbio needs this for beta div

bobular:
@d-callan - are we envisaging always using binning for continuous overlay vars?

d-callan:
nope. its a true gradient

d-callan:
![image](https://user-images.githubusercontent.com/53306535/198362908-0cc49d04-522b-42dd-9fb8-f7095094ee8b.png)

an ex of the goal, though w our own color palette(s). id imagine to start w have a diverging one like in the ex above that is applied if the metadata indicates the range crosses 0, and one thats not. it wont be perfect but gets us a decent start.

d-callan:
also see https://github.com/VEuPathDB/web-eda/issues/306 for prev thinking

d-callan:
id also be fine to leave numbers w <9 values using a categorical palette for now if we dont want to tack on discretizing one of our continuous palettes dynamically yet. can make that a separate issue.

bobular:
Thanks for the extra info. I see Ann did a lot, but it would likely be a challenge to merge the client work now. Good for reference though.

Is the back end (plot.data) ready to go for continuous (unbinned) overlay, but the constraints for the passthrough scatterplot app do not allow it? Connor will be looking into this. I guess while testing he can spoof or ignore the constraints temporarily?

d-callan:
Yea plot.data is ready, and I can get constraints in the data service on a branch somewhere without too much of a headache I think. @chowington when do you need that by?

chowington:
I haven't gotten anything spun up yet---still soaking up previous work, so no rush!

chowington:
@d-callan I might be ready to start testing this by tomorrow, FYI!

d-callan:
[...]

Issue #306, opened 2021-08-13
Scatterplot: add gradient colormap for continuous overlay variables


asizemore:
------------------
***NOTE: This is now a duplicate of https://github.com/VEuPathDB/web-eda/issues/1399***
------------------

## Tasks

- [x] Pick sequential gradient map. Decided on the darkest 90% of colors from [oslo colormap](https://www.fabiocrameri.ch/colourmaps/). 
- [x] Gradient colormap legend.
- [x] Handle how no data interacts with gradient colormap. Plan: put the No Data in it's own, non-clickable, slot in the legend. Add prop to send hasMissingData to colormap legend and it can handle adding a no data line.
- [x] plot.data handle missing data. Needs to return a new seriesGradientColormap named "no data" or similar.
- [x] data service updates. 
- [x] web-eda general implementation
- [x] web-eda gradient colorscale turns off trendlines
- [x] Use annotated displayRangeMin and max if given (look at `defaultDependentAxisRange`). If not given, use rangeMin and rangeMax.
- [x] web-eda handle the dual overlay var constraints.

## Associated PRs
- web-components [#275](https://github.com/VEuPathDB/web-components/pull/275)
- plot.data [PR 124](https://github.com/VEuPathDB/plot.data/pull/124)
- web-eda [PR #790](https://github.com/VEuPathDB/web-eda/pull/790)
- data service [PR #138](https://github.com/VEuPathDB/EdaDataService/pull/138)

## Open questions

- [ ] Markers? Filled circles?
- [ ] Pick diverging map. IP
- [ ] How to handle "discrete" vars? If we let integers with < 9 unique values in, do they get categorical, gradient, or discrete? 
- [ ] Var constraints more generally. Right now set up to use var info to determine how the viz runs (web-eda). Are we happy sending all numeric (for example) vars to go through series gradient colormap? If var has values = 1:7, they could still get a seriesGradientColormap but then when sent to the legend are legend-ized appropriately.

## To Dos for later cards

- [ ] Allow dates
- [ ] Add controls for user to select colormap, range, midpoint (maybe future phase?)

danicahelb:
This does not work for ClinEpi

**I cannot select a continuous overlay variable**


![Image](https://user-images.githubusercontent.com/25450900/202043478-baa86b99-7522-4a59-a376-b0f19b19ecbc.png)

**When I choose a categorical overlay variable with numeric values I (incorrectly) get the gradient legend but the colors don't match what is being plotted**


[...]

PR #1484, opened 2022-11-16
Check that scatterplot response has colormap data before showing colormap legend

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1481
Fixes https://github.com/VEuPathDB/web-eda/issues/306
Fixes https://github.com/VEuPathDB/web-eda/issues/1399

### DIFF
--- a/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -588,8 +588,9 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
       let gradientColorscaleType: string | undefined;
 
       if (
-        overlayVariable?.type === 'integer' ||
-        overlayVariable?.type === 'number'
+        'seriesGradientColorscale' in response.scatterplot.data[0] &&
+        (overlayVariable?.type === 'integer' ||
+          overlayVariable?.type === 'number')
       ) {
         const defaultOverlayMin: number =
           overlayVariable.distributionDefaults.displayRangeMin ||
@@ -2068,6 +2069,7 @@ function processInputData<T extends number | string>(
   let dataSetProcess: any = [];
 
   // drawing raw data (markers) at first
+  // Nasty 'any' type here...
   responseScatterplotData.some(function (el: any, index: number) {
     // initialize seriesX/Y
     let seriesX = [];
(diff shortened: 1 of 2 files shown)


src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -588,8 +588,9 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
       let gradientColorscaleType: string | undefined;
 
       if (
-        overlayVariable?.type === 'integer' ||
-        overlayVariable?.type === 'number'
+        'seriesGradientColorscale' in response.scatterplot.data[0] &&
+        (overlayVariable?.type === 'integer' ||
+          overlayVariable?.type === 'number')
```
chowington, line 593: Is checking the first series enough/correct, or should I check all series?

dmfalke: I'm guessing it would be a bug if they weren't all the same, in this regard. If we have some assurance of upstream of this code, then I think only checking the first is fine.

That being said, if `response.scatterplot.data` is an empty array, a runtime error will be thrown here.

The safest thing might be to do something like this:

```typescript
response.scatterplot.data.every(t => 'seriesGradientColorscale' in t)
```

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -2068,6 +2069,7 @@ function processInputData<T extends number | string>(
   let dataSetProcess: any = [];
 
   // drawing raw data (markers) at first
+  // Nasty 'any' type here...
   responseScatterplotData.some(function (el: any, index: number) {
```
dmfalke, line 2073: Why is this `any`?

dmfalke (commented):
See my comments.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
bobular: That sounds like a plan from Dave.

My deep contribution is to add a blank line after this `if {}` block just before
```
      const overlayVocabulary = computedOverlayVariableDescriptor
```

bobular (approved):
I +1 Dave's suggestion, otherwise all good.

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
chowington: I agree with both suggestions. Thanks y'all! :)

Issue #1172, opened 2022-06-03
Update UI for data accessibility information on downloads tabs


danicahelb:
Discussed in 6/3/2022 EDA UX meeting. We would like to make improvements to the way the data accessibility information is presented on the downloads tab (which was initially implemented in https://github.com/VEuPathDB/web-eda/issues/1143).

See mockups: https://www.figma.com/file/mDKIOrpnSwwMjz9acRg3qC/Add-data-accessibility-information-(study-details-%26-downloads-tabs)

This ticket needs sign off in a UX meeting before implementing

danicahelb:
Final mockups are here: https://www.figma.com/file/mDKIOrpnSwwMjz9acRg3qC/Add-data-accessibility-information-(study-details-%26-downloads-tabs)?node-id=47%3A365

<img width="836" alt="image" src="https://user-images.githubusercontent.com/25450900/172906902-7ebb044b-ebeb-42b1-ae07-b3454805334c.png">

**To do:**
- [ ] Create a header for "Data Accessibility" using same font styling as for "My Subset" and "Full dataset"
- [ ] Place data accessibility sentence under the "Data Accessibility" header of 
- [ ] Update text explaining data accessibility (see below)

**_For users who don’t have access permissions or are not logged in:_**

For studies with public access:
> Data downloads for this study are public. Data is available without logging in.

For studies with controlled access:
> To download data please register or log in and `request access`. Data will be available immediately following request submission.

For studies with protected access:
> To download data please register or log in and `request access`. Data will be available upon study team review and approval.

**_For users who are logged in and have access to the study data:_**

For studies with public access:
> Data downloads for this study are public. Data is available without logging in.

For studies with controlled access:
> You have been granted access to download the data.

For studies with protected access:
> You have been granted access to download the data.

_Note: `request access` is a link that should take you to the popup:_
<img width="668" alt="image" src="https://user-images.githubusercontent.com/25450900/172907637-27a3eed0-aa37-489b-8c62-f24da47e5fe9.png">

nkittur-uga:
It is looking good except for one small thing: there's a line between the "Study accessibility" row and the text below, which should be removed. 

<img width="602" alt="Screen Shot 2022-11-15 at 9 57 08 AM" src="https://user-images.githubusercontent.com/67710160/201951867-5c0bf32b-b528-4b88-91f1-09aee43b8600.png">

SheenaTomko:
the space has been fixed

PR #1485, opened 2022-11-16
Remove margin from header element

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1172

(addresses https://github.com/VEuPathDB/web-eda/issues/1172#issuecomment-1315433250)

For some reason I wasn't able to see the extra margin noted in that comment---it only appears on QA, not my dev site

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.12.5",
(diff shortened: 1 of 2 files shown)


dmfalke:
There might be a global style rule for `h5` that isn't included in local dev sites.

dmfalke (approved):

PR #1492, opened 2022-11-23
FSM: Least ancestral entity logic

### LINKED ISSUE #1491: Stand alone map overlay not working for parent entities of geo-located entity
This is only demo-able on in the alpha popbio megastudy: https://feature.vectorbase.org/vectorbase.feature/app/workspace/analyses/DS_4b8cb8bb2e

In full-screen map, try to overlay with a study variable, like Study identifier...

Gives 2 x 500s on the back end.

The `outputEntity` should be Collection site, but we are sending Study to the back end.

bobular:
Fixes #1491


dmfalke (commented):
Is there any concern when the entities are siblings? It looks like we will end up preferring the xaxis variable, in that case.

bobular:
Sorry about the assignment mix-up, and thanks for the review.

That's a great question about sibling entities.

We'd have to imagine a hypothetical entity diagram, like below.  Collection site is asterisked because it is the "geo-entity".
```
        collection site* - collection - sample
study < 
        publication
```

I think we have been assuming that the x-axis variable's entity and the geo-entity lie on the same branch. In fact the map overlay plugin's constraints specify this too: https://github.com/VEuPathDB/EdaDataService/blob/master/src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapMarkersOverlayPlugin.java#L52

In the regular map viz, we pass these constraints (`pieConstraints` and `pieDependencyOrder`) to the `InputVariables` component.

In the FSM it seems that we don't, and this needs to be fixed (by me!) - I can make a ticket tomorrow, or maybe we can include it in this issue+PR?

So when that is done, we'll have a situation where sibling entities should never be compared with the `leastAncestralEntity` function, but it will be implied rather than explicit in the code.  Would comments be sufficient in this case?

dmfalke:
A comment is probably adequate, although it wouldn't hurt to check and throw an error if the entities are not on the same branch. I'll leave that up to you :smile:

bobular:
Hi @dmfalke - not urgent but I've added a sanity check as suggested.

dmfalke (approved):

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -462,16 +459,31 @@ export function leastAncestralVariable(
...
+
+    if (
+      variables.every(
+        (v) =>
+          isEqual(v, variable) || ancestors.find((e) => e.id === v.entityId)
+      )
+    )
+      return variable;
+    else
+      throw 'Error: variables not expected to be in different branches of entity diagram';
```
dmfalke, line 485: I just noticed this. You should throw an instance of an Error object, eg `throw new Error(message)`

dmfalke (changes requested):
Sorry, just caught something

bobular:
Impossible to test this "in anger" until we have a study where the geo-entity isn't the root.
The PopBio megastudy will be that soon.
Should we wait until we can test before merging?

dmfalke:
How do you feel about writing some unit tests for this?

bobular:
I feel that that would be the right thing to do in the circumstances.

dmfalke:
Not that I'm aware of. I think we can get away with entities with no variables here, so maybe it wouldn't be too hard to create some simple test data?

bobular:
Dave and I decided to test this with data rather than unit tests (jest/babel saga will wait until after monorepo)

feature site is back online - but map-markers is hanging, but I can confirm that the output entity logic is working because the entity-related errors that you see on feature.vectorbase.org mega-study when selecting **Study name** as the overlay variable, do not appear on my dev site (using feature site services).

So I'm going to merge!

Issue #1487, opened 2022-11-17
Visualizations: scatter plot opacity slider 


d-callan:
The introduction of the gradient colormap is making it more important we have a means other than hollow markers to address over plotting. The proposal is a unique type of opacity slider which uses filled markers, except on the single condition where the marker is fully  transparent, then it'd switch to the hollow marker.

danicahelb:
I made a duplicate ticket (now closed). adding info from that ticket here in case it is helpful

Discussed in 11/16 data viz mtg

currently scatterplot uses open circles, but circle outline colors are difficult to distinguish. As gradient color map gets implemented, it is increasingly difficult for users to differentiate hues

Add an opacity slider to the plot controls.
- Markers should be filled unless the fill is completely transparent, then switch to hollow markers.
- Default should still be hollow markers, but users can use the slider to add the fill color to increase the colored area, allowing for opaque filled markers will help address overplotting of filled markers

asizemore:
<img width="555" alt="Screen Shot 2023-01-12 at 12 48 57 PM" src="https://user-images.githubusercontent.com/11710234/212141403-1a97c24f-9a0b-4b63-ba25-3dd7ed02345a.png">

Just saw this! Very exciting!
The setting is "0" when we first land on the viz. Marker opacity should be 1 by default. Opacity 1 = markers fully opaque, where opacity=0 means markers are totally transparent. We have the right idea just swap the direction of the slider :)

danicahelb:
since the markers have some amount of transparency, the legend should also have that amount of transparency so that the colors actually match up.

danicahelb:
opacity slider is really nice, thanks @moontrip!

<img width="800" alt="image" src="https://user-images.githubusercontent.com/25450900/215197419-20445cbb-3083-4639-9896-a5edbdcbf109.png">

PR #1494, opened 2022-11-30
marker opacity for scatter plot viz

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1487. Relying on https://github.com/VEuPathDB/web-components/pull/419. This also works with overlay and facet variables: tested but no screenshot is attached. 

Note that currently, the step of opacity is set to be 0.1.

Here are some screenshots:
- a general marker with opacity of 0
 scatter-viz-opacity0

- a general marker with opacity of 1
scatter-viz-opacity1

- a gradient marker with opacity of 0
scatter-viz-gradient-opacity0

- a gradient marker with opacity of 1
scatter-viz-gradient-opacity1


src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -2246,6 +2276,15 @@ function processInputData<T extends number | string>(
             seriesGradientColorscale?.length > 0
               ? markerSymbolGradient
               : markerSymbol(index),
+          // need to set marker.line for a transparent case (opacity != 1)
```
bobular, line 2279: The `opacity: 0.7` a few lines above here is being applied to the final plot, so when the user chooses 1 on the slider, they actually only get 0.7.

The solution may just to remove this line.

moontrip: @bobular Thank you for your suggestion. I will remove it! 👍

bobular (changes requested):
It looks/works great! The only things missing are
a) opacity should be stored in vizConfig, not local state (unless that was explicitly not in the spec)
b) opacity should be added to thumbnail dependencies

moontrip:
> It looks/works great! The only things missing are a) opacity should be stored in vizConfig, not local state (unless that was explicitly not in the spec) b) opacity should be added to thumbnail dependencies

@bobular Thank you for your valuable suggestions! 👍 I missed those parts at Viz. Will work for that, probably tomorrow and after done, I will certainly use "re-review request" icon 😉

moontrip:
> It looks/works great! The only things missing are a) opacity should be stored in vizConfig, not local state (unless that was explicitly not in the spec) b) opacity should be added to thumbnail dependencies

@bobular I have addressed your suggestions. Since the corresponding web-components PR was already merged, you can test this PR without using npm-pack-here. But please don't forget to update your local repo (git pull origin scatterplot-opacity & yarn install) 😉

bobular (changes requested):

bobular:
Also, would it be useful to consolidate `markerColorOpacity` and `markerOpacity` into one name?  Perhaps `markerBodyOpacity` or `markerCenterOpacity`?

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -1220,7 +1232,7 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
       ? plotSpacingOptions
       : undefined,
     // need to define markerColorOpacity for faceted plot
-    markerColorOpacity: markerColorOpacity,
+    markerColorOpacity: vizConfig.markerOpacity,
```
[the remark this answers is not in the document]
moontrip: @bobular Ah, yes. You are absolutely right. I will start working on this PR after the other work you commented/suggested, including new spec such as two banners etc. 👍

moontrip:
> Also, would it be useful to consolidate `markerColorOpacity` and `markerOpacity` into one name? Perhaps `markerBodyOpacity` or `markerCenterOpacity`?

@bobular thank you for your suggestion. Perhaps markerBodyOpacity may work? I will need to change it at web-components as well 😄

bobular:
> > Also, would it be useful to consolidate `markerColorOpacity` and `markerOpacity` into one name? Perhaps `markerBodyOpacity` or `markerCenterOpacity`?
> 
> @bobular thank you for your suggestion. Perhaps markerBodyOpacity may work? I will need to change it at web-components as well 😄

Oh, that's a pain re. web-components.  Maybe find a consistent name on the EDA-side that means no changes over there?

moontrip:
> markerBodyOpacity

@bobular well, the current prop name, markerColorOpacity, is defined in web-components so it should be changed in there anyway. It may not be a big deal, except you will have one additional review request 😄

moontrip:
@bobular I have addressed your feedbacks. Since the change of the prop name required to make a change at web-components, this PR now depends on https://github.com/VEuPathDB/web-components/pull/433.

moontrip:
@bobular update. Due to fixing a bug for marker opacity work at web-components, https://github.com/VEuPathDB/web-components/pull/433 is now merged owing to Jeremy. So from now one this PR does not have a dependency with web-components.

bobular (approved):
This looks good - I'm going to approve without testing because I'm having trouble getting my dev instance up, and will focus my testing on #1464 when that is sorted

moontrip:
@bobular Thank you for your reviews 👍

Issue #1503, opened 2022-12-03
update line plots to keep user selections & NOT draw plot when selection combinations are incompatible


danicahelb:
Line plot currently uses snackbars to inform the user that log scale wasn't appropriate (and gets switched off) when error bars were switched on. Scatter plots have been updated to use content messages for this.

Please update the line plots to use content message for consistency with scatter

danicahelb:
Change: 
`Log scale is no longer appropriate and has been disabled`
to:
`Log scale and binning are not available concurrently. Why?`
and:
`Log scale and binning are not available concurrently. Binning of non-log transformed raw data cannot be accurately plotted on log scale axes. Read less`

Change:
`Error bars are no longer appropriate and have been disabled`
to:
`Y-axis log scale and error bars are not available concurrently. Why?`
and:
`Y-axis log scale and error bars are not available concurrently. Error bars for non-log transformed raw data cannot be accurately plotted on log scale y-axis. Read less`

danicahelb:
we do not want to leave a context message permanently after it is called. it should only be visible when the user selections are incompatible. the current behavior of scatter is what we want for line plots.

bobular:
Hi @danicahelb I'd just like to clarify on behalf of @moontrip 

In lineplot, when a user makes an "illegal" toggle selection, do we want the plot to disappear and the banner to show until the user corrects their "mistake"?  This is what happens in scatter.

danicahelb:
As of 1/13/2023, the content messages are being used appropriately.

However, when a user makes an "illegal" combination of selections, we do not want to auto-change their selections. We just want the content message to appear and the points/lines/data on the plot area to disappear.

currently, we are still auto-changing the user selections so that the plot can be drawn.

I'm changing the title of this ticket to make what has to happen still more clear

danicahelb:
I made a line plot and added both binning and log scale for the x-axis. The appropriate snack bar showed up and no data appeared on the plot.

However, i clicked on the "Restore to defaults" button, which switched the toggles for both binning and log scale off. Data points reappeared on the plot, however the snack bar did NOT disappear


![Image](https://user-images.githubusercontent.com/25450900/221004757-c1618307-b0b5-420e-bfdf-cd49f18a8f5b.png)

Please ensure that the snackbar only shows when selections are incompatible

danicahelb:
same issue is happening with the y-axis controls:



[...]

Issue #1502, opened 2022-12-03
Reserve space between plot area & controls for content message


danicahelb:
Plot controls are not in the same position after the content message appears.

The content message currently pushes all of the plot controls down when it appears. Please instead reserve space for banner under Scatterplot (and also under all other viz tools to accommodate future content messages)

We do not want there to be an oversized white space gap under the plot when there is no content message, so please make the content message less tall and remove any gaps between the content message and the plot controls

danicahelb:
in 10/21/2022 EDA UX meeting we had decided that: 

> There can be more than 1 plot control banners
> if 2+, the banners will be stacked. 

I suggest we start by leaving space for 2 banners & see how it looks

moontrip:
@danicahelb I am afraid that I am confused about the behavior of the banner, so please correct me if I am wrong. Do you want to leave a banner permanently after a banner is called? Currently, the banner behavior is like this: a) make a banner disappear if an addition change of a plot control by user is irrelevant to banner. For an example with Lineplot Viz, a banner is shown if user toggles on the log scale control while the error bar toggle is on. Under this condition, if user toggles off the log scale control, the banner disappears; b) make a banner be replaced with another banner because additional user action to trigger a new banner is unrelated to the previous action. So we currently show single banner depending the state of user's actions and those actions are irrelevant each other.

For reference, speaking of a possible number of banners, there are five banners for Scatter plot Viz and four banners for Lineplot Viz. The former cases are: 1) no smoothed mean; 2) X log scale and Smoothed mean; 3) Y log scale and Smoothed mean; 4) X log scale and Best fit; 3) Y log scale and Best fit. The latter cases are: 1-2) while X log scale on,  toggle on Bin, and vice versa; 3-4) While Y log scale is one, toggle on Error bar, and vice versa. Assuming that we will always keep banners, e.g., a user took actions of 1) and 2) with Scatter plot, which will have two banners, then what should we do if user will take another action of 3)? Removing the banner 1), moving up/down the banner 2), and showing a banner 3) to the empty place? If this is what you want, it is not that simple task to implement those behaviors because certainly current banner does not have that feature: it is designed for a single banner at each case. Honestly, I cannot estimate exactly how long it would take to make them happen.

danicahelb:
Hey @moontrip 
(banner is wrong verbiage, we should use context message. i corrected my post above) 

[...]

PR #1505, opened 2022-12-07
banner spacing for Vizs

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1502 and https://github.com/VEuPathDB/web-eda/issues/1503, but the latter, Lineplot Viz is still in progress, thus this is currently a draft PR. Note that this relies on https://github.com/VEuPathDB/CoreUI/pull/131. 

**UPDATE 1: now CoreUI version (https://github.com/VEuPathDB/CoreUI/pull/131) is up so no dependency requires**
**UPDATE 2: due to adding timeout, this PR now relies on https://github.com/VEuPathDB/CoreUI/pull/134**

When https://github.com/VEuPathDB/web-eda/issues/1503 is ready, this PR will be switched to a regular PR.

Some screenshots with a Scatterplot Viz are shown in the following:
scatter-log-banner0

scatter-log-banner1


moontrip:
Lineplot Viz now has banner instead of snackbar. A screenshot is attached:

lineplot-banner

bobular (commented):
It looks great, but it's implied that the expanded message should never take up more than two lines.  If there's a longer `additionalMessage` (as I have simulated below for scatterplot) then there are some layout problems. I assume this can be handled with `min-height` rather than `height` but I haven't looked into the CSS at all.

image

I tried to test in Lineplot and found a [bug](https://github.com/VEuPathDB/web-eda/issues/1523) with categorical x-vars - that is not present in `main` - but I assume a merge from main will fix that.

Testing lineplot with non-categorical x-variables: the banners work fine, but the temporary nature of the snackbars was a good thing. Now they hang around until either a) the offending toggle has been reversed (e.g. cancelling y-axis logscale gets rid of "Error bars are no longer appropriate and have been disabled") or b) creating a banner from the other axis, and then reverting the toggle so that message disappears.

It seems a bit inconsistent.  Could we consider timeouts?  I'll add a topic for Wednesday.

bobular:
Yes, that all sounds wise!

moontrip:
@bobular I have updated this banner work, which now has timeout: currently it is set to be 5 sec (5000 ms). It would be great if you can re-review this. Also changed styles as we discussed before.

chowington:
Hi @moontrip and @bobular, I reviewed the timeout change and I think it works for the line plot's log scale/error bar case, but it's not appropriate for the scatter plot's smoothed mean/best fit line/log scale case. That settings combination causes the plot to not show up until the settings are changed, so the content message needs to stay rather than disappear so the user always knows why the plot isn't showing up

moontrip:
@chowington Thank you for your observation! As discussed in yesterday meeting, we won't use banner timeout for Vizs, so banner will always show up. For banner component itself (CoreUI), I think we can keep because it it is am additional feature for banner.

chowington:
Ok, that sounds good, DK!

moontrip:
@bobular I have updated this PR by addressing the feedbacks from Danica and you. Since the corresponding CoreUI work was merged, this PR no longer needs to use npm-pack-here. 

Here are screenshots (GEMS1 lineplot viz):

- vertically centered when only one banner exists (your request at Slack, https://epvb.slack.com/archives/C0132PM1MT5/p1671190173467599?thread_ts=1671053277.443879&cid=C0132PM1MT5 😃)
lineplot-banner1

- two banners (Danica's request) - banner messages are changed following https://github.com/VEuPathDB/web-eda/issues/1503#issuecomment-1353693069
lineplot-banner2

chowington (approved):
Looks great, @moontrip!

moontrip:
@chowington thank you for your tests and reviews! 👍

bobular (changes requested):
One tiny issue: the plot controls seem to shift downwards by one pixel (Firefox, Linux) when the second banner is displayed for lineplot. They may need a tiny bit more space or less padding.

One larger issue: the behaviours of scatterplot and lineplot are not consistent.

Scatterplot, which I believe has the correct/desired behaviour: user makes a "mistake" the banner is shown, no plot is shown, the banner goes away when the user corrects their mistake.

Lineplot is currently allowing the user to make a mistake, another toggle is changed to compensate, and a banner is shown. That banner can be removed by clicking again on the last toggle, but the user already has a plot in front of them and may not want to change the plot they already have.  (I understand that fixing this is possibly more than "low cost", though I see that #1503 doesn't have a cost.)

Also, with Scatterplot, when I ask for x-axis logscale I get the "Data may have been truncated" warning, which is a bit misleading because there is no data nor are there any yellow warning bars.

image

bobular:
> Also, with Scatterplot, when I ask for x-axis logscale I get the "Data may have been truncated" warning, which is a bit misleading because there is no data nor are there any yellow warning bars.

Hi @moontrip 

I looked further into this minor issue. I guess it is something we didn't catch from the previous work on this functionality, not the current PR.

I see a few minor code issues: one is that `showLogScaleBanner` is a dependency of the `data = usePromise` but is not used inside that block.  Instead there's a bit of [code duplication](https://github.com/VEuPathDB/web-eda/blob/9de12ab715ef2d4232c4d1e8703d1e4a0781cb2c/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L501-L506).  It would be good to use `showLogScaleBanner` here instead.

For a quick fix we can probably add `&& !showLogScaleBanner` to the condition at https://github.com/VEuPathDB/web-eda/blob/9de12ab715ef2d4232c4d1e8703d1e4a0781cb2c/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L1519   or maybe even the more generic `&& data != null` which would handle more cases.

moontrip:
> > Also, with Scatterplot, when I ask for x-axis logscale I get the "Data may have been truncated" warning, which is a bit misleading because there is no data nor are there any yellow warning bars.
> 
> Hi @moontrip
> 
> I looked further into this minor issue. I guess it is something we didn't catch from the previous work on this functionality, not the current PR.
> 
> I see a few minor code issues: one is that `showLogScaleBanner` is a dependency of the `data = usePromise` but is not used inside that block. Instead there's a bit of [code duplication](https://github.com/VEuPathDB/web-eda/blob/9de12ab715ef2d4232c4d1e8703d1e4a0781cb2c/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L501-L506). It would be good to use `showLogScaleBanner` here instead.
> 
> For a quick fix we can probably add `&& !showLogScaleBanner` to the condition at
> 
> https://github.com/VEuPathDB/web-eda/blob/9de12ab715ef2d4232c4d1e8703d1e4a0781cb2c/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L1519
> 
> or maybe even the more generic `&& data != null` which would handle more cases.

Thank you for your identifying this, Detective! 😃 I will look at them next week.

bobular:
> DK: It may be more UI/UX discussion before taking an action. But Danica wanted to keep showing a banner if user does not fix it and stacking banners (X and Y). So I think that current behavior is what she wants?

I'm not sure that's exactly what she wants, but I've asked for clarification here:

https://github.com/VEuPathDB/web-eda/issues/1503#issuecomment-1373960117

moontrip:
@bobular I have tested your findings and tried to fix them: a) linux firefox spacing: I could check it out so I have added a bit more spacing, so it seems to be resolved for me; b) truncation warning: yes it left a truncation warning in some cases. Thus, I have added `data.value != null` (`data != null` was not sufficient) condition at https://github.com/VEuPathDB/web-eda/blob/9de12ab715ef2d4232c4d1e8703d1e4a0781cb2c/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx#L1519

bobular (approved):
Hi @moontrip 
I have removed the "resolves 1503" so we can address the behavioural issue when Danica gets back to us (though I'm pretty sure she'll want no plot when the user has selected an illegal combination). Currently she has more important things to worry about (in California).

The truncation stuff works great now, and the spacing seems fine in my FF. Thanks!

moontrip:
@bobular Thank you for your review and tests 👍 Yes, if that no plot is needed, then we can probably handle it at new ticket.

Issue #1506, opened 2022-12-07
Bug: Unexpected backend type(s) encountered in line plot


danicahelb:
To recreate (QA.clinepi): 
Study: SCORE S. mansoni CRT
Filter: Village study arm = Arm 6: Gaining-shsh
Line plot viz tools with the following parameters selected

<img width="1402" alt="image" src="https://user-images.githubusercontent.com/25450900/206062572-b941ed0f-558c-407c-a236-b03cb6b1476b.png">

aaronwlsong:
Recreated the case, looks good now. Closing ticket. 


![Image](https://user-images.githubusercontent.com/61077845/207441939-ff8e0b60-70df-489e-ae33-1e7e49678c6b.png)

moontrip:
@aaronwlsong thank you for your tests and confirmation! 👍

PR #1507, opened 2022-12-08
address type issue for Lineplot Viz

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1506.

After extensive tests, I realized that the backend did not return binSlider, binSpec, and viewport anymore for ordinal variables, which caused the error in the ticket. Changed type({}) to partial({}) for relevant type definition and the error is gone.

Here is the screenshot for the same Study and variables (the presence of the filter does not matter):
lineplot-bug0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.13.5",
+  "version": "3.13.6",
   "dependencies": {
     "debounce-promise": "^3.1.2",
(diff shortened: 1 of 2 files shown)


dmfalke (approved):
Nice find!

moontrip:
@dmfalke Thank you for your quick review and test! 👍

dmfalke:
I will make sure this makes it into build 61. Thanks!

Issue #1411, opened 2022-10-13
Bug: Line graph shows a data marker even when no data exists for that point


nkittur-uga:
Issue identified by users during the TMRC workshop (10/12/2022)

To recreate: 
SCORE S. mansoni Cluster Randomized Trial
Filter: Country=Tanzania, Village study arm= Arm 3 
Make a Line Plot with the parameters as in screenshot below

<img width="1215" alt="Screen Shot 2022-10-13 at 1 45 30 PM" src="https://user-images.githubusercontent.com/67710160/195668535-07a134a2-d0a2-4721-b256-a4f7a37466da.png">

Notice that the line plot shows markers at Y=0 for Year 3 and Year 4. In fact, there was no data for these years as these were treatment holiday years (see screenshot below showing no participants for these years).

<img width="544" alt="Screen Shot 2022-10-13 at 1 45 12 PM" src="https://user-images.githubusercontent.com/67710160/195668531-cdc8ae4e-3708-4430-a47a-734b3f1c04c4.png">

There should be no markers when there is no data.

danicahelb:
Discussed in 12/2/2022 EDA UX meeting. Markers at 0 when y-axis proportions are 0/0 are confusing to users. This was brought up by participants during the TMRC workshop in October

For continuous variables, we do NOT show markers at 0 when there is no data

The birds eye view & variable completeness table exist so that missing data is accounted for

**Decision**: do not add markers at 0 when the proportion is 0/0

danicahelb:
This ticket essentially duplicates https://github.com/VEuPathDB/web-eda/issues/1504

moontrip:
@danicahelb Yes, I thought so: that was the reason why I gathered them by myself as next cycle 😃 Thank you for handling this 👍

aaronwlsong:
Recreated the example, looks good now, closing ticket.



![Image](https://user-images.githubusercontent.com/61077845/211399349-2048e754-50fb-4db5-8cdb-f5c17141e74d.png)

danicahelb:
This ticket has had off-target affects on markers for undefined proportions (0/0), so I am reopening it.

See this example from LLINEUP (unfiltered)

**QA site looks like this: (birds eye view shows 115,542 data points but not all are represented on the plot since 0/0 markers are missing):**
<img width="1006" alt="image" src="https://user-images.githubusercontent.com/25450900/224828165-5209f1a2-ce54-4e5a-947a-7596c59446d8.png">

[...]

Issue #1519, opened 2022-12-07
Line plot bug: no line is being drawn between discrete points


danicahelb:
Requires UX discussion before implementation:

In SCORE S. mansoni study, I filtered the data so that `Village study arm `= `Arm 6: Gaining-shsh`

This results in having data at just the following timepoints:

<img width="565" alt="image" src="https://user-images.githubusercontent.com/25450900/206065284-2e1e9f1f-f547-4d37-a144-f7fd967e7c8b.png">

When I make a line plot of this subset, plotting S. mansoni against study timepoint I get the following plot:

<img width="1247" alt="image" src="https://user-images.githubusercontent.com/25450900/222212454-90146734-2e3a-4a27-9df8-af660a898286.png">

My issue with the line plot is that is is a series of discrete points instead of a line plot. It is appropriate for a line to be drawn connecting the markers at Year 1, Year 3, and Year 5 to see how prevalence of schisto goes down over the course of the study

danicahelb:
Discussed in 12/12 EDA UX meeting. This is a BUG related to https://github.com/VEuPathDB/web-eda/issues/1411

Once the related bug is fixed, we would expect that a line gets drawn between the remaining data points. If that does not happen, then @moontrip please fix the bug in this ticket as well. Thanks!

danicahelb:
@nkittur-uga now that the bug has been fixed, can you see if there is the expected gap in the line now?

nkittur-uga:
I'm afraid I still don't see the line connecting the points. (Same example as above)
![Screenshot 2023-01-19 at 1 28 20 PM](https://user-images.githubusercontent.com/67710160/213529655-0a485a63-7380-4baf-8453-8d1522f21e33.png)

danicahelb:
Hey @moontrip I think this ticket does NOT depend on implementation of https://github.com/VEuPathDB/web-eda/issues/1082

moontrip:
@danicahelb Thank you for your confirmation 👍 Then it is almost ready for merging actually 😄 I will do that

danicahelb:
Looks great, thanks @moontrip!


![Image](https://user-images.githubusercontent.com/25450900/223195009-8a9454ce-1c12-48b6-8d64-d1519c916cf1.png)

moontrip:
@danicahelb Thank you for your reviews and tests. Glad it worked out 👍

PR #1521, opened 2022-12-09
remove marker for missing data at Lineplot Viz

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1411. 

After extensive tests and check codes, I found that Bob (I guess) nicely implemented nullZeroHack() function not to connect lines for missing data and the processed data was passed through the function. The request by the ticket is to remove those data points, thus the simplest solution is not to use the function.

Here is the screenshot for the same example in the ticket. Clearly, data points for Year 3 and Year 4 are now gone as compared to the screenshot in the ticket (https://github.com/VEuPathDB/web-eda/issues/1411#issue-1409463546).

 lineplot-no-data1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.13.6",
+  "version": "3.13.7",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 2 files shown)


dmfalke:
Sorry for the belated review. Should there also be a line connecting **Year 2** and **Year 5**? Maybe @danicahelb can weigh in on this?

moontrip:
@dmfalke I think so because there exists another ticket to do that https://github.com/VEuPathDB/web-eda/issues/1519 😄

dmfalke (approved):

moontrip:
@dmfalke thank you for your review and approval! 👍

PR #1531, opened 2022-12-15
connect points regardless of missing data for Lineplot Viz

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1519. Although this will depend on https://github.com/VEuPathDB/web-components/pull/431, I have added a temporary solution to test and review this without using `npm-pack-here` (please see inline comment at the code). So you can just do `yarn start` at web-eda

(I was working for finding a solution via handling data, which was too complex, but found an easy solution that is supported by plotly.js itself)

For a demo purpose, I used the same example shown at https://github.com/VEuPathDB/web-eda/issues/1411. You can compare the figure at the PR and below one.

To recreate:
SCORE S. mansoni Cluster Randomized Trial
Filter: Country=Tanzania, Village study arm= Arm 3
Make a Line Plot with the parameters as in screenshot below

lineplot-connect-points1


dmfalke (approved):
I tested this, and it looks good to me. A nice and simple solution :smile:

PR #1553, opened 2023-01-12
check if a computed overlay var is present in birds eye

### LINKED ISSUE #1547: Ranked abundance scatter not showing Data for axes & strata value
Confirmed the value is coming to us from the backend. 
Screen Shot 2023-01-09 at 12 29 51 PM

Alpha diversity scatter and ranked abundance box behavior is normal.

asizemore:
Resolves #1547 

Ranked abundance scatter has a computed overlay var, which means it has an overlay var but `overlayVariable` is undefined. This PR adds a check for a computed overlay var in the birds eye plot.


bobular (approved):
Works for me. Glad it was a simple fix!

Issue #1559, opened 2023-01-13
Gray out whole (3) if the computation hasn't finished.


asizemore:

asizemore:
absorbed by #1561

PR #1577, opened 2023-01-24
Disable steps in computed vars until data/configuration exists

jernestmyers:
Resolves https://github.com/VEuPathDB/web-eda/issues/1559

### DIFF
--- a/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -692,10 +692,20 @@ export function FullScreenVisualization(props: FullScreenVisualizationProps) {
                     stepNumber: 2,
                     stepTitle: `Generate ${computationAppOverview.displayName} results`,
                   }}
+                  isStepDisabled={
+                    !plugin.isConfigurationValid(
+                      computation.descriptor.configuration
+                    )
+                  }
                 >
                   <RunComputeButton
                     computationAppOverview={computationAppOverview}
                     status={computeJobStatus}
+                    showStatus={
+                      !!plugin.isConfigurationValid(
+                        computation.descriptor.configuration
+                      )
+                    }
                     createJob={createComputeJob}
                   />
                 </ComputationStepContainer>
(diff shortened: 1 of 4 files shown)


asizemore:
Not sure if this should be a separate ticket, but sometimes it looks like the config selectors changing makes the visualization try to update.
So below, i made a new ranked abundance box, selected my app config, then got this screen
Screen Shot 2023-01-26 at 2 58 33 PM

Sometimes this happens if i have already made a plot, and then go change config values. So here i had a fully working plot with Data: Phylum. I changed to this metagenomic parameter, and saw the following.
Screen Shot 2023-01-26 at 3 05 04 PM
I switched the param back to Phylum, all works. 
I go back again to that metagenomic one, and now the job is complete and i can use the plot!
Screen Shot 2023-01-26 at 3 06 39 PM

So i think what's happening is the config changing is autostarting a job, and the viz is looking for that data. If i go back again to that config, the compute job has finished so it's all there!

I'd guess for sake of storage space we wouldn't want to autostart a job when the config changes. 

Anyways, might be a separate issue but logging it here to start!

jernestmyers:
Ah, yes. That's a problem. Thanks for detailed examples and steps to re-create. I'll dig into it now.

src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -808,20 +788,25 @@ function ConfiguredVisualizationGrayOut({
...
               primary: 'Computation failed',
               secondary: 'Please contact us for support.',
             }
-          : computeJobStatus === 'no-such-job'
+          : computeJobStatus === 'no-such-job' ||
+            !isComputationConfigurationValid
           ? {
-              primary: 'Computation not started',
+              primary: `Computation not ${
+                isComputationConfigurationValid ? 'started' : 'configured'
```
asizemore, line 808: This is great

jernestmyers:
Removing [this logic and conditional rendering for job status](https://github.com/VEuPathDB/web-eda/pull/1577/commits/43d80bda67aeb6e417c7d7b039fdca1ed44e5d38#diff-3572d468c2c2368d816244221b36f129a04f95f4a20621805a9905c6589e24ffL713-L718) appeared to be the culprit. 

I've more or less reverted back to where we use plot SVGs based on the job status.

image

PR #1580, opened 2023-01-25
Revive mapveu app

### LINKED ISSUE #1546: Standalone map prerequisite: move map state storage to analysis
Previously/currently the full-screen map state was stored in user preferences, but this makes it non-shareable.

There will be some back end work for this too.

The information that needs storing (in addition to filters and visualizations) is

- map center lat/lon and leaflet zoom level
- background tile choice (e.g. regular, satellite, etc)
- mouse mode (regular or magnifying)
- overlay variable (or when null, use first featured variable)
- display entity (or when null, use overlay variable's entity)
- marker mode (pie, count, proportion)
- checked legend items (or when null, all are checked as is the case with regular visualizations)
- customised screen positions of draggable panels
  - there are the main "management" panels: study info, saving, sharing, legend, filters, visualizations, downloads
  - plus the subsetting/filtering panel itself (e.g. in [this slide](https://docs.google.com/presentation/d/17Yg4Kz00pW6wbNo3bK0DpQjW4cVEzrq0VQ75pw1xZNc/edit?pli=1#slide=id.g1c96a892add_0_59))
  - then potentially there are saveable positions for each visualization, but for starters we could just allow one visualization on the screen at any time, and save just one position for it

### LINKED ISSUE #1549: Standalone map prerequisite: move map to its own primary/base route
`workspace/maps/DS_6b79157ec1` ?

dmfalke:
fixes #1549
fixes #1546

This PRs revives the mapveu application. It is a starting point for further development and refinement. The goal is to provide the basic wiring and implemention of core features (state management, filtering, and visualizing).

### The following is included:

- [x] Added a link on the homepage to the mapveu app. When this is added to a website, the `MapVEuContainer` component will need to be integrated with that site's router.
- [x] Implemented basic map-based analysis in `MapVEuAnalysis` component, including "full screen" styling and crude overlay panels.
- [x] Persist state as a part of the analysis
- [x] Implement visualizations
- [x] (?) Implement crude subsetting panel
- [ ] Remove website header from view (or make it optional), in preparation for map-specific header drawer. This may require some work in WDKClient so that the header can be hidden **(Will be a follow-up)**
- [ ] ~Implement "viewport" filter~ Decided not to do this
- [ ] Rename "MapVEu" to something else. Perhaps just "Map". **(Partially done)**

### DIFF
--- a/src/lib/core/components/GlobalFiltersDialog.tsx
+++ b/src/lib/core/components/GlobalFiltersDialog.tsx
@@ -8,6 +8,7 @@ import { Alert } from '@material-ui/lab';
 import _ from 'lodash';
 import 'react-resizable/css/styles.css';
 import './GlobalFiltersDialog.scss';
+import { VariableLinkConfig } from './VariableLink';
 
 interface Props {
   // Whether the dialog is open
@@ -22,6 +23,7 @@ interface Props {
   setFilters: (filters: Filter[]) => void;
   // A function to remove a given filter
   removeFilter: (filter: Filter) => void;
+  variableLinkConfig: VariableLinkConfig;
 }
 
 /**
@@ -51,6 +53,7 @@ export default function GlobalFiltersDialog(props: Props) {
               filters={matchingFilters}
               entities={props.entities}
               removeFilter={props.removeFilter}
+              variableLinkConfig={props.variableLinkConfig}
             />
           </div>
         );
--- a/src/lib/core/components/GlobalFiltersDialog.tsx
+++ b/src/lib/core/components/GlobalFiltersDialog.tsx
@@ -23,6 +23,7 @@ interface Props {
   setFilters: (filters: Filter[]) => void;
   // A function to remove a given filter
   removeFilter: (filter: Filter) => void;
+  // Determines if we render a link or a button.
   variableLinkConfig: VariableLinkConfig;
 }
 
--- a/src/lib/core/components/VariableLink.tsx
+++ b/src/lib/core/components/VariableLink.tsx
@@ -1,32 +1,66 @@
 /*
  * Link to the page for a variable
  */
-import { useMemo } from 'react';
+import { forwardRef, Ref } from 'react';
 import { Link, LinkProps } from 'react-router-dom';
-import { useMakeVariableLink } from '../hooks/workspace';
+
+type VariableValue = {
+  entityId?: string;
+  variableId?: string;
+};
+
+export type VariableLinkConfig =
+  | {
+      type: 'link';
+      makeVariableLink: (value?: VariableValue) => string;
+    }
+  | {
+      type: 'button';
+      onClick: (value?: VariableValue) => void;
+    };
 
 export interface Props<S = unknown> extends Omit<LinkProps<S>, 'to'> {
   entityId?: string;
   variableId?: string;
+  linkConfig: VariableLinkConfig;
 }
 
-export function VariableLink(props: Props) {
-  const { entityId, variableId, ...rest } = props;
-  const makeVariableLink = useMakeVariableLink();
-  const variableLink = useMemo(
-    () =>
-      makeVariableLink({
-        entityId,
-        variableId,
-      }),
-    [makeVariableLink, entityId, variableId]
-  );
+export const VariableLink = forwardRef(
+  (props: Props, ref: Ref<HTMLAnchorElement>) => {
+    const { entityId, variableId, linkConfig, ...rest } = props;
+    const value = { entityId, variableId };
 
-  return (
-    <Link
-      replace
-      {...rest}
-      to={{ pathname: variableLink, state: { scrollToTop: false } }}
-    />
-  );
-}
+    return linkConfig.type === 'link' ? (
+      <Link
+        replace
+        {...rest}
+        to={{
+          pathname: linkConfig.makeVariableLink(value),
+          state: { scrollToTop: false },
+        }}
+      />
+    ) : (
+      // Typically, we would just use a <button> for this case,
+      // but this needs to work in an SVG fragment, which does
+      // not support <button>, so this is a compromise
+      // eslint-disable-next-line jsx-a11y/anchor-has-content
+      <a
+        ref={ref}
+        role="button"
+        aria-pressed="false"
+        tabIndex={0}
+        onKeyDown={(event) => {
+          if (event.key === 'Enter' || event.key === ' ') {
+            event.preventDefault();
+            linkConfig.onClick(value);
+          }
+        }}
+        onClick={(event) => {
+          event.preventDefault();
+          linkConfig.onClick(value);
+        }}
+        {...rest}
+      />
+    );
+  }
+);
--- a/src/lib/core/components/VariableLink.tsx
+++ b/src/lib/core/components/VariableLink.tsx
@@ -27,12 +27,14 @@ export interface Props<S = unknown> extends Omit<LinkProps<S>, 'to'> {
 
 export const VariableLink = forwardRef(
   (props: Props, ref: Ref<HTMLAnchorElement>) => {
-    const { entityId, variableId, linkConfig, ...rest } = props;
+    const { entityId, variableId, linkConfig, style, ...rest } = props;
     const value = { entityId, variableId };
 
     return linkConfig.type === 'link' ? (
       <Link
+        ref={ref}
         replace
+        style={style}
         {...rest}
         to={{
           pathname: linkConfig.makeVariableLink(value),
@@ -47,8 +49,8 @@ export const VariableLink = forwardRef(
       <a
         ref={ref}
         role="button"
-        aria-pressed="false"
         tabIndex={0}
+        style={{ cursor: 'pointer', ...style }}
         onKeyDown={(event) => {
           if (event.key === 'Enter' || event.key === ' ') {
             event.preventDefault();
--- a/src/lib/core/hooks/study.ts
+++ b/src/lib/core/hooks/study.ts
@@ -134,7 +134,7 @@ export function useWdkStudyRecords(
       );
     },
     [attributes, tables]
-  )?.records;
+  )?.records.filter((record) => record.attributes.eda_study_id != null);
 }
 
 /**
--- a/src/lib/core/utils/data-element-constraints.ts
+++ b/src/lib/core/utils/data-element-constraints.ts
@@ -218,7 +218,9 @@ function variableConstraintPredicate(
   );
 }
 
-export type VariablesByInputName = Partial<Record<string, VariableDescriptor>>;
+export type VariablesByInputName<T extends string = string> = Partial<
+  Record<T, VariableDescriptor>
+>;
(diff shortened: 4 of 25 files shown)


dmfalke:
@bobular what are your thoughts on the subsetting map filter sharing the same component as the map analysis? 

The reason I ask is that it might make sense to persist the state of the map as variable UI state (similar to what we do with the histogram and table filters). We can make certain features optional, if we want something in one context, but not the other. The only kink would be if there is a bit of state that just doesn't make sense as variable UI state. Nothing really comes to mind, with the assumption that can disable features (like the study details panel, menu, etc).

dmfalke:
Okay, I see. I guess we wouldn't have an overlay variable in geospatial subsetting, either?

bobular:
> Okay, I see. I guess we wouldn't have an overlay variable in geospatial subsetting, either?

I can see a perfectly functional geospatial subsetting without it. In fact, it might be seen as bad to draw regions on the map based on the distribution of a variable, as opposed to the "pure" drawing of regions based on geography. We can always add it later if there's a pressing use case.

bobular (commented):
All makes sense! The main app component is nice and clean at the moment. I wonder how long that will last...

src/lib/core/hooks/study.ts
```diff
@@ -134,7 +134,7 @@ export function useWdkStudyRecords(
       );
     },
     [attributes, tables]
-  )?.records;
+  )?.records.filter((record) => record.attributes.eda_study_id != null);
```
adnauseum, line 137: **question**: why would an `eda_study_id` be `null`?

src/lib/core/utils/data-element-constraints.ts
```diff
@@ -218,7 +218,9 @@ function variableConstraintPredicate(
   );
 }
 
-export type VariablesByInputName = Partial<Record<string, VariableDescriptor>>;
+export type VariablesByInputName<T extends string = string> = Partial<
```
adnauseum, line 221: **comment**: TypeScript n00b here! What is `T extends string = string>` doing?

bobular: And the `= string` is a fall-back if the "user" doesn't use the generic-style declaration.

Sorry I never did CS so I don't use the correct words all the time.

bobular:
Hi @dmfalke - had another quick look. Looking great.

Should we add a new checkbox TODO item for constraining the floating plot overlay to the global overlay variable, or shall we add this as a separate ticket?

src/lib/core/utils/data-element-constraints.ts
dmfalke: What @bobular said is correct. Another way to say it is, the type parameter `T` must extend `string`, and will default to `string` if not specified by the consumer.

bobular (approved):
Still in draft but I've had a third look (and a quick test) and all for you merging. It looks like a nice clean slate to start working on.

src/lib/mapveu/analysis/MapAnalysis.tsx
```diff
@@ -0,0 +1,424 @@
...
+                  />
+                </div>
+                <FilledButton
+                  text="Add a plot"
+                  onPress={() => setIsVizSelectorVisible(true)}
+                />
+                <ul>
+                  {analysisState.analysis?.descriptor.computations.map(
+                    (computation) => (
+                      <li>
```
adnauseum, line 327: **suggestion**: `<li key={computation.computationId}>`

src/lib/mapveu/analysis/MapAnalysis.tsx
```diff
@@ -0,0 +1,424 @@
...
+                        <strong>
+                          {computation.displayName} (
+                          {computation.descriptor.type})
+                        </strong>
+                        <ul>
+                          {computation.visualizations.map((viz) => (
+                            <li>
```
adnauseum, line 334: **suggestions**: `<li key={viz.visualizationId}>`

dmfalke:
Sorry for the delay. I have been working on integrating the subletting panel and had to make some substantial changes to how we render links/buttons. I'll have some changes pushed tonight.

src/lib/mapveu/MapVeuAnalysisList.tsx
```diff
@@ -5,32 +5,46 @@ import { safeHtml } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
...
+        )
+      : undefined;
     const { analysisId } = await analysisStore.createAnalysis(
-      makeNewAnalysis(studyId)
+      makeNewAnalysis(studyId, computation)
     );
     history.push(`${url}/${analysisId}`);
-  }, [analysisStore, history, studyId, url]);
+  }, [analysisStore, history, singleAppMode, studyId, url]);
   return (
```
adnauseum, line 45: 😊  While you're in here, there's another key error

```js
        <ul>
          {list.value?.map((analysis) => (
            <li> // 👈 key={analysis.analysisId}
              <Link to={`${url}/${analysis.analysisId}`}>
```

https://github.com/VEuPathDB/web-eda/blob/6e264755f0f4e888cae961d22cb8e69e20cc296c/src/lib/mapveu/MapVeuAnalysisList.tsx#L61

src/lib/core/components/GlobalFiltersDialog.tsx
```diff
@@ -22,6 +23,7 @@ interface Props {
   setFilters: (filters: Filter[]) => void;
   // A function to remove a given filter
   removeFilter: (filter: Filter) => void;
+  variableLinkConfig: VariableLinkConfig;
```
adnauseum, line 26: **nitpick**: Might want to add a comment for uniformity. Maybe something like: "a configuration option allowing items in the VariableList to be either buttons or anchors."

src/lib/core/components/VariableLink.tsx
```diff
@@ -1,32 +1,66 @@
...
+      />
+    ) : (
+      // Typically, we would just use a <button> for this case,
+      // but this needs to work in an SVG fragment, which does
+      // not support <button>, so this is a compromise
+      // eslint-disable-next-line jsx-a11y/anchor-has-content
+      <a
+        ref={ref}
+        role="button"
+        aria-pressed="false"
```
adnauseum, line 50: **question**: will `aria-pressed` update on press?

src/lib/core/components/GlobalFiltersDialog.tsx
dmfalke: Ah, yes. I'll add an appropriate comment

src/lib/core/components/VariableLink.tsx
dmfalke: Oh, yeah. I'll look at this in more detail soon. I might be missing some other aria attributes...

src/lib/mapveu/analysis/MapAnalysis.tsx
```diff
@@ -0,0 +1,517 @@
...
+    fieldTree,
+    studyMetadata.rootEntity.id,
+  ]);
+
+  const fullScreenActions = (
+    <>
+      <div>
+        <Tooltip title="Delete visualization">
+          <button
+            type="button"
```
adnauseum, line 246:
```suggestion
            aria-label={`Delete ${activeViz?.displayName || "visualization."}`}
            type="button"
```

adnauseum: This was from the github.dev tool, so apologies if this looks weird. Trying out the new tool :D

src/lib/mapveu/analysis/MapAnalysis.tsx
```diff
@@ -0,0 +1,517 @@
...
+            }}
+          >
+            <i className="fa fa-trash"></i>
+          </button>
+        </Tooltip>
+      </div>
+      <div>
+        <Tooltip title="Copy visualization">
+          <button
+            type="button"
```
adnauseum, line 265:
```suggestion
            aria-label={`Create a copy of ${activeViz?.displayName || "visualization."}`}
            type="button"
```

src/lib/mapveu/analysis/MapAnalysis.tsx
```diff
@@ -0,0 +1,517 @@
...
+              if (activeViz == null) return;
+              updateVisualizations((visualizations) =>
+                visualizations.filter(
+                  (v) => v.visualizationId !== activeViz.visualizationId
+                )
+              );
+              setActiveVisualizationId(undefined);
+            }}
+          >
+            <i className="fa fa-trash"></i>
```
adnauseum, line 258:
```suggestion
            <i 
                aria-hidden 
                className="fa fa-trash"
```

src/lib/mapveu/analysis/MapAnalysis.tsx
```diff
@@ -0,0 +1,517 @@
...
+                  visualizationId: vizCopyId,
+                  displayName:
+                    'Copy of ' +
+                    (activeViz.displayName || 'unnamed visualization'),
+                })
+              );
+              setActiveVisualizationId(vizCopyId);
+            }}
+          >
+            <i className="fa fa-clone"></i>
```
adnauseum, line 282:
```suggestion
            <i 
                aria-hidden 
                className="fa fa-clone"
```

src/lib/mapveu/analysis/MapAnalysis.tsx
```diff
@@ -0,0 +1,517 @@
...
+      </div>
+      <Tooltip title="Minimize visualization">
+        <Link
+          to=""
+          onClick={(e) => {
+            e.preventDefault();
+            setActiveVisualizationId(undefined);
+          }}
+        >
+          <i className="fa fa-window-minimize" />
```
adnauseum, line 294:
```suggestion
            <i 
                aria-hidden 
                className="fa fa-window-minimize"
```

adnauseum (approved):

src/lib/core/components/VariableLink.tsx
dmfalke: I think the solution is to remove `aria-pressed`. That's more useful for indicating toggle state, which is not needed here.

dmfalke: I'm now really wanting to use a `<button>`. I will explore using `<foreignObject>` in EntityDiagram.

dmfalke: `<foreignObject>` really doesn't seem viable. I'll stick with the current approach. Hopefully is meets a11y requirements. @adnauseum do you know of a good tool for checking that?

adnauseum: Well, if we're ever in a situation were we have like an ADA audit: [Assistive Labs](https://assistivlabs.com/pricing) is a good choice, but that is most definitely not free 😄 . The free options are not great. I use [this Chrome Extension](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en) to make sure things are useable. I think it's process of gradual enhancement anyways

PR #1584, opened 2023-01-30
Add viz-specific compute config

### LINKED ISSUE #1548: Ranked abundance scatter shows too many overlay values
Screen Shot 2023-01-09 at 12 50 06 PM

Above we have 10 overlay values instead of 8. Not seen on the beta site.

asizemore:
Resolves #1548 

## Important
This PR requires the `add-numbervarsreturned-abundance` branch from service-eda-compute.

In rare cases, the viz type affects what configuration the computation should use. Our first example is for ranked abundance. The computation returns 10 variables by default. The boxplot can handle 10 variables, since those vars are mapped to x axis values, but the scatterplot can only handle 8 variables because those vars are going to the overlay. 

History on the topic: This viz-specific configuration was handled in the data service before the compute service was wired up.

This PR adds `additionalComputeConfig` to a viz's options in the ranked abundance app. This way, the app plugin can control which config goes to which visualization. 

One issue i ran into while making this change was adding this extra bit to the compute config without making it think it's "updating", and thus re-requesting the compute, every 10 seconds. To fix this i added the conditional on the compute config being empty.

### DIFF
--- a/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/src/lib/core/components/computations/plugins/abundance.tsx
@@ -156,15 +156,19 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
   assertComputationWithConfig<AbundanceConfig>(computation, Computation);
 
   // Add visualization type specific computation configuration with plugin.additionalComputeConfig
-  // (for example, here scatterplot has as additional parameter that the boxplot does not need)
+  // (for example, here scatterplot and boxplot differ on the numberVariablesReturned param)
   // Only need to add this extra config when the config is empty.
-  const visualizationType = computation.visualizations.find(
+  const visualizationType: string = computation.visualizations.find(
     (viz) => viz.visualizationId === visualizationId
-  )?.descriptor.type;
+  )
+    ? computation.visualizations.find(
+        (viz) => viz.visualizationId === visualizationId
+      )!.descriptor.type
+    : '';
 
   if (!computation.descriptor.configuration) {
     computation.descriptor.configuration = {
-      ...plugin.visualizationPlugins[visualizationType as string]?.options
+      ...plugin.visualizationPlugins[visualizationType]?.options
         .additionalComputeConfig,
     };
   }
--- a/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/src/lib/core/components/computations/plugins/abundance.tsx
@@ -162,11 +162,14 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
   const visualizationType = computation.visualizations.find(
     (viz) => viz.visualizationId === visualizationId
   )?.descriptor.type;
-  computation.descriptor.configuration = {
-    ...computation.descriptor.configuration,
-    ...plugin.visualizationPlugins[visualizationType as string]?.options
-      .additionalComputeConfig,
-  };
+
+  if (!computation.descriptor.configuration) {
+    computation.descriptor.configuration = {
+      ...plugin.visualizationPlugins[visualizationType as string]?.options
+        .additionalComputeConfig,
+    };
+  }
+
   const configuration = computation.descriptor.configuration;
   console.log(configuration);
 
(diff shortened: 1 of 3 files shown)


jernestmyers (approved):

dmfalke (commented):
I don't see where `numberVariablesReturned` is being read. If it's not being read anywhere, how is it being used? Sorry, I must be missing something.

dmfalke:
> I don't see where `numberVariablesReturned` is being read. If it's not being read anywhere, how is it being used? Sorry, I must be missing something.

Apologies, I see what's happening now. It's being included in the request to the viz endpoint.

asizemore:
> My first question is, doesn't this cause each viz to have a different compute job (since they are based on the compute config)? If that is the case, how does the compute job status work with this?

Oh.... oh wait I see. So at first I thought it would be fine because computes are unique based on their parameters. So we'd just have an extra parameter around. But when I read your question again, I realized that (as you point out, im just thinking aloud) if the vizs have different compute jobs, if we grouped them by compute parameters the scatters will be in separate apps from the box plots. 

Right. Okay so where to put this info? The compute service doesn't know about the viz type unless we tell it, which then gives us the viz-specific-parameter issue above. And we don't talk to the data service before asking for the compute job, it just goes straight to the compute service if i understand correctly.

So one option would be to remove this extra parameter and let the scatterplot continue to receive data that has 10 overlay values. We could update the scatterplot to only plot the first 8. The advantage is that the scatterplot then is resistant to us accidentally sending too much data in the future. The downside is that sending too much data will fail silently, and also for ranked abundance we'll *always* send data for 10 variables but only use 80% of that. That could be thousands of points send to the client but not plotted.

I'll check out the data service. It's possible we can manipulate the data there so that we only send the correct data to plot.data and consequently the frontend... It still means that there's one data handoff that's sending too much data, but at least that's not going to the client.

dmfalke:
> I'll check out the data service. It's possible we can manipulate the data there so that we only send the correct data to plot.data and consequently the frontend... It still means that there's one data handoff that's sending too much data, but at least that's not going to the client.

This is definitely not ideal, but maybe it's good enough for now? I do wonder if the viz plugin could be allowed to pass addition params that are not considered part of the config, sort of like a "format config". I feel like we should get @ryanrdoherty to weigh in.

ryanrdoherty:
Ok, I read the comments above.  Sounds like the compute plugin returns 10 vars but we only want 8 making their way into the R code?  Is that correct?  Do we know which columns need to be trimmed?  There are a bunch of options here.  Easiest for me is that the R code ignore the 2 vars.  If we can't do that, we can trim them at multiple places along the way.  The best place (because it doesn't require any new features) is probably in the Java code of the abundance scatterplot.  If we think this is going to be a regular thing we can add some tooling, or most extreme would be to have the viz specify which of the computed vars they want from the merge service.  Right now they just get them all.  That would be a new feature in merging and would be involved because the viz plugin would probably have to get the computed var metadata first, select the vars it wants, and then ask merging to only include those.

asizemore:
> Sounds like the compute plugin returns 10 vars but we only want 8 making their way into the R code

Yup! The default behavior of the compute plugin is to return 10 vars. The code it runs has an optional parameter to return N variables (compute service [PR](https://github.com/VEuPathDB/service-eda-compute/pull/40)).

> Do we know which columns need to be trimmed?

Also yes! The computation is a ranking of variables, so it's returning the top 10 variables by default. We only want to keep the top 8. If I remember correctly, the compute itself returns the variables in the ranked order, but @d-callan would be better to confirm that's the case. If that's true, we can just chop off the last two variables.

> If we think this is going to be a regular thing we can add some tooling

I don't foresee another use case for anything we have planned in the next 6 months. I could imagine cases where it's helpful in the future so i don't want to say it won't happen again, but none of those cases are even on the prioritization list yet so it should be a while :)

d-callan:
It should be returning them in order, so we can display them in order in the box plot. And while I don't think it's ideal, I agree the ranked abund scatter plugin in the data service is the best place for this. I think its reasonable enough for a viz to be a bit opinionated about how it wants to use compute results.

ryanrdoherty:
In that case, @asizemore should probably throw away the compute service PR (remove the numCols config param), and we can add a trimmer to the viz plugin.  @d-callan do you want help with that?  I can write a stream that will trim off the last two cols of each line (including header).  That is correct given Ann's ordering comment above, right?

d-callan:
I'll take a look, because I think I can probably just tell R to not read them.

ryanrdoherty:
I'm taking a look at trimming in Java.  Looking like it may slow down the data transfer... but maybe not a lot.  We'll have to see.

PR #1588, opened 2023-02-02
Display error messaging in download modal

### LINKED ISSUE #1522: Downloads: Unhandled backend error
There is a backend error from the `/tabular` endpoint. The UI is not handling this error, so I will make another issue for that.

_Originally posted by @dmfalke in https://github.com/VEuPathDB/EdaDataService/issues/226#issuecomment-1344278636_

jernestmyers:
Resolves #1522 

Displays a rudimentary message informing user of data fetching error. The error response expected in this component differed from what is currently being received. An error modal is being rendered as a result of this mismatch, but it appears behind the download modal. Instead, I've rendered the error in the download modal itself though it may be preferred to render it differently in terms of location, styling, and/or messaging.

**Before** (image 1 shows blank space where data grid is expected and one can faintly see the error modal behind the download modal; image 2 shows the error modal when the download modal is closed):
image
image

**After** (notice there's no more error modal resulting from the `JSON.parse()` syntax error):
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.15.9",
+  "version": "3.15.10",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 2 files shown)


dmfalke (commented):
Nice work, Jeremy. I have some comments/suggestions:
1. Use an error "banner" to indicate that there was an error.
2. Retain the previous behavior (using `JSON.parse`), since it appears to be handling a subclass of errors (see suggestion).
3. Do not display the error message, since it's not useful for users to see.

Let me know what you think.

asizemore:
I definitely like having the error within the modal - great idea. 
I think it's just the table part that errs(?), but the download button should still work (just tested on feature). Can we move the messaging underneath the 2 View Table and Download? Then the Download button could still function, and it'd be more clear that it's just the viewing the table part that's erring.

dmfalke (approved):

PR #1589, opened 2023-02-02
banner control behaviors for lineplot

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1503. Several edge cases are also addressed:

a) the very basic and fundamental feature: no plotly display if banner is activated
b) if no display, then no truncation or warning is shown either.
c) better display that involves logScale banner: this relies on https://github.com/VEuPathDB/web-components/pull/447
d) keep banner(s) when going out of Viz and coming back


bobular:
Hi @moontrip 
Sorry I don't have time for a full review.
Y-axis looks great.
I did some testing and noticed a small issue with the x-axis
With the following GEMS1 example
image
Is it possible to avoid the data call and spinner when enabling binning?

moontrip:
@bobular Thank you for your review and tests. I will check if it is possible, "avoid the data call and spinner when enabling binning"

moontrip:
@bobular I have contemplated and tested your suggestion. My conclusion is that it is better to leave the current behavior as is. It is because that log scale toggle does not request a new data. For example, if both log scale and Binning are turned on then banner is shown and no plot is displayed. Under this condition, if only log scale is turned off (while Binning is on), banner disappears, however, no plot is still displayed because there is no data request.

Similar to this, Y-axis's Error bars request a new data (like Binning at X-axis), thus that's the reason why Y-axis case works

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -944,11 +944,31 @@ function LineplotViz(props: VisualizationProps<Options>) {
...
+  ] = useState(false);
+  const [showBinningBanner, setShowBinningBanner] = useState(false);
+  const [
+    showDependentLogScaleBanner,
+    setShowDependentLogScaleBanner,
+  ] = useState(false);
+  const [showErrorBarBanner, setShowErrorBarBanner] = useState(false);
+
+  // add showBanner prop in this Viz
+  const [showBanner, setShowBanner] = useState(true);
```
bobular, line 960: I don't think this state is needed at all?

In CoreUI's `Banner`, `showBanner` defaults to `true` (the only value we ever set it to here in this component) and `setShowBanner` is only used by `Banner` if `autoHideDuration` is non-null (we don't use that prop here).

I mentioned some unclear behaviours here: https://github.com/VEuPathDB/CoreUI/issues/147

bobular (changes requested):
Happy to test this again without the `showBanner`/`setShowBanner` state. Everything else looks good.

I looked into the spinner issue and confirm that it also happens on the y-axis. I agree with you that we can keep the current behaviour, but maybe for different reasons: 
1. the banner warning comes up straight away, we don't have to wait for the spinner/request to complete
2. having made the new request, we get an instant result if we turn off logscale, which is kind of nice (though if turning off binning or error bars, we have to wait)
3. I tried a quick fix (returning null from the `data = usePromise` when the toggles are incompatible) but this introduced more serious problems (mainly because the `data` response is needed for the bin controls).

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: Good afternoon, @bobular. Thank you for your detailed reviews and tests. 👍 
As for showBanner/setShowBanner, it was necessary with the previous logic to enable Banner component again after closing it. But with this new logic, it may not be required. I will check it carefully on next Monday (today, Friday, I am off to VEuPathDB).

bobular (approved):
Glad that was helpful.

I've approved the no-data dticks PR in web-components too. Thanks for the nudge!

moontrip:
Thank you for your detailed reviews/tests/suggestions which enhance this PR 👍

PR #1591, opened 2023-02-02
use placeholderDisplayName if no variableId for y label

### LINKED ISSUE #1585: Ranked abundance scatter: fix display name of y axis
Screen Shot 2023-01-30 at 10 55 34 AM

Currently "To be determined by client values". Change to "Abundance".

asizemore:
Resolves #1585 

Previously the scatterplot was showing a value returned by the compute. This PR tells the scatterplot to use the placeholder name given by the abundance app instead of the variable display name returned by the compute, when a true variable isn't actually going to be returned.

I saw two options for solving this issue. The first is what i've done here, and the second is to update [this line](https://github.com/VEuPathDB/microbiomeComputations/blob/1516c3e626d58c5cf06a757bd1c32029c513e5c6/R/method-rankedAbundance.R#L85) with a nicer display name. 

I chose this route of updating the frontend because eventually this display name should be driven by collection variable metadata (that doesn't exist yet). The compute, however, does not know about collection metadata. It's called "Ranked Abundance" but really it could rank any type of data (for example, it also currently can be used to rank pathway coverage data. Anyways, once the collection variable metadata with display name does exist, we would plug that into the relevant part of the plugin Options.

Tested on mbio and clinepi

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.15.7",
+  "version": "3.15.8",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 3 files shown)


asizemore:
courtesy tag @dmfalke

jernestmyers (approved):

PR #1597, opened 2023-02-06
Fix/1582/apps grayout emptyplot

### LINKED ISSUE #1582: Replace job status svgs with empty plot
Instead of the iconographic plot, show the regular viz in step 3. To be more precise, let's replace the step 3 here:
image
with the step 3 shown below

Some issues with this switch arose here: https://github.com/VEuPathDB/web-eda/pull/1577#issuecomment-1405591194

asizemore:
Resolves #1582 

First, removed the compute status svgs from rendering following @jernestmyers 's work in https://github.com/VEuPathDB/web-eda/pull/1577.

Second, had viz listen to the `computeJobStatus` and prevented them from asking for data if the compute was not complete. To test this part, create a plot in an app (have to create the whole plot, like in alpha div actually pick an x axis var and everything), then go change config to one that has not yet been run. There should be no error now :)

Tested on all three mbio compute apps, mbio XY Relationships scatter, Distributions box, and clinepi vizs.

Another note - the jobs sometimes autostart when selecting config. So for example i'll select compute params, see that there's no such job, change params, go back to the original and now woah there's a complete result! This was preexisting behavior, and i'll make a new ticket for this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.15.11",
+  "version": "3.15.12",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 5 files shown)


jernestmyers (approved):
Looks good overall. I played around with this on mbio and clinepi and it behaves as desired.

PR #1599, opened 2023-02-06
remove evenness from alpha div

### LINKED ISSUE #1598: Alpha diversity: remove "evenness" method
Discussed in the microbiome channel, the "evenness" method in alpha diversity is not appropriate for the data we have. We load relative (normalized) abundance, but the "evenness" method assumes absolute (non-normalized) abundance data.

Until we have the ability for apps to constrain data they use, which is a few build cycles off at least, we decided to simply remove the "evenness" method from alpha diversity.

asizemore:
Resolves #1598 

Removes the 'evenness' method from alpha diversity. Pretty straightforward :) 

Before:
Screen Shot 2023-02-06 at 10 30 09 AM

After:
Screen Shot 2023-02-06 at 10 30 32 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "3.15.10",
+  "version": "3.15.11",
   "dependencies": {
     "debounce-promise": "^3.1.2",
     "fp-ts": "^2.9.3",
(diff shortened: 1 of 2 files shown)


jernestmyers (approved):
Be sure to get some rest after this exhausting effort! 🥇

PR #1604, opened 2023-02-08
Fix two key warnings in InputVariables component

adnauseum:
Discovered some missing keys whilst poking around the code today. There wasn't an values that were obviously stable (to me), but I think I picked them adequately.


dmfalke (approved):
Thanks for fixing these pesky warnings!

Issue #1605, opened 2023-01-13
Visualizations: Scatter plot opacity slider defaults


d-callan:
I think we should change the default value for the opacity slider to something between .5 and .7 .. its way easier to see overplotted regions of the plot that way. 

also, i think if the user changes the value of the slider, it should be remembered when changing variables rather than going back to default.

moontrip:
@danicahelb @asizemore @d-callan I have one question about the direction of the slider. As you can remember Ann initiated the idea of reverse slider starting from 1 to 0 (darker to lighter) and Danielle liked the idea, but finally Danica wanted to leave it as is (normal slider). 

The final decision was made while I was absent due to my sickness at the end of Jan. So, I missed it and implemented the feature of reverse slider, which is certainly optional/controllable for both normal and reverse slider: https://github.com/VEuPathDB/web-components/pull/450. My question is that do you envision that we will need such a reverse slider in the future? If true, then we will include the reverse feature to the current slider.

moontrip:
Thanks @asizemore for very detailed information.

SheenaTomko:
Looks good

moontrip:
Thanks @SheenaTomko for your reviews and confirmation 👍

PR #1606, opened 2023-02-09
changes for scatterplot opacity slider

moontrip:
This resolves https://github.com/VEuPathDB/web-eda/issues/1605. Also this ticket relies on https://github.com/VEuPathDB/web-components/pull/450.

- update: The isReverseSlider prop is not used (for now?), thus I commented out the prop. By doing so, this PR does not necessarily rely on the web-component PR. 

Here is an screenshot:
scatterplot opacity slider01


moontrip:
@jernestmyers if you have time, it would be great if you can review this PR 👍 (related one too as I asked at https://github.com/VEuPathDB/web-components/pull/450 😃)

src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -402,7 +402,7 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
           ? vizConfig.independentAxisValueSpec
           : 'Full',
         dependentAxisValueSpec: 'Full',
-        markerBodyOpacity: 0,
+        markerBodyOpacity: 0.5,
```
jernestmyers, line 405: Looks like we're still setting `markerBodyOpacity` to the now-default value (0.5) whenever the `vizConfig` changes.

moontrip: Good question @jernestmyers! That part is related to the change of input variable (X, Y, etc.). We usually reset most of plot control related ones to its default whenever input variable is changed: e.g., just above the line, axis range option of dependent axis is reset to its default, 'Full' when a variable is changed. Due to this reason, I followed that rule for marker opacity as well. Does this make sense to you?

moontrip: @jernestmyers Oh... I missed that part " i think if the user changes the value of the slider, it should be remembered when changing variables rather than going back to default.". When I said "remembering user's choice of opacity" in my summary, I meant a case when a user goes out of Viz and comes back. Good catch! 👍 I will change it accordingly.

moontrip: @jernestmyers I reflected it. The line can be simply removed to preserve the marker opacity value when changing an input variable 👍

jernestmyers (approved):
Looks and works now in accordance with my interpretation of the issue comments!

Before merging, could we remove all the commented-out code related to the reversed slider since it's not being implemented anymore?

moontrip:
> Looks and works now in accordance with my interpretation of the issue comments!
> 
> Before merging, could we remove all the commented-out code related to the reversed slider since it's not being implemented anymore?

Thank you for your detailed review and catching the one I missed 👍 Yes I will remove the comments concerning the reverse slider one.

Issue #1575, opened 2023-01-11
2x2 Mosaic: update parameter selection frontend to allow users to select reference groups for all variables


danicahelb:
Related to https://github.com/VEuPathDB/EdaDataService/issues/130

Final parameter selection UI can be found here: https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=44%3A431&t=ZCxlWaPOOG1AbgKr-0

danicahelb:
<img width="706" alt="image" src="https://user-images.githubusercontent.com/25450900/214134698-9aea2d59-0fc5-44be-9d27-e005cbfc6d3c.png">

nkittur-uga:
I keep running into this bug while trying to do this:
<img width="1417" alt="Screenshot 2023-02-22 at 12 10 43 PM" src="https://user-images.githubusercontent.com/67710160/220704142-94654671-9b96-4bba-963b-1318148816c2.png">

jernestmyers:
This is currently only testable on a dev site due to the backend changes.

> Also, I noticed that value selection is only available (enabled) after both X and Y variables have been selected. It left me wondering for a second why the X value selection was not available after I selected the X var.
I initially implemented this along the lines of your thinking, but then [I interpreted the mockups](https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=44%3A263&t=LZfj6TMAUPmdfHKA-0) to enable those buttons only after table vars are selected.

nkittur-uga:
@jernestmyers - I spoke to the team and we feel that the Quadrant A values should become active immediately after their corresponding x- or y-axis variable has been selected. Sorry this is different from the mock-ups! And thank you for making the update.

jernestmyers:
I've merged changes to get the disable/enable behavior to work as requested. The QA site's backend is having trouble getting up-to-date with the client code, but in the meantime things can be viewed/tested [on my dev site](https://jmyers.clinepidb.org/ce.jmyers/app/).

nkittur-uga:
Looks great, thanks Jeremy!

PR #1608, opened 2023-02-10
Update/2x2 params 1575

jernestmyers:
Resolves https://github.com/VEuPathDB/web-eda/issues/1575 and I believe the current implementation would obviate https://github.com/VEuPathDB/web-eda/issues/1610.

The [mockups request](https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=44%3A169&t=isizyx4tOql3RUu4-0) that:
1. The X and Y variables/values are stacked
2. The reference values are grayed out until X and Y variables are selected
3. Stratification variables wrap beneath the X and Y variables/values selections

Regarding 1 and 3, I applied these changes to _**all**_ visualizations, not just 2x2 mosaic. I also stacked stratification variable/toggle selectors.

Regarding 2, I relied on CSS to gray out and disable those selections but it raises the need to handle disabling in the `Select` components' APIs in `CoreUI`. I will open an issue in `CoreUI` to do so.

Regarding 3 alone, I find that the toggle's label can get longer than necessary. I found myself wanting to limit its width by passing a `styleOverride` into the `CoreUI` Toggle component, but its API doesn't allow it as far as I can tell. I will open another issue in `CoreUI` to address this limitation, if others agree.

Testing the UI with the updated backend can be done [on my dev site](https://jmyers.clinepidb.org/ce.jmyers/app/).

Screenshots (1: shows 2x2 without wrapping; 2: shows 2x2 with wrapping; 3: shows line plot potentially obviating the aforementioned issue):
image
image
image


dmfalke:
I think the screenshots look pretty good. In the case where the stratification variable selection wraps below the x- and y-axis variable sections, I find it a little hard to process. I wonder if adding some indentation (left margin) to the selectors would help, to make it more clear that they are subordinate to the heading.

I also think the no data toggle should go above the variable selectors in the stratification section, but it's possible that's been discussed before. It just seems a little lost being beneath them.

I'll take a look at the implementation shortly.

src/lib/core/api/DataClient/types.ts
```diff
@@ -441,16 +441,31 @@ export const LineplotResponse = type({
...
-    facetVariable: ZeroToTwoVariables;
-    showMissingness?: 'TRUE' | 'FALSE';
-  };
+  config: TwoByTwoRequestConfig;
+}
+
+export interface MosaicRequestParams
+  extends Omit<TwoByTwoRequestParams, 'config'> {
+  config: MosaicRequestConfig;
 }
```
dmfalke, line 469: Rather than using `Omit`, it might read better to define the Mosaic-related interfaces, and then have the TwoByTwo-related interfaces extend them:

```typescript
interface MosaicRequestConfig {
  outputEntityId: string;
  xAxisVariable: VariableDescriptor;
  yAxisVariable: VariableDescriptor;
  facetVariable: ZeroToTwoVariables;
  showMissingness?: 'TRUE' | 'FALSE';
}

interface TwoByTwoRequestConfig extends MosaicRequestConfig {
  xAxisReferenceValue: string;
  yAxisReferenceValue: string;
};

export interface MosaicRequestParams {
  config: MosaicRequestConfig;
  studyId: string;
  filters: Filter[];
}

export interface TwoByTwoRequestParams extends MosaicRequestParams {
  config: TwoByTwoRequestConfig;
}
```

src/lib/core/components/visualizations/InputVariables.tsx
```diff
@@ -16,14 +16,17 @@ import { Tooltip } from '@veupathdb/components/lib/components/widgets/Tooltip';
...
+
 export interface InputSpec {
   name: string;
   label: string;
   /** Provide a string here to indicate that the input is readonly.
    * The string will be displayed instead of a variable selector.
    */
   readonlyValue?: string;
-  role?: 'axis' | 'stratification';
+  role?: 'axis' | 'stratification' | 'twoByTwoAxis';
```
dmfalke, line 29: I'd like to find a way to avoid hard coding viz-specific values here. It might require refactoring a bit so that the consumer of `InputVariables` can specify `order` and `title`. Otherwise, we will likely end up with a lot of viz-specific stuff in here, as we continue to evolve the layout of various visualizations. What do you think @jernestmyers?

dmfalke (commented):
I have a few preliminary comments. I will continue to review the changes to `MosaicVisualization.tsx`.

dmfalke (approved):
I think this looks great. Nice work!

Issue #1594, opened 2023-02-03
Standalone map: semi-transparent header panel basics


bobular:
Mock-up: https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?node-id=148%3A642&t=kZAYjT6DD5vVW5X1-4

Don't worry too much about the content.

The first two items should probably be a reusable component with `orientation` and `position` (i.e. left or top) props

- [x] semi transparent white panel across the top of the map
- [x] collapse/expand arrow that toggles collapsing to a very thin horizontal strip at the top of the screen
      - similar to the way the [left hand menu will collapse to a thin strip](https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?node-id=241%3A781&t=IcAqDr8QQpuxzeOR-4)
      - don't worry about the styling of the button/icon
- [x] placeholder site icon link to '/' in target='_blank'
       - the true site icon will be figured out in https://github.com/VEuPathDB/web-eda/issues/1593 
- [x] study name and analysis name with edit pencil
- [x] filter chips (with no entity subheadings)
      - UX decision to make: do we allow the panel to expand downwards to accommodate a large number of filter chips?  What happens when it overlaps with the left-hand menu? Can that shift down as the header panel expands?
- [x] some kind of status line showing the most easily obtainable entity count and output entity (we will tweak this in UX)

bobular:
Moved to 'complete' because we're not ready for QA yet.

Issue #1498, opened 2022-12-03
Standalone map: filter panel


bobular:
Requires https://github.com/VEuPathDB/CoreUI/issues/139

This will be a draggable panel showing the

- [ ] entity diagram
- [x] subsetting filters (just as they are in the EDA subsetting tab)

For now, filter chips can stay there, even though they will be repeated in the header.

dmfalke:
The entity diagram still needs to be added.

adnauseum:
> The entity diagram still needs to be added.

On it!

PR #1613, opened 2023-02-14
1594: standalone map semitransparent header panel

adnauseum:
#1594 

https://user-images.githubusercontent.com/24446573/220468584-dfe4d982-0826-41fb-ab0c-2770b1b3a60f.mp4

**TLDW** (Too Long, Didn't Watch). I tried to illustrate the following (not in this order):
* Open/close the header.
* Add/remove filter chips.
   * Show filter chip wrapping.
   * Show no filter chip state.
* Rename the analysis.


src/Header.tsx
```diff
@@ -60,6 +60,7 @@ export default function Header() {
         fontSize: '2.5em',
         fontWeight: 400,
         textAlign: 'left',
+        display: 'none',
```
bobular, line 63: A temporary measure I assume!

What's the plan @dmfalke for the site header? I guess it's not a `web-eda` issue.

adnauseum: Oh yes, this code is not in decent shape right now lol. There's a prop now for hiding site headers that I can use.

adnauseum: Updated in: https://github.com/VEuPathDB/web-eda/pull/1613/commits/a1ab96716bf0e6393dd729bfcce40e3ba9a144cf

jernestmyers (approved):
This looks great to me, both in testing and in code. Nicely done!

src/index.css
```diff
@@ -62,3 +62,19 @@ a,
...
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  display: inline-block;
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
+  padding: 0;
+  position: relative;
+  width: 1px;
+}
```
dmfalke, line 80: Note that this CSS is not included in the published npm package. It is only used in the local dev site (`yarn start`).

dmfalke (approved):
This looks great! See my note about CSS.

Do you think it would be worth wiring the filter chips with the appState `subsetVariableAndEntity` and `isSubsetPanelOpen` properties?

src/index.css
adnauseum: Thank you for catching that! I've relocated this utility in https://github.com/VEuPathDB/web-eda/pull/1613/commits/d82b03707176bcb81959773e08a0a863913d771e.

adnauseum: > Do you think it would be worth wiring the filter chips with the appState subsetVariableAndEntity and isSubsetPanelOpen properties?

I see. These would be the ingredients needed to achieve the following: 
chips

I wasn't sure if we wanted that in this unit of work or in this one: https://github.com/VEuPathDB/web-eda/issues/1498.

I'm fine with either, but do I understand you correctly?

dmfalke: Awesome! :rocket:

PR #1615, opened 2023-02-15
Replace existing service url parameters with a single service url parameter

### LINKED ISSUE #1614: Replace existing service url parameters with a single service url parameter
In the past, we treated different endpoints of the eda service stack as separate, configurable endpoints. Since those services are now deployed as a docker-compose stack, this is not needed. If we want different versions of services, we can deploy the stack locally, using different tags for each service container.

Additionally, the `MapVeuContainer` has the service url hard coded. That value should become a prop.

dmfalke:
fixes #1614 

This PR introduces the following changes
- Use a single prop for the EDA service url base.
- Add a prop to the top-level map component.
- Initialize all api-service clients in root components.
- Use a default export for root components to simplify consumer imports.

Note that this PR will require a major version update, due to the top-level API changes.


src/lib/mapveu/index.ts
```diff
@@ -1 +1 @@
-export { MapVeuContainer } from './MapVeuContainer';
+export { MapVeuContainer, MapVeuContainer as default } from './MapVeuContainer';
```
adnauseum, line 1: Does this make these two import statements equivalent?

```js
import { MapVeuContainer } from '<path>'
import MapVeuContainer from '<path>'
```

dmfalke: It does, and maybe that's not ideal. I was thinking we would support either, but maybe that's not great?

adnauseum: Gotcha! I was just curious. I assumed it was so that we wouldn't have to change as many files.

adnauseum (approved):
🚀

src/lib/mapveu/index.ts
dmfalke: Part of my rationale for the `default` export is so that we don't have to expose the name of the component (I really don't like it). I suppose I could rename the component to something like `MapApp`.

dmfalke:
@bobular I just pushed a change to rename the dir `mapveu => map`. Any objections?

bobular:
> @bobular I just pushed a change to rename the dir `mapveu => map`. Any objections?
> 
> Edit: This will allow us to `import EdaMap from '@veupathdb/eda/lib/map';`

No objections here. I don't see `EdaMap` anywhere though - was that a typo?

dmfalke:
> > @bobular I just pushed a change to rename the dir `mapveu => map`. Any objections?
> > Edit: This will allow us to `import EdaMap from '@veupathdb/eda/lib/map';`
> 
> No objections here. I don't see `EdaMap` anywhere though - was that a typo?

No, it was just an example of how we might import this in a website. Thanks for the feedback!

PR #1623, opened 2023-02-16
resolving lineplot binning with axis range

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1544. Note that since there existed banner control behavior-related PR (https://github.com/VEuPathDB/web-eda/pull/1589) that involves many changes, this PR is branched out from that PR (`lineplot-banner-no-data` branch). But if #1589 is approved and merged, then I will eventually update this PR to be merged into main branch later.

Tried various approaches, e.g., modifying onChangeHandlerFactory, etc., but arguably the current implementation is the simplest one we can resolve the issue. I didn't notice specific issue for both numeric and date xAxisvariable.


src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2122,10 +2120,12 @@ function getRequestParams(
   const viewport =
     vizConfig?.independentAxisRange?.min != null &&
     vizConfig?.independentAxisRange?.max != null
-      ? {
-          xMin: String(vizConfig?.independentAxisRange?.min),
-          xMax: String(vizConfig?.independentAxisRange?.max),
-        }
+      ? vizConfig.useBinning
```
bobular, line 2123: If we do this we don't get the correct behaviour.
When binning we must send the viewport to the back end.

I've tried reverting this change and the original issue remains fixed (binning turns off and on correctly).

The only problem is the truncation warnings, which is a [known issue](https://github.com/VEuPathDB/web-eda/issues/1295) and probably something we should consider in the back end.

So I think with some more testing we could go forward with the original code here.

bobular (changes requested):

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: Thank you for your insight @bobular Actually I left my comment yesterday but somehow it is gone so I am leaving this comment again. 

Please correct me if I am wrong. If I understand correctly about the role of viewport, it is for automatically adjusting binWidth. But if useBinning turns on by user, then we respect user's choice about binWidth, so I thought that we don't need to worry about automatic binWidth in this case. An advantage to have the useBinning condition for the viewport is that it actually resolves truncation warning issue. 

But if you think I misunderstand this viewport one, then I can remove the useBinning condition from viewport as you suggested, which would not be a big deal.

bobular: Hi @moontrip 

I think what we want to implement is the behaviour we currently have in the histogram visualization:

* If the user _has not_ manually entered a binWidth, when changing the custom x axis range, the binWidth will be determined automatically by the back end.
* If the user _has_ entered a binWidth, this is honoured for all custom ranges (even if this doesn't really make sense/returns way too many bins, etc)

In order to make the back end automatically determine the binWidth, we have to send the viewport and the binSpec has no `value` property.

```
{
	"binSpec": {
		"type": "binWidth"
	}
}
```

moontrip: Hi @bobular 👋 

Thank you for your detailed information. I left my comment below to your comment in the following (starting with DK):
> Hi @moontrip
> 
> I think what we want to implement is the behaviour we currently have in the histogram visualization:
> 
> * If the user _has not_ manually entered a binWidth, when changing the custom x axis range, the binWidth will be determined automatically by the back end.

DK) Yes this PR basically fixed incorrect binWidth when vizConfig.useBinning is false, as you probably observed in your previous test 

> * If the user _has_ entered a binWidth, this is honoured for all custom ranges (even if this doesn't really make sense/returns way too many bins, etc)

DK) So, this is the case when vizConfig.useBinning is true. Although current PR used a condition at viewport so no viewport is sent to the backend, the frontend still sends binSpec/binWidth value based on user's input/choice. Thus, the backend honors binSpec/binWidth value and no automatic decision from the backend is made. Thus no problem anyway?

Accordingly, purely based on above two conditions, this PR seems to be valid to fix binWidth issue as well as truncation warning?

bobular: Hi @moontrip 

Firstly, I need to apologise about my assumption that the lineplot back end was doing "semantic zooming" (choosing the best `binWidth` based on the viewport if no `binWidth` is supplied). I did some testing and also checked with Danielle and this is **not** the case, and there are no immediate plans to implement it on the back end (though it would be easy to do so).

While I was still believing that we wanted semantic zooming, I did some investigations. I found that the `binSpec` formulation in `getRequestParams` was almost impossible to understand, so I simplified it:

https://github.com/VEuPathDB/web-eda/compare/temp-lineplot-binning?expand=1  (this is a diff with `main`)

My version fixes the "binning stays on" bug* and supports back end semantic zooming if that is ever implemented. However, it triggers a lot more back end requests which are wasteful when there's no semantic zooming. It also doesn't solve the truncation issues (which we haven't solved for histogram either). I'll test your branch carefully and report back in the next message.

* the trick was to send `binWidth: 0` when binning is off (and "1 day" for dates). This is special (undocumented?) back end behaviour to turn off binning, and this is different compared to histogram (where binning is always on, of course).

bobular: Your branch works correctly based on my tests. At least with respect to x-axis binning and full/custom ranges.

I have noticed that when binning is off, we are still sending viewport specs to the back end, which are ignored by the back end when not binning. This will be confusing to future maintainers, so we should just remove the viewport from the request and make a comment about the back end not doing semantic zooming. This is the one small change request I am making.

HOWEVER...

The client code figures out the y-axis min and max from the response, so the Auto-Zoom y-axis currently is set for the whole data and does not respond to custom x-axis range changes.  With my temporary branch, a whole new request is made and the y-axis auto range is adjusted:

image

Whereas in your branch the truncation warnings are correct but the y-axis auto-range is calculated globally.

image


So, in summary, I think we can go with your approach, but I will check in the eda ux channel what people think about the y-axis auto range issue.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2078,10 +2076,12 @@ function getRequestParams(
   const viewport =
```
bobular, line 2076: After we hear back from the UX crew, I think we can remove the viewport from the request in all circumstances.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
bobular: EDA UX discussion says it's OK to continue with the auto y-range NOT responding to the x-range.

Scatterplot (also no semantic zoom) does this already!

moontrip:
@bobular Thank you for your detailed information 👍 
As for the y-axis auto-range, if I remember correctly, it is the same behavior to other Vizs like Scatter and Lineplot Vizs: y-axis auto range does not consider x-axis range? For example, the difference between Full and Auto-zoom in the Scatterplot Viz's Y-axis range is that the former relies on backend's range/displayRange min/max while the latter is computed based on full data.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: @bobular I didn't test the case without viewport, but if we completely remove it, then I think that binWidth won't automatically change for no binning case (vizConfig.useBinning: false)?

bobular: I'm not sure what you mean by "binWidth won't automatically change for no binning case"?

When not binning, do we care about binWidth?

moontrip: @bobular If I remember correctly, for the case of vizConfig.useBinning: false, binWidth is automatically changed/adjusted if X-axis custom range is changed (for a certain range only though)?

moontrip: @bobular viewport is removed

bobular: Sounds like my imagined semantic zooming on lineplot x axis!

bobular (approved):

bobular:
Thanks DK - sorry it was a long haul!

moontrip:
@bobular No problem! Thank you for your kind attention on this matter 👍

Issue #1384, opened 2022-09-19
Mosaic plot (2x2) - statistics tab frontend updates


danicahelb:
For the 2x2 mosaic plot viz tool, mockups for the statistics tab can be found here: https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=123%3A233

dmfalke:
Related to https://github.com/VEuPathDB/CoreUI/issues/114 and #1383

moontrip:
@danicahelb May I ask a couple of questions regarding this ticket and relevant one (https://github.com/VEuPathDB/web-eda/issues/1383)?

First, I made a collapsible banner and screenshots can be found in https://github.com/VEuPathDB/CoreUI/pull/129#issue-1453839015, which is a mock up for #1383. But from my understanding, there are no Quadrant input variable yet, thus the banner won't be included till the Quadrant input variable is made, right? (Quadrant work is not part of #1383 nor #1384 (this ticket)).

Second, this is a question for this ticket, statistics tab updates: a) If I understand correctly, we will also have a separate banner for this tab only. Can you please confirm this? b) the scoop of this work is to have statistic table too? It is currently shown as "Coming soon".

danicahelb:
Hi @moontrip thanks for raising this topic!

I apologize, this key ticket got lost in the shuffle! https://github.com/VEuPathDB/EdaDataService/issues/130 (the ticket title, major feature & release info are now up to date)

Yes, the stats tab will have a separate banner. See: https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=123%3A234&t=eO7GflQwHS5GJltr-0

We are still finalizing the stats tab and need to make decisions on:
1. what stats to include
2. table layout
3. verbiage for the banner

Please pause work on this ticket until the stats tab has completed review by the EDA UX group

danicahelb:
styling ticket for stats tab: https://github.com/VEuPathDB/web-eda/issues/1253

danicahelb:
backend tickets to calculate the various statistics: https://github.com/VEuPathDB/EdaNewIssues/issues/533 & https://github.com/VEuPathDB/EdaNewIssues/issues/534

We would like to roll out these stats for b62:

1. Chi-square
2. Fischer's exact
3. Prevalence
4. Odds Ratio
5. Risk Ratio

[...]

PR #1625, opened 2023-02-17
Mosaic 2x2 stats table

moontrip:
This PR intends to resolve https://github.com/VEuPathDB/web-eda/issues/1384. 

As a first step, a groundwork for typing the new Mosaic 2x2 stats table (backend response) is made. Since backend response should be established for this work, this PR is branched out from Jeremy's nice PR (https://github.com/VEuPathDB/web-eda/pull/1608) (`update/2x2-params-1575` branch). So, for now, the base branch is set to `update/2x2-params-1575` but it may be changed to `main` branch later if `update/2x2-params-1575` is merged. 

It seems to work fine to see correct stats table props with this PR, so perhaps we just need to make a table (with requested styles) and map the props to corresponding table cells. Just in case, I am leaving a screenshot of the requested table format in the following:

mosaic 2x2 stat table itself


moontrip:
Mosaic 2x2 Stats table is now implemented. Screenshots for both with or without facet variables are attached in the following:

- without facet variable
mosaic-2x2-stats-table-single

- with facet variable
mosaic-2x2-stats-table-facet

jernestmyers:
Looking into this now, but in the meantime we should [update this code](https://github.com/VEuPathDB/web-eda/pull/1625/files#diff-556a996edbde2eb7605da318eb9e2d90be4af8ee28a245e9193524488a1960b8L793-L797) now that we have the stats tab.

moontrip:
> Looking into this now, but in the meantime we should [update this code](https://github.com/VEuPathDB/web-eda/pull/1625/files#diff-556a996edbde2eb7605da318eb9e2d90be4af8ee28a245e9193524488a1960b8L793-L797) now that we have the stats tab.

Thanks @jernestmyers By the way, can you please elaborate what the update means?

jernestmyers:
Yes, we currently don't render "Please select all *required parameters" in the 2x2 stats tab [because of this code](https://github.com/VEuPathDB/web-eda/blob/f16ff0e4a35e43821660e9f95f8eaf7f3f77861b/src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx#L793-L797).

So we just need to remove the comment and the `isTwoByTwo` conditions like so:
```js
showRequiredInputsPrompt = { !areRequiredInputsSelected }
```

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -802,55 +812,192 @@ function MosaicViz(props: Props<Options>) {
...
-            <th>P-value</th>
-            <td className="numeric">
-              {props.pValue != null ? quantizePvalue(props.pValue) : 'N/A'}
+            <td className="stats-table_leftmost-cell">
+              Association between 2 categorical variables
+            </td>
+            <td className="stats-table_middle-cell">
+              <b>Chi-squared (df=1)</b>
+            </td>
+            <td>{props.chiSq?.value ?? 'n/a'}</td>
```
jernestmyers, line 869: Looks like we're preferring lowercase for `n/a` in the mockups included at the top of this PR. As a result, let's [update `ContTableStats` to use lowercase as well](https://github.com/VEuPathDB/web-eda/pull/1625/files#diff-556a996edbde2eb7605da318eb9e2d90be4af8ee28a245e9193524488a1960b8L888-L901).

moontrip: Thank you for your detailed explanations and suggestions @jernestmyers 👍 I did address your feedbacks so can you please check if they are what you had in mind? https://github.com/VEuPathDB/web-eda/pull/1625/commits/bc3462accb5d146ff18dc2d676e7ef78b8f95604

jernestmyers (approved):
I'm going to approve, but I do have some other thoughts for possible improvement.

1. When should we create a new component file? Our viz implementation files get rather large and complex. Would it help to break things off into their own file? In this case, couldn't `TwoByTwoStats` be its own component? This could be a topic at the next frontend dev meeting.
2. I'm left wondering if there's a more DRY way to render both mosaic tables, though it may not be worth the debt since the styling could be tricky.
3. I prefer the more uniform spacing that the mockup displays for the actual data than what's currently implementing. The center (`95% CI`) column's data feels a little jumbled because it is so close to the `Value` column's data. I think we should add some space between the `Value` and `95% CI` columns in a way that tries to match the space between `95% CI` and `P-value`.

Overall, very nice work!

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -801,10 +801,7 @@ function MosaicViz(props: Props<Options>) {
         controlsNode={controlsNode}
         tableGroupNode={tableGroupNode}
         // statistics tab is disabled in 2x2, so no need to prompt for required inputs
```
jernestmyers, line 803: Let's remove this comment too, since the Stats tab won't be disabled anymore.

moontrip:
> I'm going to approve, but I do have some other thoughts for possible improvement.
> 
> 1. When should we create a new component file? Our viz implementation files get rather large and complex. Would it help to break things off into their own file? In this case, couldn't `TwoByTwoStats` be its own component? This could be a topic at the next frontend dev meeting.
> 2. I'm left wondering if there's a more DRY way to render both mosaic tables, though it may not be worth the debt since the styling could be tricky.
> 3. I prefer the more uniform spacing that the mockup displays for the actual data than what's currently implementing. The center (`95% CI`) column's data feels a little jumbled because it is so close to the `Value` column's data. I think we should add some space between the `Value` and `95% CI` columns in a way that tries to match the space between `95% CI` and `P-value`.
> 
> Overall, very nice work!

@jernestmyers Thank you for your detailed comments. For your #1, yes I tend to agree that the stats table may become a separate component out of Mosaic Viz, but I will leave it as a technical debt :). For #2, yes that specific styling prevents from minimizing DRY. Finally, yes I noticed that too. It is because value and P-value are right-aligned while CI is center-aligned. Perhaps your suggestion may work by having some space. Will test it. 👍

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
moontrip: yep will remove it 👍

moontrip:
@jernestmyers I have added some space for 95% CI and I think that it looks better than before: a screenshot is attached in the following.

mosaic-2x2-stats-table1

PR #1629, opened 2023-02-17
Update peerDependencies

dmfalke:


moontrip (approved):
@dmfalke Thank you for handling this 👍

Issue #1383, opened 2022-08-31
Mosaic plot (2x2) more information box for parameters - finalize styling & implement


danicahelb:
As discussed in 8/29/2022 EDA UX meeting, we need a new explanatory information box that allows users to read help / additional information so the sure can refer to it while they are making parameter selections for visualizations or apps.

In ClinEpi, we will use the explanatory info box for the 2x2 mosaic viz tool. The box will be open by default until all required plot parameters are selected, then will auto-close (but the user can click to expand/reopen)

In mBio, an explanatory information box that is open during parameter selection is also needed for many of the mbio apps

The Styling WG should meet to discuss the styling (including header verbiage) of the explanatory info box.

Danica mocked up the function of the box in action in the 2x2 mosaic tool, but styling needs help! https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=20%3A116

Changes to the 2x2 UI cannot happen until decisions on the styling of the box have been made

dmfalke:
I think this is related to https://github.com/VEuPathDB/CoreUI/issues/114, in that we might want to first implement a "CollpsibleSection" component in CoreUI, and use that for this info box.

jernestmyers:
We finalized our style decisions [in this Slack thread](https://epvb.slack.com/archives/C02R14DUYLX/p1665586919851079). Final mockups can be found [here](https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=139%3A233&t=sRnMJ2CGeNmUYsMM-1).

danicahelb:
@moontrip this ticket and https://github.com/VEuPathDB/EdaDataService/issues/130 should be rolled out to QA at the same time

danicahelb:
Discussed in 1/23/2023 EDA UX meeting

@moontrip we made a few (hopefully easy!) changes to the UI:

1. We want to move the position of the more information box to _ABOVE_ the selectors
2. We want to change the verbiage of the first line to: `Learn how to set up a 2x2 contingency table in order for statistics to be calculated correctly. Read less...`

<img width="711" alt="image" src="https://user-images.githubusercontent.com/25450900/214135209-5de0f1a0-559e-4de9-8fc3-305569106111.png">

danicahelb:
**Styling decisions approved**
- background color - blue, like banners
- border - gray, like other containers
- toggle collapse/expansion icon - single arrow/caret
[...]

Issue #1574, opened 2023-01-11
Mosaic plot (2x2) more information box for stats tab - finalize styling & implement


danicahelb:
In ClinEpi, we will use the explanatory info box for the 2x2 mosaic viz tool statistics tab. The box will be closed by default but the user can choose to expand / open the box to read more about which statistics are appropriate to use for their analysis. this info box should be similar in styling to the 2x2 mosaic parameter selection screen

the explanatory information box allows users to read help / additional information so the sure can refer to it while they are making parameter selections for visualizations or apps.

mockups here: https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=317%3A623&t=ZCxlWaPOOG1AbgKr-0

danicahelb:
Hey @moontrip we finalized the more info box for the stats tab in the 1/23/2023 EDA UX meeting

See final mockups, which contain approved verbiage: https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=403%3A725&t=sDSejTfcwyglUYu9-0

**For styling:**
- background color - blue, like banners
- border - gray, like other containers
- toggle collapse/expansion icon - single arrow/caret
- info icon - include dark blue info icon

More info here: https://www.figma.com/file/Su1eNm3Bby85CfrnX42L7q/2x2-table-UI?node-id=139%3A233&t=C7dsTpu6Y9mLsmoz-0

nkittur-uga:
Looks good!
<img width="1079" alt="Screenshot 2023-02-27 at 4 52 29 PM" src="https://user-images.githubusercontent.com/67710160/221694558-b2ecd16c-f118-4556-8c4f-ab9baa243941.png">

PR #1636, opened 2023-02-22
Mosaic 2x2 banners

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1383 and https://github.com/VEuPathDB/web-eda/issues/1574. Screenshots are attached in the following:

a) showing top and stats table banners at once (unfolded)
mosaic-2x2-banner00

b) showing top banner content after folded
mosaic-2x2-banner01

c) showing stats table banner content after folded: no facet variable
mosaic-2x2-banner02

d) a case with a facet variable (unfolded)
mosaic-2x2-banner03

### DIFF
--- a/src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
@@ -66,6 +66,7 @@ import { useInputStyles } from '../inputStyles';
 import { ClearSelectionButton } from '../../variableTrees/VariableTreeDropdown';
 import { Tooltip } from '@veupathdb/components/lib/components/widgets/Tooltip';
 import { MEDIUM_GRAY } from '@veupathdb/components/lib/constants/colors';
+import Banner from '@veupathdb/coreui/dist/components/banners/Banner';
 
 const plotContainerStyles = {
   width: 750,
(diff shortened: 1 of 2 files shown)


src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -736,272 +771,722 @@ function MosaicViz(props: Props<Options>) {
...
+}) {
+  return props != null ? (
+    <>
+      {props.facetVariableDetails != null ? (
+        ''
+      ) : (
+        <div
+          style={{
+            marginLeft: '0.9em',
+            width: '750px',
```
jernestmyers, line 1280: I wonder if we want the width to increase when faceted?

moontrip:
Thank you for your suggestions concerning the width of stats' table banner for the case of facet variable, @jernestmyers @asizemore Actually I was gauging whether it should be full width or not so your suggestion convinced me of making a decision. I will update it 👍

moontrip:
> 
> Also, I think we should open a follow-up ticket in `CoreUI` to change the cursor of the `collapsibleIcon` to be `pointer`, which would better indicate that it's a clickable button.
> 
> [Here's the relevant code](https://github.com/VEuPathDB/CoreUI/blob/e9686d345ec809e74843e65be91491bc9fe3b55e/src/components/banners/Banner.tsx#L240-L255) in the `Banner` component.

Definitely I like the idea @jernestmyers: I was thinking the case before but didn't have time to do that. Perhaps you can easily handle it at a follow-up ticket in the near future 😄

jernestmyers:
> Definitely I like the idea @jernestmyers: I was thinking the case before but didn't have time to do that. Perhaps you can easily handle it at a follow-up ticket in the near future 😄

Yes, I'll open the issue now and try to get to it soon!

moontrip:
@jernestmyers @asizemore As suggested, I made a full width of stats table banner when a facet variable exists. Here is a screenshot:

mosaic-2x2-banner04

PR #1643, opened 2023-02-22
Enable quadrant A selections when its corresponding table var is selected

jernestmyers:
Responds to [these comments](https://github.com/VEuPathDB/web-eda/issues/1575#issuecomment-1440471807) in https://github.com/VEuPathDB/web-eda/issues/1575 (comments shown below for ease):
>@jernestmyers - I spoke to the team and we feel that the Quadrant A values should become active immediately after their corresponding x- or y-axis variable has been selected. Sorry this is different from the mock-ups! And thank you for making the update.

### DIFF
--- a/src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
@@ -99,7 +99,7 @@ const modalPlotContainerStyles = {
   margin: 'auto',
 };
 
-const twoBytwoInputStyle: React.CSSProperties = {
+const twoByTwoInputStyle: React.CSSProperties = {
   display: 'grid',
   gridTemplateColumns: '115px auto',
   marginBottom: '0.5em',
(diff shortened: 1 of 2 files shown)


src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -616,22 +616,20 @@ function MosaicViz(props: Props<Options>) {
...
-  const twoByTwoQuadrantStyle: React.CSSProperties | undefined = !isTwoByTwo
-    ? undefined
-    : {
-        ...twoBytwoInputStyle,
-        pointerEvents: areQuadrantSelectionsDisabled ? 'none' : undefined,
-        opacity: areQuadrantSelectionsDisabled ? 0.5 : 1,
-      };
+   * TEMPORARY:
+   * CoreUI's selection components need to be updated to enable disabling them. We lose the ability to convey messages via tooltips
+   * and cursors when we use pointerEvents: 'none'
```
asizemore, line 621: do we have an issue for this already in coreui?

asizemore (approved):
Makes sense to me! Also works great!

(tested in mbio)
Screen Shot 2023-02-22 at 3 53 42 PM

src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
jernestmyers: Yes! https://github.com/VEuPathDB/CoreUI/issues/142

PR #1645, opened 2023-02-23
Bump web-components version for react-leaflet-v3

moontrip:
This is for upgrading web-components version concerning react-leaflet-v3. A good thing is that I did not experience the issue of npm-pack-here at web-eda (e.g., https://github.com/VEuPathDB/web-components/pull/445#issue-1563388756) after merging web-components (@bobular). 

I have extensively tested Mapviz with the new react-leaflet-v3, but did not notice a specific issue.


dmfalke (approved):
:rocket:

PR #1646, opened 2023-02-23
Prevent job deps and status from becoming out-of-sync

### LINKED ISSUE #1600: Do not autostart jobs on configuration change
Observed that in an app (seen in multiple apps, both viz types), and i create a visualization, if i go to change params to  a job that has not started, i go back to the original params, go back to the not-started params, that job is already done! So a job got autostarted somehow.

More specifically,
1. Choose app, say alpha div box
2. Choose params and run computation if necessary
3. make plot
4. Choose params that give a not started status
5. Go back to params used for the plot. Pause a moment.
6. Go back to the params in (4). You'll see the job is complete!

dmfalke:
fixes #1600

This PR employs a technique described in https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes.

The previous implementation had the job status, which is tracked with a `useState` hook, being set to `undefined` inside of a `useEffect` hook callback function. Callback functions passed to a `useEffect` hook are scheduled to be called after the current render phase is complete. This means that the component would render once with new compute job dependencies, but an old job status.

The implementation in this PR moves that logic to outside of the `useEffect` callback, which ensures that the component will render with both the new compute job dependencies, and a job status of `undefined`, rather than the old job status.

### DIFF
--- a/src/lib/core/components/computations/ComputeJobStatusHook.ts
+++ b/src/lib/core/components/computations/ComputeJobStatusHook.ts
@@ -1,10 +1,11 @@
 import { delay } from '@veupathdb/wdk-client/lib/Utils/PromiseUtils';
 import { useCallback, useEffect, useRef, useState } from 'react';
+import { isEqual, omit } from 'lodash';
 import { JobStatusReponse } from '../../api/ComputeClient';
 import { useComputeClient, useStudyMetadata } from '../../hooks/workspace';
 import { Analysis, NewAnalysis } from '../../types/analysis';
-import { Computation } from '../../types/visualization';
 import { plugins } from './plugins';
+import { Computation } from '../../types/visualization';
 
 export type JobStatus = JobStatusReponse['status'] | 'requesting';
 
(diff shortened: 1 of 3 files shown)


adnauseum (approved):
Subtle indeed.

asizemore (approved):
Very interesting! I learned a lot, thank you!

Might be fun to discuss this in next Wednesday's meeting if others are interested!

src/lib/core/components/computations/ComputeJobStatusHook.ts
```diff
@@ -45,8 +66,6 @@ export function useComputeJobStatus(
   useEffect(() => {
     // Track if effect has been "cancelled"
     let cancelled = false;
-    // Clear existing job status
-    setJobStatus(undefined);
```
bobular, line 49: So the key part is not doing this inside the useEffect, but instead above at line 54?

bobular (approved):
Thanks for the link to that page - very interesting. I hope I haven't been too much of a useEffect junkie up til now...

src/lib/core/components/computations/ComputeJobStatusHook.ts
dmfalke: Yes! This is the key change for fixing the bug. The rest is either cleanup or code to support detecting when we need to set the job status to `undefined`.

PR #1647, opened 2023-02-24
reset banners when resetting to default

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1503#issuecomment-1442287274 and https://github.com/VEuPathDB/web-eda/issues/1503#issuecomment-1442288919. In short, when a user clicks "Reset to default" while a banner is shown, the banner should disappear.


dmfalke (commented):
Can you explain why we need to use state for determining if we show the banner? I would expect that we can determine that directly from the vizConfig settings. I didn't review earlier work in the area, so I may be missing some context. Thanks!

moontrip:
@dmfalke  Thank you for offering insight 👍 The initial implementation of those banners at Viz was made quite a long time ago, so I needed some time to remind myself why I introduced those useStates 😅 It was because initially there were two banners per each axis: in other words, banner had four different contents for the condition of toggles (x log-scale, x binning,, y log-scale,, and y error bar) . Thus it was necessary to distinguish banner among those four toggles. That was the reason why I introduced four different useStates initially.

But more recently, it was simplified into two banners content only,, one for axis regardless of the order of the toggling. Honestly, I didn't think of this carefully, but in this case, we may not need those useStates as you questioned. Perhaps we can leave this as a tech debt or I may try to test it next week as I'm off from VEuPathDB tomorrow (Friday).

dmfalke:
Thank you for the information. If it's not too much extra effort, I think it would be nice to address it in this PR. It's fine for it to wait until next week. Does that work for you?

moontrip:
> Thank you for the information. If it's not too much extra effort, I think it would be nice to address it in this PR. It's fine for it to wait until next week. Does that work for you?

@dmfalke Yes I will try it early next week 👍 Have a great weekend!

moontrip:
@dmfalke I have tried to remove those useStates and tested extensively. I couldn't find any negative side effects after the removal, which is good, so can you please test it at your side as well? (including Reset to default as originally intended in this PR)

dmfalke (approved):
Looks great, thanks!

moontrip:
Thanks @dmfalke for your review and tests 👍

Issue #45, opened 2021-04-14
Visualization: Mosaic plot (2x2 table)


dmfalke:
**API:** https://veupathdb.github.io/EdaDataService/api.html#resources:/apps/pass/visualizations/twobytwo:post
**Plot component:** https://veupathdb.github.io/web-components/?path=/story/plots-mosaic--two-by-two

PR #1649, opened 2023-02-27
replace 'distance' with 'dissimilarity' in beta div

### LINKED ISSUE #1628: Update beta diversity "distance" to "dissimilarity"
Technically, we are calculating dissimilarity values (do not respect triangle inequality) instead of distances.

asizemore:
Resolves #1628 

Replaces "distance" with "dissimilarity" in the beta div app.

Relies on compute service [PR #45](https://github.com/VEuPathDB/service-eda-compute/pull/45)

Note: If approved, i'm planning to call this a minor version upgrade since the api changed but beta div is not in production yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "4.0.21",
+  "version": "4.1.0",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


asizemore:
@dmfalke courtesy tag :)

Issue #1477, opened 2022-11-14
Bug: Faceted Mosaic rxc Plot not showing hover-over text


nkittur-uga:
To recreate: GEMS1 Case Control Study-
Visualize tab -> Mosaic plot, RXC table
X-axis: Age group
Y axis: Breastfeeding status 
Facet: Case or control participant

When you click on either of the faceted plots to enlarge them, the hover-over text should show the proportions for each cell. But this text is not appearing. 
![Screen Shot 2022-11-14 at 9 22 03 AM](https://user-images.githubusercontent.com/67710160/201688442-6e50bbd6-746c-40f5-90b3-edf333b6b613.png)

nkittur-uga:
Issue is fixed. Mosaic plots now have interactive mode for both 2x2 and rxc. QA is complete. 

(However, unfaceted plots don't have interactive mode. I am opening a separate ticket for that issue.)

moontrip:
@nkittur-uga I think that for the unfaceted plot, the interactive plot is disabled because of the presence of the full-screen plot when pressing each unfaceted plot. Accordingly, when doing mouse-over on the unfaceted plot, you can see a tooltip (forgot about the exact wording) like click to expand/enlarge instead.

moontrip:
@nkittur-uga I saw your new ticket. So, nevermind my comment above: I am sorry that I misunderstood your question about it. I will look into it

PR #1650, opened 2023-02-27
enable interactive mode for mosaic facet

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1477. Tried to find the reason at other components such as Mosaic, facetedMosaicPlot component (at web-components), etc., but I realized it later that interactive prop was disabled for faceted cases. Thus simply removed it. Actually this happened for both Mosaic RxC as well as 2x2. 

Here is an screenshot showing mouseover tooltip at full screen modal with the same example case at the ticket.
mosaic rxc faceted interactive 1477

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "4.0.19",
+  "version": "4.0.20",
   "dependencies": {
     "debounce-promise": "^3.1.2",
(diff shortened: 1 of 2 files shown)


chowington (approved):
I've never seen a more perfect PR 👍

moontrip:
@chowington Thank you for your review and praise 😄

PR #1651, opened 2023-02-28
Update README.md with service URL changes

bobular:


dmfalke (approved):
Thanks for doing this! No need to update the version, imho

PR #1653, opened 2023-02-28
fix: hide My Subset downloads if assay entity present

### LINKED ISSUE #1652: Bug: show user-friendly message when subset results table preview cannot be generated
I tried to download this:
image

and got this:
image

asizemore:
Resolves #1652 

The assay entities were giving us such a headache that we just decided to remove the whole My Subset part of the downloads tab. See slack thread in mbio channel for more details. 

I thought about using the entity ids, but then we'd have a hardcoded list of entity ids in the code which didn't seem good either. As long as clinepi doesn't have any assay entities (they don't yet!) then this fix will be okay. Still, this is a temporary fix for a problem that requires a harder solution (allowing tables with >1000 vars).

### DIFF
--- a/src/lib/workspace/DownloadTab/index.tsx
+++ b/src/lib/workspace/DownloadTab/index.tsx
@@ -249,15 +249,26 @@ export default function DownloadTab({
     );
   }, [WDKStudyReleases, downloadServiceStudyReleases]);
 
+  // We can't show (or sometimes even create!) my subset tables for entities with many vars. Assays in mbio often
+  // have >1000 vars. Until we can handle these cases, hide the whole My Subset section if any entites
+  // are 'assay' entities.
+  const showMySubset =
+    enhancedEntityData.filter((entity) => entity.displayName.includes('assay'))
+      .length === 0;
+
   return (
     <div style={{ display: 'flex', paddingTop: 10 }}>
       <div key="Column One" style={{ marginRight: 75 }}>
         {dataAccessDeclaration ?? ''}
-        <MySubset
-          datasetId={datasetId}
-          entities={enhancedEntityData}
-          analysisState={analysisState}
-        />
+        {showMySubset ? (
+          <MySubset
+            datasetId={datasetId}
+            entities={enhancedEntityData}
+            analysisState={analysisState}
+          />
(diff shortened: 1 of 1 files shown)


src/lib/workspace/DownloadTab/index.tsx
```diff
@@ -249,15 +249,26 @@ export default function DownloadTab({
     );
   }, [WDKStudyReleases, downloadServiceStudyReleases]);
 
+  // We can't show (or sometimes even create!) my subset tables for entities with many vars. Assays in mbio often
+  // have >1000 vars. Until we can handle these cases, hide the whole My Subset section if any entites
+  // are 'assay' entities.
+  const showMySubset =
```
jernestmyers, line 255: Thought I'd pass along some feedback I got early on when doing something similarly, even though your implementation is perfectly fine since the `enhancedEntityData` array is always small.

For a large array, it would be better to leverage a method like `find` since it'll stop iterating through the array as soon as it satisfies the provided callback function. Of course, if it cannot find an array entry that satisfies the condition, then it'll have to iterate through the entire array. But a method like `filter` will always iterate through the entire array.

jernestmyers (approved):

src/lib/workspace/DownloadTab/index.tsx
asizemore: Oh clever!! I'll implement that change. Thank you so much!

dmfalke:
I wonder if another solution is to check `entity.variables.length > 1000`? That way, we don't have to hard code anything. It does mean that downloads will work for studies that don't have such entities. Do all assay entities have > 1000 variables?

asizemore:
@dmfalke that would also work! But the point i get stuck on is exactly what you said - then My Subset will show for some studies but not others. I worry that without an explanation to the user that might be confusing?

asizemore:
> Do all assay entities have > 1000 variables?

No, all assay entities do not have >1000 variables, sadly

asizemore:
in slack, decided we're happy with the error only appearing in the view table part of the download popup (@jernestmyers 's great work). So i'll close this PR.
Sorry for the confusion!

PR #1655, opened 2023-02-28
Skip job status polling if compute config is invalid or undefined

dmfalke:
This PR introduces the following changes:
1. Fixes requesting a job status for apps without a compute (such as the `pass` app).
2. Fixes using an app name, instead of a compute name, to request a job status.

### DIFF
--- a/src/lib/core/components/computations/ComputeJobStatusHook.ts
+++ b/src/lib/core/components/computations/ComputeJobStatusHook.ts
@@ -14,7 +14,8 @@ export type JobStatus = JobStatusReponse['status'] | 'requesting';
  */
 export function useComputeJobStatus(
   analysis: Analysis | NewAnalysis,
-  computation: Computation
+  computation: Computation,
+  computeName: string,
 ) {
   const computeClient = useComputeClient();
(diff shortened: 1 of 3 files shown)


src/lib/core/components/computations/ComputeJobStatusHook.ts
```diff
@@ -75,7 +80,6 @@ export function useComputeJobStatus(
 
     // Fetch the job status and update state
     async function updateJobStatus() {
-      if (!computePlugin.isConfigurationValid(jobStatusDeps.config)) return;
```
asizemore, line 78: okay so the idea is that if the compute config isn't valid, we shouldn't ever even enter the loop. This makes sense

asizemore (approved):
looks good!

moontrip:
I have a chance to test this with ClinEpi and it seems to work fine for me without throwing that popup error.

dmfalke:
Thanks for the feedback. I added another fix for a different regression. The wrong value was being passed to the compute service as the compute name. That should be fixed now.

asizemore:
briefly tested the latest on mbio and the abundance app works great!

Issue #1640, opened 2023-02-16
stand alone map header: filter chips vs show all filters button


danicahelb:

danicahelb:
Discussed in 2/27 EDA UX meeting.

Extremely rough draft summarizing the outcome of our decision about handling filter chips:
<img width="579" alt="image" src="https://user-images.githubusercontent.com/25450900/221669200-07d40401-bf52-4570-9748-86065993b81d.png">


- Other than placement of the container, we will align how filter chips are handled between EDA sites and the stand-alone map (see additional notes below)
- Filter chips will be given a container (single line) in the header
- We will NOT limit the size of a filter chip; if a variable has a long label, the filter chip will be wide. 
- Cap filter chips at a the number that fit into the container. Because the filter chips can be of varying widths, there will NOT be a standard number of chips that fit in the container
- Provide a button to open the filters panel on the left-most side of the container. This button should also indicate the number of filters. `Show all XXXX filters`, where `XXX` indicates the number of filters the user has applied

For the map, on the righthand side, provide numbers for:
- The total number of samples in the study.
- The total number of samples after subsetting.
- The total number of samples visible at the user’s current viewport dimensions.


Try to be consistent with verbiage between SAM & EDA
- “Subset” instead of “Filtered data” (“filter” should not be used as an adjective to describe the data)
- “All” instead of “Total”

bobular:
Hi @danicahelb and @adnauseum 

This number
> The total number of samples visible at the user’s current viewport dimensions

can potentially be derived in two ways (and perhaps presented in both ways?)

Firstly I will replace "number of samples" with "count of entities for the relevant entity" - where relevant means the entity of the chosen overlay variable and soon maybe also a user-selected entity of choice (especially if we allow no overlay variable to be selected).

1. The count of entities for the relevant entity
2. The count of entities for the relevant entity that have data for the chosen overlay variable

1 and 2 would be equivalent if we always showed an additional "no data" segment in the donut charts and legend.

In https://veupathdb.atlassian.net/wiki/spaces/WG/pages/181796873/2023-02-02+Map+UX we didn't make a decision on that:
> leave no-data to future decision, but counts shown will never include entities with no values for the overlay variable (no data)

What I **do** know is that the final number in the top right part of the header should be the sum of the numbers shown in the centre of the markers.  So, given that any back end changes to no-data behaviour will be several releases away, we should go with option 2 above but change the mouse-over text to something like:

[...]

PR #1656, opened 2023-02-28
1640: include "Add filters" and "Remove all" filters buttons and provide sample counts.

adnauseum:
| Before                                                                                                                                     | After                                                                                                                                      |
|--------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
| image | image |

This PR implements changes decided upon at the [02/27/2023 EDA Workspace WG meeting](https://veupathdb.atlassian.net/wiki/spaces/FG/pages/199983105/Monday+2023-02-27+1PM). This PR includes:

- Limiting filter chips to the 5 most recently selected filter chips.
   - Many have mentioned that 90%+ users have ≤ 5 filters applied to visualizations.
   - All filter chips are discoverable by click the `Show all X filters` button.
- The header now shows: 1) the total amount of samples in the study, 2) the total amount of samples in the subset, and 3) the total amount of subset samples visible on the map at the user's given viewport.

Demo:

https://user-images.githubusercontent.com/24446573/221977624-10ca667d-3aa7-4bdf-bb9e-0273c0b5ec62.mp4

### DIFF
--- a/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
@@ -177,7 +177,7 @@ function MapViz(props: VisualizationProps<Options>) {
 
   const {
     markers,
-    totalEntityCount,
+    totalEntitiesCount: totalEntityCount,
     completeCasesAllVars,
     completeCases,
     vocabulary,
--- a/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
@@ -177,7 +177,7 @@ function MapViz(props: VisualizationProps<Options>) {
 
   const {
     markers,
-    totalEntitiesCount: totalEntityCount,
+    totalEntityCount,
     completeCasesAllVars,
     completeCases,
     vocabulary,
--- a/src/lib/map/analysis/MapAnalysis.tsx
+++ b/src/lib/map/analysis/MapAnalysis.tsx
@@ -339,10 +339,6 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
 
     if (!studyEntities || !filters) return <></>;
 
-    // At the time of writing, 90%+ users have ≤ 5 filters
-    // applied to their analysis.
-    const MAX_NUMBER_OF_FILTER_CHIPS_IN_HEADER = 6;
-
     return (
       <div
         style={{
@@ -376,10 +372,7 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
         )}
         <div>
           <FilterChipList
-            filters={filters
-              .slice(0)
-              .reverse()
-              .slice(0, MAX_NUMBER_OF_FILTER_CHIPS_IN_HEADER)}
+            filters={filters}
             removeFilter={(filter) =>
               analysisState.analysis &&
               analysisState.setFilters(
--- a/src/lib/map/analysis/MapAnalysis.tsx
+++ b/src/lib/map/analysis/MapAnalysis.tsx
@@ -339,6 +339,8 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
 
     if (!studyEntities || !filters) return <></>;
 
+    const MAX_NUMBER_OF_FILTER_CHIPS_IN_HEADER = 5;
+
     return (
       <div
         style={{
@@ -372,7 +374,10 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
         )}
         <div>
           <FilterChipList
-            filters={filters.slice(0).reverse().slice(0, 5)}
+            filters={filters
+              .slice(0)
+              .reverse()
+              .slice(0, MAX_NUMBER_OF_FILTER_CHIPS_IN_HEADER)}
             removeFilter={(filter) =>
               analysisState.analysis &&
               analysisState.setFilters(
@@ -383,8 +388,8 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
             }
             variableLinkConfig={filterChipConfig}
             entities={studyEntities}
-            // selectedEntityId={studyEntities.id}
-            // selectedVariableId={selectedVariables.id}
+            selectedEntityId={subsetVariableAndEntity.entityId}
+            selectedVariableId={subsetVariableAndEntity.variableId}
           />
         </div>
       </div>
--- a/src/lib/map/analysis/MapAnalysis.tsx
+++ b/src/lib/map/analysis/MapAnalysis.tsx
@@ -23,10 +23,10 @@ import {
   FullScreenVisualization,
   NewVisualizationPickerModal,
 } from '../../core/components/visualizations/VisualizationsContainer';
-import { Close, FilledButton, FloatingButton } from '@veupathdb/coreui';
+import { Close, FilledButton, FloatingButton, Filter } from '@veupathdb/coreui';
 import { Visualization } from '../../core/types/visualization';
 import { useEntityCounts } from '../../core/hooks/entityCounts';
-import { makeStyles, Tooltip } from '@material-ui/core';
+import { Tooltip } from '@material-ui/core';
 import { Link } from 'react-router-dom';
 import { ComputationPlugin } from '../../core/components/computations/Types';
 import { ZeroConfigWithButton } from '../../core/components/computations/ZeroConfiguration';
@@ -150,7 +150,7 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
     legendItems,
     basicMarkerError,
     overlayError,
-    totalEntityCount,
+    totalEntityCount: totalEntityInSubsetCount,
     totalVisibleEntityCount,
   } = useMapMarkers({
     requireOverlay: false,
@@ -164,15 +164,6 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
     checkedLegendItems: undefined,
     //TO DO: maybe dependentAxisLogScale
   });
-
-  // Material UI CSS declarations
-  const useStyles = makeStyles((theme) => ({
-    chipListLabel: {
-      margin: '-2px 5px 5px 5px',
-      fontSize: '16px',
-      fontWeight: 500,
-    },
-  }));
 
   const finalMarkers = useMemo(() => markers || [], [markers]);
 
@@ -255,6 +246,12 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
     studyMetadata.rootEntity.id,
   ]);
 
+  const totalEntitiesInSampleCount = (() => {
+    if (!totalCounts.value || !subsetVariableAndEntity.entityId) return 0;
+
+    return totalCounts.value[subsetVariableAndEntity.entityId];
+  })();
+
   const fullScreenActions = (
     <>
       <div>
@@ -319,7 +316,6 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
   );
 
   const [mapHeaderIsExpanded, setMapHeaderIsExpanded] = useState<boolean>(true);
-  const classes = useStyles();
 
   const FilterChipListForHeader = () => {
     const filterChipConfig: VariableLinkConfig = {
@@ -331,6 +327,18 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
     };
 
     const filters = analysisState.analysis?.descriptor.subset.descriptor;
+
+    function makeButtonText() {
+      const { isSubsetPanelOpen } = appState;
+
+      if (isSubsetPanelOpen) {
+        return `Hide all filters`;
+      }
+      if (!filters) return '';
+
+      const suffix = filters.length === 1 ? '' : 's';
+      return `Show ${filters.length} filter${suffix}`;
+    }
 
     if (!studyEntities || !filters) return <></>;
 
@@ -344,12 +352,30 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
         }}
         className="FilterChips"
       >
-        <p className={classes.chipListLabel}>
-          {filters.length === 0 ? `No filters applied.` : `Filters:`}
-        </p>
+        {filters.length > 0 ? (
+          <FilledButton
+            onPress={() => {
+              setIsSubsetPanelOpen &&
+                setIsSubsetPanelOpen(!appState.isSubsetPanelOpen);
+            }}
+            text={makeButtonText()}
+            icon={Filter}
+            size="small"
+            styleOverrides={{
+              container: {
+                marginBottom: 5,
+                marginRight: 16,
+              },
+            }}
+          />
+        ) : (
+          <p style={{ padding: 0, margin: '0 0 0 5px', fontSize: 16 }}>
+            No filters applied.
+          </p>
+        )}
         <div>
           <FilterChipList
-            filters={filters}
+            filters={filters.slice(0).reverse().slice(0, 5)}
             removeFilter={(filter) =>
               analysisState.analysis &&
               analysisState.setFilters(
@@ -387,7 +413,8 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
                 onAnalysisNameEdit={analysisState.setName}
                 onToggleExpand={() => setMapHeaderIsExpanded((c) => !c)}
                 studyName={studyRecord.displayName}
-                totalEntitesCount={totalEntityCount}
+                totalEntitesInSampleCount={totalEntitiesInSampleCount}
+                totalEntitiesInSubsetCount={totalEntityInSubsetCount}
                 visibleEntitiesCount={totalVisibleEntityCount}
               />
               <MapVEuMap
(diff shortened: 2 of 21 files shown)


src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -344,12 +353,33 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
...
+            No filters applied.
+          </p>
+        )}
         <div>
           <FilterChipList
-            filters={filters}
+            filters={filters
+              .slice(0)
+              .reverse()
+              .slice(0, MAX_NUMBER_OF_FILTER_CHIPS_IN_HEADER)}
```
adnauseum, line 382: ☝️ I'm just going to let these stack

src/lib/map/analysis/SemiTransparentHeader.tsx
```diff
@@ -69,10 +71,32 @@ export function SemiTransparentHeader({
...
       <div className={semiTransparentHeader('__SampleCounter')}>
-        <p>
-          {visibleEntitiesCount || 0} of {totalEntitesCount || 0} samples
-          visible
-        </p>
+        <table>
+          <thead>
+            <tr>
+              <th title="Counts"></th>
+              <th title="Values">Samples</th>
```
adnauseum, line 78: ☝️ This shouldn't always be hardcoded. It should be the entity of the overlay variable. Could be "participants"

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -332,8 +328,21 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
...
+      const showOrHide = isSubsetPanelOpen ? 'Hide' : 'Show';
+      const suffix = filters.length === 1 ? '' : 's';
+      return `${showOrHide} ${filters.length} filter${suffix}`;
+    }
+
     if (!studyEntities || !filters) return <></>;
 
+    // At the time of writing, 90%+ users have ≤ 5 filters
+    // applied to their analysis.
+    const MAX_NUMBER_OF_FILTER_CHIPS_IN_HEADER = 6;
```
adnauseum, line 344: ☝️ We're just gonna let the chips stack at this point.

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -255,6 +246,12 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
     studyMetadata.rootEntity.id,
   ]);
 
+  const totalEntitiesInSampleCount = (() => {
+    if (!totalCounts.value || !subsetVariableAndEntity.entityId) return 0;
+
+    return totalCounts.value[subsetVariableAndEntity.entityId];
+  })();
+
```
adnauseum, line 254: Is `totalCounts.value[subsetVariableAndEntity.entityId]` have the data we're looking for?

bobular (commented):
Looking great!

This is just a UX/shifting sands specs comment.  I'll review properly in a minute.

I have some ideas about the filter functionality in the header (mainly a "remove all filters" button) but will leave those to next week's UX meeting.

Just one thing I think needs addressing now, so that we have something consistent to throw out to folk for general QA (once we have a couple more things sorted). 

When there are filters active, there's a "Show/Hide X Filters" button. When there are no active filters there's just the text "No filters applied".

I think that "Show/Hide" button ought to be there even when there are no filters. Or it shouldn't be there at all.

But I think there should always be some way to open up the filter panel from the header. So maybe there should always be an "add filters" button to the right of the chips (or the "No filters applied" text).  I'm not sure it should be a show/hide toggle, but I'll leave that to your discretion.

(We don't need the number of filters in the button any more, now that we're not truncating the list.)

There's another issue with the "Show X filters" button. We are likely going to remove the chips from the filter panel itself, so it won't really be showing X filters.  "Show filter(ing?) panel" would be OK, but I think "add filters" as I suggested above is probably all we need.

bobular:
Reviewing this and see something that might benefit from a comment (git blame says this is for @dmfalke)

The redefinition of the return type of `useAppState` to have a mandatory/non-null `appState` (previously it was optional) isn't obvious.  And links in with the `if (appStateAndSetters.appState == null) return null;` a few lines above. 

https://github.com/VEuPathDB/web-eda/blame/60fd51d2582e941f4ffc6193d0d5726c95959937/src/lib/map/analysis/MapAnalysis.tsx#L110

Also, is there a way to get rid of the `as CompleteAppState` ?  It seems like something similar to the trick you can do with chained array operations is needed here `array.filter((foo) : foo is CompleteAppState => foo.appState != null).map(...)`

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -387,7 +410,8 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
                 onAnalysisNameEdit={analysisState.setName}
                 onToggleExpand={() => setMapHeaderIsExpanded((c) => !c)}
                 studyName={studyRecord.displayName}
-                totalEntitesCount={totalEntityCount}
+                totalEntitesInSampleCount={totalEntitiesInSampleCount}
```
bobular, line 413: For consistency I would name this back to the original minus the typo, so `totalEntitiesCount` 

This is the total unfiltered count of entities that comes from `totalCounts = useEntityCounts()`

To get the correct entity we need to capture the `outputEntity` return value from `useMapMarkers`

e.g. `totalCounts.value[outputEntity.id]` (or similar)

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -387,7 +410,8 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
...
+                totalEntitiesInSubsetCount={totalEntityInSubsetCount}
                 visibleEntitiesCount={totalVisibleEntityCount}
```
bobular, line 415: We will also need to pass at least the name of the outputEntity to the header component too.

We should probably (in a new ticket) make this a drop-down select menu.

But for now, you can get the name of the output entity as follows:

Ah... I see, you pass the StudyEntity object itself (e.g. `outputEntity` returned from `useMapMarkers`) and we have `makeEntityDisplayName` function in study-metadata.ts which will format the plural for you.

src/lib/map/analysis/SemiTransparentHeader.tsx
bobular: See above :-)

bobular (commented):
I'm going to look deeper into the true meaning of the entity counts returned from `useMapMarkers` but this mini-review should be helpful to get started with. Thanks!

bobular:
Final comment!
I commented on the parent issue: https://github.com/VEuPathDB/web-eda/issues/1640#issuecomment-1453712772 

And to clarify what numbers we should go ahead and use:

1. All = `totalCounts.value[outputEntity.id]`
2. Subset = `filteredCounts.value[outputEntity.id]`
3. Visible = hmm, we need `useMapMarkers` to calculate what I described in the comment linked above.  I'll make a ticket for that.

adnauseum:
> Looking great!
> 
> This is just a UX/shifting sands specs comment. I'll review properly in a minute.
> 
> I have some ideas about the filter functionality in the header (mainly a "remove all filters" button) but will leave those to next week's UX meeting.
> 
> Just one thing I think needs addressing now, so that we have something consistent to throw out to folk for general QA (once we have a couple more things sorted).
> 
> When there are filters active, there's a "Show/Hide X Filters" button. When there are no active filters there's just the text "No filters applied".
> 
> I think that "Show/Hide" button ought to be there even when there are no filters. Or it shouldn't be there at all.
> 
> But I think there should always be some way to open up the filter panel from the header. So maybe there should always be an "add filters" button to the right of the chips (or the "No filters applied" text). I'm not sure it should be a show/hide toggle, but I'll leave that to your discretion. (Will the menu bar links secretly (or openly) be show/hide toggles? I don't have a strong opinion and would go for the simplest UX and implementation.)
> 
> (We don't need the number of filters in the button any more, now that we're not truncating the list.)
> 
> There's another issue with the "Show X filters" button. We are likely going to remove the chips from the filter panel itself, so it won't really be showing X filters. "Show filter(ing?) panel" would be OK, but I think "add filters" as I suggested above is probably all we need.

What do we think about the result of this commit: https://github.com/VEuPathDB/web-eda/pull/1656/commits/7adecc22db19f033cfb7d62e3b04f8d46cddcf59.

add_filters

bobular:
Now with added PR!  (I'll now read through your comments and watch the videos)
https://github.com/VEuPathDB/web-eda/pull/1675

bobular:
> What do we think about the result of this commit: [7adecc2](https://github.com/VEuPathDB/web-eda/commit/7adecc22db19f033cfb7d62e3b04f8d46cddcf59).

The "HIDE FILTERING PANEL" seems very verbose.

I think I prefer my suggestion ;-)

Actually I've refined it a bit just now:

* Filters: none currently applied (add)
* Filters: [chip1 x] [chip2 x] [chip3 x] (add) (remove all)

Or maybe switch the add and remove all buttons around. The remove all button can be another ticket if not completely trivial.

I personally don't think the show/hide toggling is needed. And remember that each chip also opens the filter panel (and does not act as a toggle, which is absolutely fine).

bobular:
I guess I am diverging from ClinEpi requirements...

adnauseum:
> 



> The "HIDE FILTERING PANEL" seems very verbose.
> 
> I think I prefer my suggestion ;-)
> 
> Actually I've refined it a bit just now:
> 
> * Filters: none currently applied (add)
> * Filters: [chip1 x] [chip2 x] [chip3 x] (add) (remove all)
> 
> Or maybe switch the add and remove all buttons around. The remove all button can be another ticket if not completely trivial. (These could be (+) and (x) buttons instead of text.)
> 
> I personally don't think the show/hide toggling is needed. And remember that each chip also opens the filter panel (and does not act as a toggle, which is absolutely fine).

Roger that! I think the design I implemented was to match what was in EDA. What do you think of this (I don't think we have any readily available Add/Remove icons, but I could be mistake.

https://user-images.githubusercontent.com/24446573/222828315-5e6c371a-9166-47ca-a68d-1d820983da00.mp4

src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -177,7 +177,7 @@ function MapViz(props: VisualizationProps<Options>) {
 
   const {
     markers,
-    totalEntityCount,
+    totalEntitiesCount: totalEntityCount,
```
bobular, line 180: I'm not sure we really want to start renaming things that affect other components, such as `MapVisualization`.
Sorry if I gave that impression in [this comment](https://github.com/VEuPathDB/web-eda/pull/1656#discussion_r1124553122).

I was really just referring to the property names of `SemiTransparentHeader` and I should have been clearer that the inclusion of "Sample" in those names was my main objection.

adnauseum: 👍 My bad. I'll revert those.

src/lib/map/analysis/MapAnalysis.tsx
adnauseum: This commit replaces the hardcoded "samples" with the plural display name value of `outputEntity`: https://github.com/VEuPathDB/web-eda/pull/1656/commits/3a8aabcfc36c4cec9ff99539be7fa4dfca53a0d0

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -255,6 +248,12 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
     studyMetadata.rootEntity.id,
   ]);
 
+  const totalEntitiesInSampleCount = (() => {
+    if (!totalCounts.value || !outputEntity) return 0;
+
+    return totalCounts.value[outputEntity.id];
+  })();
```
dmfalke, line 255: What is the reason for the iife here? It seems unnecessary. Also, I think this variable should be named something like `outputEntityTotalCount`. "Samples" is just one possible entity type.

dmfalke:
> Roger that! I think the design I implemented was to match what was in EDA. What do you think of this?
> 
> I don't think we have any readily available Add/Remove icons, but I could be mistaken. These can be holdovers.

I find the buttons to be confusing.

What about this?

**With no filters**
```
[Add filters] No filters applied
```

**With filters**
```
[Add filters] <filter1> <filter2> [Remove all]
```

Where "Add filters" and "Remove all" are buttons. I would also make "Remove all" a smaller button.

dmfalke (commented):
Overall, I think the code looks good.

Not totally related to this PR: we really need to decide if we're going to go all-in with emotion for styling, or if we want to consider something else. I know I'm the one who started it here, but inline styles are not ideal. @adnauseum I know you've had thoughts on this in the past.

adnauseum:
> Overall, I think the code looks good.
> 
> Not totally related to this PR: we really need to decide if we're going to go all-in with emotion for styling, or if we want to consider something else. I know I'm the one who started it here, but inline styles are not ideal. @adnauseum I know you've had thoughts on this in the past.

Oh yeah inline styles are my jam! 

No, for real, I think we need to huddle on this. I do feel pretty strongly about Tailwind because it's good for shipping less css, providing a solution for building out a design system and one-off components alike. 

But yeah, that'll span beyond this PR, as you've alluded to. And I was thinking maybe once the monorepo stuff is completed?

dmfalke:
> No, for real, I think we need to huddle on this. I do feel pretty strongly about Tailwind because it's easy, an industry standard, ships less CSS, and provides a good solution for building out both a design system and one-off components alike. It's also very technology agnostic and future proof (it's just CSS). One of the authors of Glamorous (the precursor to Emotion) is also pretty bullish on Tailwind and bearish on Emotion.
> 
> But yeah, that'll span beyond this PR, as you've alluded to. And I was thinking maybe once the monorepo stuff is completed?

That's what I'm thinking, regarding the monorepo stuff.

Something to keep in mind is that the CoreUI repo has implemented a sophisticated theming solution that's built on emotion. One of the features is that you can define override rules for targeted parts of a component, as a part of the component's props interface. This makes it easy to constrain how a components styling can be altered, which helps with consistency in disparate parts of our websites. It would be great if we can keep that.

While I love plain old CSS, there isn't really a way to expose only parts of the style definitions for customization. In my experience, this can lead to complicated overrides that dramatically alter the component's style, and eventual uses of `!important`.

More generally, I typically look to "affordances" as a way to encourage certain approaches and discourage abuses. It's part of what I like about TypeScript and component based user interfaces. I've struggled to find a way to do this for styling. I'm not seeing emotion having this effect, but it may be due to lack of discipline. Plain old CSS definitely doesn't have this feature, either, given it's "expose everything" trait.

What I'm getting at is, whatever we decide, we need to think about it in terms of both end user performance, and making it easy for developers to use our solution. I'm looking forward to discussing this in more detail.

dmfalke:
Sorry to add more noise to this PR...

I think we should survey the current use of `styleOverrides`. My gut is telling me that we shouldn't need them.

adnauseum:
> Sorry to add more noise to this PR...
> 
> I think we should survey the current use of `styleOverrides`. My gut is telling me that we shouldn't need them.

Yeah, I think we'll have lots of discussions about what will work best for us!

adnauseum:
> What about this?
> 
> **With no filters**
> 
> ```
> [Add filters] No filters applied
> ```
> 
> **With filters**
> 
> ```
> [Add filters] <filter1> <filter2> [Remove all]
> ```


https://user-images.githubusercontent.com/24446573/223179217-f9c0a276-2c9f-4889-aa6c-044d3a53674e.mov

dmfalke:
The latest screencast looks great. What about using `size="small"` or `FloatingButton` for the "Remove all" button?

adnauseum:
> The latest screencast looks great. What about using `size="small"` or `FloatingButton` for the "Remove all" button?

It's presently `size="small"`. I didn't know about the `<FloatingButton />`! That does look better.

image

dmfalke:
> > The latest screencast looks great. What about using `size="small"` or `FloatingButton` for the "Remove all" button?
> 
> It's presently `size="small"`. I didn't know about the `<FloatingButton />`! That does look better.
> 
> image

Ah! I didn't notice they were both small :smile:. Sorry about that!

adnauseum:
@bobular & @dmfalke 

Are will we content with this? I'm sure we'll tweak things after we collect feedback from a larger audience.

final_demo

src/lib/map/analysis/MapAnalysis.tsx
adnauseum: Dunno, it looked good to me on Friday afternoon. I agree with you though. I changed it to in this commit: https://github.com/VEuPathDB/web-eda/pull/1656/commits/496ae2e57394b4ecf324ae772f939b0e7926b7c9.

```js
  const outputEntityTotalCount = totalCounts.value && outputEntity ? totalCounts.value[outputEntity.id] : 0;
```

dmfalke (approved):
This looks great to me!

bobular:
Just pulled main - this looks excellent Sam.
The not-too-in-your-face 'remove all' button is great.
I like everything else too.

PR #1667, opened 2023-03-01
Bug - 2x2 table setup

### LINKED ISSUE #1663: 2x2 Mosaic plot Table is not being set up correctly
Please see example below from GEMS1 Case Control. Quadrant A (top left) values as selected by the user should have been "Case" "Yes" but in the table it is "Case" "No". 

Screenshot 2023-02-28 at 12 59 28 PM

jernestmyers:
Closes #1663 

Depends on https://github.com/VEuPathDB/web-components/pull/458


src/lib/core/components/visualizations/implementations/MosaicVisualization.tsx
```diff
@@ -530,6 +530,10 @@ function MosaicViz(props: Props<Options>) {
                   facetVariable ? facetVariable.displayName : 'Facet'
                 }
                 enableSpinner={data.pending}
+                selectedReferenceValues={[
+                  xAxisReferenceValue,
+                  yAxisReferenceValue,
+                ]}
```
dmfalke, line 536: This could use a `useMemo` to prevent extra work in the `ContingencyTable` component:

```typescript
const selectedReferenceValues = useMemo(() => [
  xAxisReferenceValue,
  yAxisReferenceValue
], [xAxisReferenceValue, yAxisReferenceValue]);
```

dmfalke (changes requested):
See suggestion... otherwise, looks good!

dmfalke (approved):

PR #1668, opened 2023-03-01
resolve scroll issue of entity diagram

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1567.

Here are example screenshots based on the example in the original ticket.

a) sufficient horizontal width of one's browser: entity diagram is still centered as designed
entitydiagram01

b) intentionally narrowed down the horizontal width of one's browser: the leftmost entity, Clusters, is now shown without cut-off
entitydiagram02

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "4.0.18",
+  "version": "4.0.19",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


dmfalke (approved):

moontrip:
Thank you for your quick review, @dmfalke 👍

PR #1670, opened 2023-03-02
Add `disabled` prop to `VariableLink`

dmfalke:
This PR adds a new `disabled` prop to `VariableLink`

### DIFF
--- a/src/lib/core/components/VariableLink.tsx
+++ b/src/lib/core/components/VariableLink.tsx
@@ -1,8 +1,12 @@
 /*
  * Link to the page for a variable
  */
-import { forwardRef, Ref } from 'react';
+import { CSSProperties, forwardRef, Ref } from 'react';
 import { Link, LinkProps } from 'react-router-dom';
+
+const disabledStyle: React.CSSProperties = {
+  pointerEvents: 'none',
+};
 
 type VariableValue = {
   entityId?: string;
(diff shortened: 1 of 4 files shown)


src/lib/core/components/VariableLink.tsx
```diff
@@ -1,9 +1,13 @@
...
  * Link to the page for a variable
  */
-import { forwardRef, Ref } from 'react';
+import { CSSProperties, forwardRef, Ref } from 'react';
 import { Link, LinkProps } from 'react-router-dom';
 
+const disabledStyle: React.CSSProperties = {
+  pointerEvents: 'none',
+};
+
```
jernestmyers, line 10: This can be removed since A) `pointerEvents` would disallow the cursor styles and B) it's being overridden by the component-scoped `disabledStyle` anyway.

jernestmyers (approved):
I'll approve since everything works as expected, but I'll also point out that current sites have an underline onHover effect (see screenshot) that no longer seems to be applied. If that's intentional, okay. If not, consider adding it back in before merging.

image

dmfalke:
> I'll approve since everything works as expected, but I'll also point out that current sites have an underline onHover effect (see screenshot) that no longer seems to be applied. If that's intentional, okay. If not, consider adding it back in before merging.
> 
> image

Ah, good catch. I don't think the underline was intentional, and neither was removing it. It's just a function of the elements being used. I can add it back, just for the tree. I suspect users like the feedback... and it's less change.

src/lib/core/components/VariableLink.tsx
dmfalke: Oops, thanks!

dmfalke:
I've addressed the feedback

jernestmyers (approved):
Looks good. And good call on the `aria-disabled` addition, plus on doing the opposite of what I said by removing the component-scoped CSS object and amending the module-scoped object! 😅

PR #1671, opened 2023-03-02
Add lefthand side menu to standalone map

adnauseum:
| Before                                                                                                | After                                                                                                                      |
|-------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|
| Before | flex_layout_demo|

This PR's a minimal first iteration at the lefthand menu. In future PRs we should address points such as:

- When I've clicked on "Filter Data," do we want an active state for the side navigation item while the draggable panel is open?
- Do the navigation items work as toggles for the draggable panels?

Anyway, this PR gets the core functionality of what we want: a header and a lefthand menu that will not collide with each other.

<details>
<summary>
More words, pictures, and tomfoolery.
</summary>

<br />

Change list:
- [Refactor the layout to use flexbox.](#f)
- [Add the lefthand side navigation.](#lefthand)

<h2 id="f">Flexbox</h2>

Before this PR the header was a `<header>` tag positioned absolutely. This removes the element from the normal flow of DOM elements and sticks it onto the top of the nearest parent element set to `position: relative`. Because the `<header>` is removed from the flow of the DOM, it loses all knowledge of what elements surround it. This means that it'll happily collide with the lefthand menu. And if the lefthand menu is positioned absolutely, it too will happily collide with anything.

This PR renders both the lefthand menu and the header as children of a flexbox parent. This means the flex children know each other and we don't have to worry about them overlapping! Check out this interactive demo: https://codepen.io/skendrick366/pen/yLxMbpj?editors=1100. You can resize the textbox up and down to grow the height of the header and you'll see how the lefthand menu knows this and ducks its head.

https://user-images.githubusercontent.com/24446573/222560800-4de9d408-bcf8-4a62-8a71-91094dbca9fe.mp4

<h2 id="lefthand">Lefthand side navigation</h2>

Not much to report here. This is a simple component. The only state it cares about (for now) is whether or not it's open. If we wish to persist that in a DB,  we'll have to add a prop like `onMenuToggle`. No sweat.
</details>


src/index.tsx
```diff
@@ -189,10 +189,11 @@ initialize({
       component: () => (
         <MapApp
           logoProps={{
-            href: 'https://veupathdb.org',
-            src:
+            siteHomeUrl: 'https://veupathdb.org',
+            siteLogoSrc:
               'https://veupathdb.org/veupathdb/images/VEuPathDB/icons-footer/vectorbase.png',
             siteName: 'VectorBase',
+            loginUrl: 'https://eupathdb.org/oauth/assets/eupathdb-login.html',
```
adnauseum, line 196: FYI: These props are used to render the logo in the map header as well as the site URL and login URL in the side menu.

src/lib/map/MapVeuContainer.tsx
```diff
@@ -20,12 +20,12 @@ import {
 } from '../core/hooks/client';
 
 import './MapVEu.scss';
-import { SemiTransparentHeaderLogoProps } from './analysis/SemiTransparentHeader';
+import { SiteProps } from './analysis/MapNavigation';
```
adnauseum, line 23: I renamed these props because they're used for both the header and the lefthand side menu. These props do the following: https://github.com/VEuPathDB/web-eda/pull/1671/files?diff=split&w=1#r1123744372

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -13,53 +22,46 @@ import {
...
-  SemiTransparentHeaderLogoProps,
-  SemiTransparentHeader,
-} from './SemiTransparentHeader';
-import FilterChipList from '../../core/components/FilterChipList';
-import { VariableLinkConfig } from '../../core/components/VariableLink';
+import { AppState, useAppState } from './appState';
+import { FloatingDiv } from './FloatingDiv';
+import { MapLegend } from './MapLegend';
+import { SemiTransparentHeader, SiteProps } from './MapNavigation';
+import { MapSideNavigation } from './MapSideNavigation';
```
adnauseum, line 64: My apologies... I saw an `Optimize imports` command in VS Code and forgot to undo it after see what it'd do.

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -385,6 +391,72 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
...
+  const iconStyles: React.CSSProperties = {
+    height: 25,
+    width: 25,
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+  };
+  const labelStyles: React.CSSProperties = {
+    marginLeft: '0.5rem',
+  };
```
adnauseum, line 419: Keeping these as CSS-in-JS because we've got to align on what we're doing. Moreover, these styles apply to navigation items and I think we need to talk about our expectations for those.

For example, when I've clicked on "Filter Data," is there an active state for the side navigation item while the draggable panel is open?

And the design has a big site name header at the top of it—do we really want that?

This PR a minimal first iteration at the side menu.

src/lib/map/analysis/MapSideNavigation.tsx
```diff
@@ -0,0 +1,159 @@
...
+      </button>
+      <div
+        style={{
+          // The navigation items are put into a container
+          // which takes up 100% of available space and
+          // allows users to scroll through navigation whenever
+          // there are lots of navigation items and/or the user's
+          // viewport is short.
+          height: '100%',
+          overflowY: 'scroll',
```
jernestmyers, line 98: Is there a reason we want 'scroll' instead of 'auto'? FireFox renders a disabled scrollbar, which looks a bit strange in my opinion.
image

src/lib/map/analysis/MapSideNavigation.tsx
```diff
@@ -0,0 +1,159 @@
...
+        // including its width, height, and location on the
+        // page (which depends on the `isExpanded` piece of state.
+        background: menuBackground,
+        // height: 'calc(100% - 150px)',
+        height: '100%',
+        left: isExpanded ? 0 : -200,
+        minHeight: 125,
+        position: 'relative',
+        transition: 'left 0.1s ease',
+        width: 200,
```
jernestmyers, line 42: For maintainability, I think we should connect left's unexpanded offset to the container's width. Something could similarly be done to the button's width and its right offset. Perhaps some module-scoped vars that specify the widths, and the offsets can invert those values?

dmfalke (commented):
I've looked through the code, and there is nothing glaring to me. I think @jernestmyers asked some good questions!

I'll demo the code tonight or tomorrow.

src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -396,21 +468,76 @@ export function MapAnalysisImpl(props: Props & CompleteAppState) {
...
+                  height: '100%',
+                  width: '100%',
+                  // Attach this div container to it's parent.
+                  position: 'absolute',
+                  // Remember that just about everything in the DOM is box.
+                  // This div is sitting on top of the map. By disabling
+                  // pointer events we are saying: hey, div, become porous.
+                  // If a user clicks you, don't capture it, but let it go
+                  // to the map you're covering.
+                  pointerEvents: 'none',
```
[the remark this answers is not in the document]
adnauseum: image

By default everything in the DOM is set to `pointer-events: auto`. As you can seem, my layout is superimposed atop the map. That means it'll catch all the clicks and our poor map won't have any. If I set the parent container to `pointer-events: none` then I'm telling it to be a ghost. You can see it, but you can't touch it.

This is a problem because I _do_ want clicks on the header and side navigation. So I set `pointer-events: auto` onto the children to restore that property _only for those elements_. The parent is still `pointer-events: none` 👻

src/lib/map/analysis/MapSideNavigation.tsx
adnauseum: Oh weird! My FF isn't rendering that way! 😬

image

Anyways, this is a great callout. I was getting carried away with thinking about having a ton of navigation items. But I think we'll cross that bridge if/when we get there. 

I removed these in https://github.com/VEuPathDB/web-eda/pull/1671/commits/2ffb0860d3a1dd6ffd8cd0f42e17146b91de6dbd

src/lib/map/analysis/MapSideNavigation.tsx
adnauseum: Added in https://github.com/VEuPathDB/web-eda/pull/1671/commits/01f25de8d9f1177de8eb89ab0c50345f628330f2

jernestmyers: Yup, although I imagined them outside the component's scope!

adnauseum: Gotcha! What's the use case you have in mind?

jernestmyers: No particular use case, I just imagine that we don't need to initialize those variables each time the component renders. Not sure if my reasoning is sound at all. And even it were sound, I know it makes no difference performance-wise at such a small scale. Just thinking out loud in a way!

adnauseum: Makes sense! I think for now the variables make the CSS a bit easier to grokk, so I'm cool with them. I think we'll pick up this thread whenever we decide on what we're doing for CSS!

bobular: When you resolve this, I would appreciate an explanation!

bobular: (in a dev meeting follow-up topic or similar - don't spend hours typing)

adnauseum:
> Clicking again to close the panel is fine. Would unsaved changes, e.g. in the paint-by-variable panel discussed yesterday, always be lost? If we actually kill the component (as opposed to making it `display: none` or whatever) then this is easy to implement and avoids having some potentially large components rendered but not visible (e.g. the download panel).

Thanks for the PR review! 🙌

- I totally forgot to extract `SiteInformationProps` to a higher directory. Thank you! I did this in this commit: https://github.com/VEuPathDB/web-eda/pull/1671/commits/88d1ea46f33596d70b8a7a5f7d701b10b7dd0f9d
- I think it’ll make sense for the lefthand panel to toggle open/closed states of the associated draggable panel, but we’ll find out with we click around with it.
- Regarding the closing of panels, right now I think the implementation is that the component is not destroyed, but hidden via `visibility: hidden`. Given this, I think unsaved changes would just stay there if the user mistakenly closes a panel (a common situation, I'm sure).

bobular:
Thanks @adnauseum 

I re-approve the type definition move :-)

Thanks for your explanation/thoughts on open/close toggles and state - sounds great. Let's see how it behaves in practice.

bobular:
We could eventually add a hook (in the non-React sense) to the panel hiding to say "Apply unsaved changes?"

src/lib/map/analysis/MapSideNavigation.tsx
dmfalke: @jernestmyers I bet you have an external mouse

dmfalke (approved):
Sorry for the late approval! I did demo this and then immediately went on vacation for a couple of days. Looks great!

src/lib/map/analysis/MapSideNavigation.tsx
jernestmyers: I do often use an external mouse... 🤔

dmfalke: If you just have a trackpad, Mac's will render an overlay scrollbar, but if you have an external mouse, it will render it inline. You can change that behavior, but this is the default.

Issue #1664, opened 2023-03-01
GeoMap icon in Full screen map says 'coming soon'


SAK22:
So when in full screen map if you click new visualization, the geo map icon says coming soon still. 
@moontrip suggests perhaps Geo map icon should not be shown at the Select Viz in the full screen map? Or leaving the icon but saying like Not Available.

Sorry if this ticket is in the wrong place!

moontrip:
@bobular @dmfalke Outreach team reported that when opening Visualization (New Viz button at top right) at FSM, Map related icons (Vizs) such as Geolocation map and Map markers are listed as coming soon: e.g., https://github.com/VEuPathDB/web-eda/issues/881#issuecomment-1448367168. Under the FSM, I think that it is better not to show those two icons: it does not mean much to make another mapViz inside FSM mapViz. What do you think?

bobular:
Good spot!

The FSM will be retired soon, but the same problem exists in the prototype SAM when you add a floating plot:

![image](https://user-images.githubusercontent.com/308639/222717221-4f5c178c-5f39-49c3-9e33-afe3859be809.png)

I think this will get solved in due course by back end work. See https://github.com/VEuPathDB/web-eda/pull/1672#pullrequestreview-1323706521

PR #1672, opened 2023-03-02
hiding map-related Viz icons under FSM

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1664.
What is done is to add a prop to let the child component (NewVisualizationPicker) know whether the parent component is FSM or not. If it is from FSM, then map-related Viz icons is hidden in the Select Visualization.

Here is a screenshot under FSM, which two map-related Viz icons such as map-markers and map-markers-overlay are not shown. Certainly, in a normal mode, the New Visualization shows all Viz icons.

fsm-geomap-icons

### DIFF
--- a/src/lib/core/components/fullScreenApps/fullScreenMap.tsx
+++ b/src/lib/core/components/fullScreenApps/fullScreenMap.tsx
@@ -369,6 +369,8 @@ function FullScreenMap(props: FullScreenComponentProps) {
               visualizationsOverview={app.visualizations}
               geoConfigs={[geoConfig]}
               onVisualizationCreated={onVisualizationCreated}
+              // declare that this is from FSM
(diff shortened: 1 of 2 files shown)


bobular (commented):
Thanks for looking at this. However, I think this will be resolved when the back end has a new "pass through" app specifically for the Standalone Map.  That app will only return a list of feasible visualisations.  See https://github.com/VEuPathDB/EdaDataService/issues/237 

I don't think we need to make any fixes for the FSM as this will be retired.

So I would recommend closing this as "won't fix" - sorry @moontrip

moontrip:
@bobular Thank you for your information. I didn't know the ticket you mentioned. Yes if this is not urgent, then we can wait for the backend change. I will close this PR as won't fix.

PR #1673, opened 2023-03-03
Add warning banner for download table preview for entities with lots of vars

### LINKED ISSUE #1652: Bug: show user-friendly message when subset results table preview cannot be generated

asizemore:
Resolves #1652 

Current behavior
Screen Shot 2023-03-03 at 10 58 13 AM

Now it looks like:
Screen Shot 2023-03-03 at 10 58 38 AM

That error banner only comes up once i've clicked at least one var.

Is it weird to have both banners? I thought about removing the server-error banner, but i dont want to hide errors that might be useful to us to see later...

### DIFF
--- a/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
+++ b/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
@@ -284,6 +284,7 @@ export default function SubsetDownloadModal({
       analysisState.analysis?.descriptor.subset.descriptor,
       currentEntity,
       mergeKeys,
+      canLoadTablePreview,
     ]
   );
 
--- a/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
+++ b/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
@@ -417,7 +417,7 @@ export default function SubsetDownloadModal({
             banner={{
               type: 'warning',
               message:
-                'A table preview is not available for this entity. You can still select variables and download a tabular file.',
+                'Preview is not available. You can still configure and download a tabular file of the filtered dataset.',
             }}
           />
         )}
--- a/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
+++ b/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
@@ -365,6 +365,8 @@ export default function SubsetDownloadModal({
 
   const headerMarginBottom = 15;
 
+  const showPreviewTableWarning = currentEntity.variables.length >= 1000;
+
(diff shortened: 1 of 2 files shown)


dmfalke:
I think we want to avoid making the request if `showPreviewTableWarning === true`.

I would rename that variable to something like `canLoadTablePreview` and define it closer to the top of the component, and use that value as a part of [this condition](https://github.com/VEuPathDB/web-eda/blob/ad8a125f44a7fa6e199737d2deb7a996913c8171/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx#L225).

dmfalke:
> I would rename that variable to something like `canLoadTablePreview` and define it closer to the top of the component, and use that value as a part of [this condition](https://github.com/VEuPathDB/web-eda/blob/ad8a125f44a7fa6e199737d2deb7a996913c8171/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx#L225).

Hmm... maybe [this](https://github.com/VEuPathDB/web-eda/blob/ad8a125f44a7fa6e199737d2deb7a996913c8171/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx#L226) would be a better place to add the check.

asizemore:
Thanks @dmfalke !
Now after the update it looks like:

Screen Shot 2023-03-06 at 2 01 55 PM

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
```diff
@@ -223,7 +229,7 @@ export default function SubsetDownloadModal({
   >['serverSidePagination']['fetchPaginatedData'] = useCallback(
     ({ pageSize, pageIndex, sortBy }) => {
       if (!currentEntity) return;
-      if (selectedVariableDescriptors.length === 0) {
+      if (selectedVariableDescriptors.length === 0 || !canLoadTablePreview) {
```
dmfalke, line 232: `canLoadTablePreview` should probably be added to the list of dependencies for this callback function (https://github.com/VEuPathDB/web-eda/blob/f606310611cbeda0bfa01d30b441e8bcb981053e/src/lib/workspace/Subsetting/SubsetDownloadModal.tsx#L280-L287)

dmfalke (approved):
Once my comment is addressed, I think this is ready to go!

src/lib/workspace/Subsetting/SubsetDownloadModal.tsx
asizemore: Right, because if this value changed we'd definitely want to stop/start fetching data. Thank you!

PR #1675, opened 2023-03-03
Issue 1674 add `totalVisibleWithOverlayEntityCount` return value to `useMapMarkers`

### LINKED ISSUE #1674: `useMapMarkers` should also return number of entities in viewport with data for overlay variable (and existing bug)
However, it's not that simple. It's never that simple.

Trivially the client could sum up the values in `mapMarkers.data[].value[]`
(from `map-markers-overlay` endpoint)

However when the markers are in proportion mode, those values are no longer counts, and the sum of them is no use.

A workaround in the current code is to [sum up the relevant counts](https://github.com/VEuPathDB/web-eda/blob/main/src/lib/core/hooks/mapMarkers.tsx#L393-L398) in the `sampleSizeTable`

However, this is incorrect.  The `sampleSizeTable` is calculated before the viewport is applied. So if any geohash rectangles are truncated at the edge of the viewport, the entity count _on screen_ could be less than the entity count in `sampleSizeTable`.

So, with reference to https://github.com/VEuPathDB/web-eda/issues/1640#issuecomment-1453712772 possible solutions could be
1. disable proportion mode (for now, at least) and sum up the `mapMarkers.data[].value[]` as above
2. do some back end work to include the entity counts alongside the proportions (but is it worth doing this before the back-end rework starts, @d-callan ?)
3. don't even attempt to calculate the entity count described in option 2 of [this comment](https://github.com/VEuPathDB/web-eda/issues/1640#issuecomment-1453712772) - but instead report the number of entities that would be on the screen if we turned off the overlay variable (or if it had data for 100% entities) - which is option 1 in that comment.  Without displaying no-data counts, I don't personally like this option.

bobular:
Fixes #1674 

Also tidied up some comments and merged some reduce functions.

I haven't tested the new `totalVisibleWithOverlayEntityCount` value. Please could you do that as part of your PR review @adnauseum? It should always be the same as the sum of the marker-center numbers in the viewport.

### DIFF
--- a/src/lib/core/hooks/mapMarkers.tsx
+++ b/src/lib/core/hooks/mapMarkers.tsx
@@ -390,6 +390,7 @@ export function useMapMarkers(props: MapMarkersProps): MapMarkers {
               map[geoAggKey] = {
                 // sum up the entity count from the sampleSizeTable because
                 // the data.label values might be proportions (and sum to 1)
+                // SEE 'WARNING' BELOW ABOUT THIS BEING INCORRECT
                 entityCount: sum(
                   overlayResponse.value.sampleSizeTable.find(
                     (item) =>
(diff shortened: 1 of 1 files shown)


adnauseum (approved):
LGTM

https://user-images.githubusercontent.com/24446573/222821365-bd310afd-0fbb-4d41-bf40-ffd63e6343bb.mp4

bobular:
Cool - best to check it with a variable that isn't available for all entities.  "Hookworm, by microscopy" is a good one.

Had no idea about `console.table()` !!

bobular:
I've merged it anyway. Thanks for testing!

Issue #1150, opened 2022-05-24
Update data accessibility info in Study details & Download tabs, specifying that access requests are pending


danicahelb:
Related to: https://github.com/VEuPathDB/web-eda/issues/1172 and https://github.com/VEuPathDB/EbrcWebsiteCommon/issues/143

Discussed in 5/20/2022 EDA UX meeting. We will add responsive messaging on the download and study details tabs, immediately below the tabs (above all download buttons and links).

We are currently differentiating (1) users who were not logged in / did not have access from (2) users who had access. 

In this ticket, we want to differentiate a third group: (3) users who submitted a request FOR A PROTECTED STUDY but the request is still pending.

For users whose access requests are pending (relevant to protected studies only):
> Your data access request is pending.

_Notes:_ 
_- For public studies, anyone can download the data so we don't need verbiage pending requests._
_- For controlled studies, access requests are approved immediately upon submission and do not need verbiage for pending requests._

aurreco-uga:
in fut (steve, dave and i) on Jan 5 2023, we propose to 
(1) when the user is not approved to access a study, simplify the current verbiage  as:
"To download data please request access. Data will be available upon study team review and approval."
(2) when clicking on the "request access" link, we go to the request form straight (do not show popup)
(3) use the current endpoint permissions to find out if a user has a pending request to access a study:
        ( **api change to be agreed upon**)
       /eda/permissions/
       {  actionAuthorization: {.....}
           **request_pending: true**
           isManager = false
           isUserStudy....
             ....
        }
(4) when the user has already submitted a request and the request is pending, use a new endpoint to grab the request content and show the form filled in (cant be edited): 
      (**endpoint to be agreed upon**)
       /eda/**dataset-request-form**?datasetId=DS_fe95d702c3

aurreco-uga:
In addition to the above Danica notices:
- for prerelease studies we need to get rid of the  sentence (there is a banner already)

danicahelb:
Just to clarify (1): Update the verbiage as follows:

Data Accessibility: Protected
To download data please request access. Data will be available upon study team review and approval.

Data Accessibility: Controlled
To download data please request access. Data will be available upon study team review and approval.

Data Accessibility: Public
Data downloads for this study are public. Data is available without logging in.

Data Accesibility: Prerelease
Data downloads for this study are not yet available on ClinEpiDB.

danicahelb:
Would the current plan mean this ticket is obsolete?

[...]

PR #1676, opened 2023-03-03
Update data accessbility language and add loading text

chowington:
Works toward https://github.com/VEuPathDB/web-eda/issues/1150

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "4.1.1",
+  "version": "4.1.2",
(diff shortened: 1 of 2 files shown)


dmfalke (approved):

chowington:
Adding the loading text is slight scope creep, but it prevents (1) ugly popping-in, and (2) the user not knowing that their permissions aren't loaded yet, and so can't download anything yet

chowington:
Actually, this fixes https://github.com/VEuPathDB/web-eda/issues/1150

PR #1680, opened 2023-03-06
Replace location when compute config changes

dmfalke:
This will prevent "backing" into a non-existent compute config.


bobular:
Nice and simple!

PR #1682, opened 2023-03-06
fixing mosaic tooltip for no facet variable

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1681. 

As Dave thought, I was looking into web-components' components, however, I just found that plot interaction was simply not to be activated (didn't pass the prop when no facet variable was selected). I confirmed that this would work for both Mosaic 2x2 and RxC, including with or without facet variable.

Here is a screenshot:
mosaic-no-facet-tooltip1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "4.1.4",
+  "version": "4.1.5",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


dmfalke (approved):

moontrip:
@dmfalke Thank you for your super quick review 👍

Issue #1082, opened 2022-04-13
Line plot tool - update UI for 0/0 proportions


danicahelb:
We need to revisit the UI for line plots with proportions where we have 0/0.

This is a screenshot from the LLINEUP study. The design of this study was such that only kids 2-10 years were tested for malaria.

The issue is that -- because there are so many points between 11yrs - 110yrs, the null points along the x-axis actually look to be a connected line.

<img width="1402" alt="image" src="https://user-images.githubusercontent.com/25450900/163234962-245d9cb3-ec64-438c-8f48-a5440da1ed4e.png">

danicahelb:
Related to https://github.com/VEuPathDB/web-eda/issues/1411

Discussed in 12/2/2022 EDA UX meeting. Markers at 0 when y-axis proportions are 0/0 are confusing to users. This was brought up by participants during the TMRC workshop in October

For continuous variables, we do NOT show markers at 0 when there is no data

The birds eye view & variable completeness table exist so that missing data is accounted for

**Decision**: do not add markers at 0 when the proportion is 0/0

danicahelb:
Related topic: https://github.com/VEuPathDB/EdaNewIssues/issues/503

danicahelb:
Discussed UI updates to clarify instances of 0/0 in the line plot in the 12/12/2022 EDA UX meeting

**0/0 markers should be hollow instead of filled & will remain at y=0** 
- if the user selects an overlay variable, the hollow markers will be colored according to the legend
- We acknowledge that there may be overplotting of hollow 0/0 markers at y=0 when an overlay variable has been selected

**Mouseover information for the 0/0 markers will be updated** 
- to address overplotting when an overlay has been selected, the mouseover should have information for ALL overlay values at that x-axis position, each in boxes colored according to their legend colors
- change `y: 0` to `y: undefined`
- change `n: 0/0` to `n: count of values removed from denominator`

**Add information to the legend indicating what the hollow markers mean**

**When the user selects “Auto-zoom” for the x-axis, the 0/0 markers should NOT influence the auto-zoomed x-axis range. **

We will not update the birds eye view or variable completeness table when a user removes a value from both the numerator and denominator of the proportion selector
- Technically the user is creating a new “subset” within the line plot
- However, t birds eye view or variable completeness table refer to the ANALYSIS SUBSET created in the B&S tab and the user is not removing values from the subset in the analysis

@d-callan will mockup these updates for sign-off in an EDA UX meeting

[...]

PR #1686, opened 2023-03-09
Make line plot 0/0 proportions have hollow circle

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1082

### DIFF
--- a/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -48,6 +48,8 @@ import {
   mapValues,
   map,
   keys,
+  get,
+  set,
 } from 'lodash';
 import BinWidthControl from '@veupathdb/components/lib/components/plotControls/BinWidthControl';
 import LabelledGroup from '@veupathdb/components/lib/components/widgets/LabelledGroup';
@@ -665,7 +667,8 @@ function LineplotViz(props: VisualizationProps<Options>) {
         showMissingFacet,
         facetVocabulary,
         facetVariable,
-        neutralPaletteProps.colorPalette
+        neutralPaletteProps.colorPalette,
+        params.config.valueSpec === 'proportion'
       );
     }, [
       studyId,
@@ -1835,7 +1838,8 @@ export function lineplotResponseToData(
   showMissingFacet: boolean = false,
   facetVocabulary: string[] = [],
   facetVariable?: Variable,
-  colorPaletteOverride?: string[]
+  colorPaletteOverride?: string[],
+  dependentIsProportion?: boolean
 ): LinePlotDataWithCoverage {
   const modeValue: LinePlotDataSeries['mode'] = 'lines+markers';
 
@@ -1880,7 +1884,8 @@ export function lineplotResponseToData(
       response.lineplot.config.binSpec,
       response.lineplot.config.binSlider,
       overlayVariable,
-      colorPaletteOverride
+      colorPaletteOverride,
+      dependentIsProportion
     );
 
     return {
@@ -1953,6 +1958,8 @@ function nullZeroHack(
   dataSetProcess: LinePlotDataSeries[],
   dependentValueType: string
 ): LinePlotDataSeries[] {
+  console.log('In nullZeroHack()');
+  console.log({ dataSetProcess });
   // make no attempt to process date values
   if (dependentValueType === 'date') return dataSetProcess;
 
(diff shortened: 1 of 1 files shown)


bobular (commented):
I'm not sure it's worth trying to get rid of the tiny "london underground" horizontal lines at this point.

Approve of everything so far. But merging this would not fix everything in the rather large ticket, so I'm not approving. (I see it's still in Draft though)

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2126,8 +2138,12 @@ function processInputData(
   binSpec?: BinSpec,
   binWidthSlider?: BinWidthSlider,
   overlayVariable?: Variable,
-  colorPaletteOverride?: string[]
+  colorPaletteOverride?: string[],
+  dependentIsProportion?: boolean
```
bobular, line 2142: An alternative to passing yet another argument might be to create at io-ts type for the numeratorN/denominatorN structure, so:

```
      binSampleSize: union([
        array(
          type({
            N: number,
          })
        ),
        array(
          type({
            numeratorN: number,
            denominatorN: number,
          })
        ),
      ]),
```
would become
```
export type ProportionSampleSizes = TypeOf<typeof ProportionSampleSizes>;
const ProportionSampleSizes = type({
            numeratorN: number,
            denominatorN: number,
          });

...

      binSampleSize: union([
        array(
          type({
            N: number,
          })
        ),
        array(ProportionSampleSizes),
      ]),
```

Then in the `markerSymbol` function you could replace the `dependentIsProportion` with something like

```
el.binSampleSize.every((bss) => ProportionSampleSizes.is(bss))
```

I don't know if that makes it any more (or less) readable... (and it probably doesn't fix all the TS issues)

It would be nice to avoid the `any` and the `hasOwnProperty` check on just the first array element. But I may just be introducing more TS spaghetti to compensate...

bobular (approved):
If we choose to merge (see parent ticket discussion), I approve this. Just one style/TS comment/suggestion.

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
chowington: I might wrap back around to the `numeratorN` stuff, though there are more pressing TS issues, like the many `any`s, as you mention.

chowington:
> Had a quick look at the "splitting" code. Was wondering why the nullZeroHack function is still used. But as I say, it was a very quick look. Are you ready for a more detailed look at the new code?

I plan on removing my added code in `nullZeroHack` that handles the `marker.symbol` array, but the original version of the function should stay, right?

A detailed look might be helpful if you have any ideas on how to fix the nasty TS errors that I've been sidestepping with these `any`s! I spent time trying to handle it more robustly, but it's a bit of a monster, at least to me.

Also, there's the one property `facetVariableDetails` that's giving me a headache because it's a constant-size tuple (a TS concept, although it's just an array in JS), meaning it shouldn't be iterated over as an array, but TS/JS doesn't seem to give any method for weeding out tuple from general arrays. So I'm just excluding it manually, which is annoying. Any thoughts?

src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
bobular: I think we might need to explicitly "tag" the LineplotResponse series with an "isZeroSeries" property?

Introspecting to figure out which symbol to use is tricky. I'm noticing a `el.binSampleSize[0] is undefined` error when the zeroSeries is empty, though that may be avoidable with a length check - I haven't tried that.

Here's how to reproduce:
LLINEUP -> x=observation date, y=Plasmodium by RDT, num=Yes, denom=all
(basically there are no 0/0s)

bobular:
> I plan on removing my added code in `nullZeroHack` that handles the `marker.symbol` array, but the original version of the function should stay, right?

The `nullZeroHack` was originally designed to handle the 0/0 cases because their y values were `null` (though sadly my code comments don't say that in as many words). We can remove it if we're creating a new zeroOverZero series.

bobular (commented):
> Also, there's the one property facetVariableDetails that's giving me a headache because it's a constant-size tuple (a TS concept, whereas it's just another array in JS), meaning it shouldn't be iterated over as an array, but TS/JS doesn't seem to give any method for weeding out tuples from general arrays. So I'm just excluding it manually, which is annoying. Any thoughts?

That's annoying.

The only thing I can think of is to, explicitly exclude tuples of size 1 and 2 something like this:

```
type Foo = {
  x : number[],
  a : [ number, number ] | [ number ]
};
type PickByType<T, Value> = {
  [P in keyof T as T[P] extends Value | undefined ? P : never]: T[P];
};
type OmitByType<T, Value> = {
  [P in keyof T as T[P] extends Value | undefined ? never : P]: T[P];
};

type Goo = OmitByType<PickByType<Foo, any[]>, [ any ] | [ any, any ]>>;

// Goo is { x: number[] }
```

It avoids having to name `facetVariableDetails` directly, but if we decide to allow 3 facet variables (we won't) it will need updating.

I don't see any major and potentially impactful uses of `any`, in terms of technical debt/maintenance. Let me know if I can look again at this though.

PR #1690, opened 2023-03-16
Floating plot work in SAM

bobular:
- [x] refactor plot-related code out of MapAnalysis
- [x] visualizations aren't being saved to user database - fix this
- [x] wire up "plugin options" especially overlay variable handling
- [x] make floaters responsive to viewport

The floating visualisations are now "wired up" to use the global overlay variable, and are responsive to the viewport.

They have not been wired up to the LH menu nor into draggable panels. That should be part of another ticket (which I will make soon, but not right now)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@veupathdb/eda",
-  "version": "4.1.11",
+  "version": "4.1.12",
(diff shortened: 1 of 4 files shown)


dmfalke (approved):
This looks great!

